PR #263, opened 2026-03-14
fix(copaw_worker): render newlines as in Matrix HTML messages

Jing-ze:
## Summary                                                                                                                                       
                                                                                                                                                   
  Fix newlines not rendering in Matrix channel messages. Matrix clients (e.g. Element) ignore `\n` in the plain-text `body` field and only render
  line breaks via HTML `<br>` tags, causing multi-line AI replies to appear as a single collapsed line.                                            
                                                                      
  ## Changes                                                                                                                                       
                                                                      
  In `copaw/src/copaw_worker/matrix_channel.py`:
                                                
  1. Added `_text_to_html()` helper — escapes HTML special chars and converts `\n` to `<br>`
  2. `send()` — all outgoing messages now include `format: org.matrix.custom.html` and a `formatted_body`
  3. `_apply_mention()` — reuses the already-converted `formatted_body` from `send()` to avoid double-escaping

  ## Root Cause

  The Matrix spec requires newlines to be expressed via `formatted_body` (HTML). The plain-text `body` field is only a fallback for non-HTML
  clients and its `\n` characters are not rendered as line breaks.

### DIFF
--- a/copaw/src/copaw_worker/matrix_channel.py
+++ b/copaw/src/copaw_worker/matrix_channel.py
@@ -7,6 +7,7 @@ so CoPaw's channel registry picks it up automatically.
 from __future__ import annotations
 
 import asyncio
+import html
 import io
 import logging
 import mimetypes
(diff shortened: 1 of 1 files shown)


johnlanni (approved):
LGTM

PR #298, opened 2026-03-16
feat(install): add Alibaba Cloud Docker installation script

nillikechatchat:
## 功能描述

在安装脚本中添加了阿里云的 Docker 一键安装功能。

## 主要改动

1. 当检测到 Docker 未安装时，自动使用阿里云镜像源安装 Docker
2. 安装完成后自动启动 Docker 服务
3. 将安装过程的详细输出重定向到日志文件，保持终端界面简洁

## 优势

- 使用阿里云镜像源，对于中国用户安装速度更快
- 自动化安装流程，减少用户手动操作
- 安装失败时给出明确的错误提示

---
## Function description

Alibaba Cloud's one-click Docker installation function has been added to the installation script.

##Main changes

1. When it is detected that Docker is not installed, automatically use the Alibaba Cloud image source to install Docker
2. Automatically start the Docker service after the installation is completed
3. Redirect the detailed output of the installation process to the log file to keep the terminal interface simple

## Advantages

- Using Alibaba Cloud image source, the installation speed is faster for Chinese users
- Automate the installation process to reduce user manual operations
- Give a clear error message when the installation fails


johnlanni:
Thanks for the contribution! I reviewed the changes and here are my thoughts:

## Summary
This PR adds automatic Docker installation via the Alibaba Cloud mirror when Docker is not found during the HiClaw install process. It handles both interactive (offers to install) and non-interactive (exits with error) modes.

## Review

The code looks good overall. A couple of minor observations (not blocking):

1. **`bash -s docker --mirror Aliyun`** — the word `docker` is passed as a positional argument (`$1`) to the get.docker.com script. This is likely unintentional; the standard invocation is `bash -s -- --mirror Aliyun`. The extra positional arg is harmless since the script ignores it, but it would be cleaner to drop it or add `--` as a separator.

2. **`sudo` usage** — the install script may already be running as root (common for server setups). `sudo` works fine in that case (no-op), so this is not a bug, just something to be aware of.

No functional bugs found. **LGT** 👍

Key points:
- Correctly gates the auto-install behind non-interactive check
- Properly verifies Docker is actually available after install before continuing
- Falls back to clear error messages on failure
- Good practice redirecting verbose output to the log file

shiyiyue1102:
Thank you for trying to improve installation for users in China. The current installer deliberately detects Docker or Podman and reports prerequisites instead of downloading a remote installation script and performing sudo host mutations automatically. Running get.docker.com, starting services, and enabling them from inside the AgentTeams installer does not match the current safety and platform model. We are therefore closing this PR. Mirror acceleration remains supported through documented build and registry configuration.

PR #348, opened 2026-03-18
fix(manager): use gateway health check for welcome message readiness in aliyun deployment

Jing-ze:
## Summary
- Replace Matrix room member polling with local OpenClaw gateway health check (`http://127.0.0.1:18799/`) to determine readiness before sending the welcome message
- Increase timeout from 120s to 300s to accommodate slower gateway startup on Aliyun

## Test plan
- [x] Deploy on Aliyun and verify welcome message is sent after gateway becomes ready
- [x] Verify timeout behavior when gateway fails to start


johnlanni (approved):
LGTM

PR #421, opened 2026-03-24
fix(cloud): reliable runtime detection and welcome message delivery

Jing-ze:
## Summary
- Set `HICLAW_RUNTIME=aliyun` explicitly via Dockerfile ENV instead of relying on OIDC file detection at runtime, which can be fragile depending on when `hiclaw-env.sh` is sourced
- Updated `hiclaw-env.sh` to respect a pre-set `HICLAW_RUNTIME` and only auto-detect when unset (no regression for docker/local mode)
- Added explicit Matrix room join with retry (3 attempts) before sending the welcome message, fixing a race condition where the message could land before Manager joins the DM room

## Test plan
- [x] Build `Dockerfile.aliyun` and verify `HICLAW_RUNTIME=aliyun` is set in the container environment
- [x] Verify docker/local mode still auto-detects correctly (HICLAW_RUNTIME not pre-set)
- [x] Deploy to cloud and confirm welcome message is delivered on first boot
- [x] Simulate transient Matrix failure to verify join retry logic


johnlanni (approved):
LGTM

PR #434, opened 2026-03-25
fix: Fix the reinstall bug in Powershell script

CH3CHO:
1. Fix the script exiting issue when choosing reinstall.
2. Use the unified exit method and remove the colon after "Press Enter to exit".

fixes https://github.com/alibaba/hiclaw/issues/403


johnlanni (approved):
Thanks

PR #524, opened 2026-03-31
fix(install): create /tmp/import before apply copy

### LINKED ISSUE #522: hiclaw-apply.sh copies files into /tmp/import before ensuring the directory exists
## Bug Description

What happened:
`install/hiclaw-apply.sh` copies YAML files into `hiclaw-manager:/tmp/import/<filename>` before ensuring `/tmp/import` exists inside the container.

On a fresh install, or any environment where `/tmp/import` has not already been created by another script, the wrapper can fail during `docker cp` with an error like:

```text
Could not find the file /tmp/import in container
```

This causes the host-side declarative apply wrapper to fail before `hiclaw apply` is even invoked.

What I expected:
`hiclaw-apply.sh` should create `/tmp/import` before copying any files into the Manager container, so declarative apply works reliably on fresh environments.

## Steps to Reproduce

1. Start from a fresh HiClaw installation where `hiclaw-manager` is running.
2. Prepare a valid declarative YAML file, for example a Team or Worker resource.
3. Run `bash install/hiclaw-apply.sh -f <resource.yaml>`.
4. Observe that the wrapper may fail at `docker cp` if `/tmp/import` does not already exist in the container.

## AI Analysis

Tool used: Codex

Analysis:
- `install/hiclaw-apply.sh` is the host-side wrapper for declarative YAML management and forwards to the container-internal `hiclaw apply` CLI.
- In the original flow, the script copied files to `hiclaw-manager:/tmp/import/<filename>` before ensuring `/tmp/import` existed.
- Other wrappers already follow the correct pattern and create `/tmp/import` before copying files:
  - `install/hiclaw-import.sh`
  - `install/hiclaw-import.ps1`
- This makes `hiclaw-apply.sh` inconsistent with the existing import/apply wrappers and explains why it can fail on otherwise valid fresh installs.
- The bug is isolated to the host-side wrapper. It does not require changing controller behavior or Manager runtime behavior.

## Relevant Logs

```shell
Successfully copied 2.05kB to hiclaw-manager:/tmp/import/repro-delete-team.yaml
Error response from daemon: Could not find the file /tmp/import in container hiclaw-manager
```

## Component

Install / Setup

## Version / Commit

0042e55

codesssss:
## Summary

Fixes the host-side declarative apply wrapper so it creates `/tmp/import` before copying YAML files into the Manager container.

Fixes #522

## Changes

- move `/tmp/import` creation ahead of the file-copy loop in `install/hiclaw-apply.sh`
- keep the wrapper behavior otherwise unchanged

## Why

`hiclaw-apply.sh` could fail on fresh environments because it tried to copy files into `hiclaw-manager:/tmp/import/...` before ensuring that `/tmp/import` existed.

## Validation

- `bash -n install/hiclaw-apply.sh`
- verified `bash install/hiclaw-apply.sh -f /tmp/repro-delete-team.yaml` succeeds without the previous missing `/tmp/import` error

### DIFF
--- a/install/hiclaw-apply.sh
+++ b/install/hiclaw-apply.sh
@@ -45,6 +45,9 @@ if ! ${CONTAINER_CMD} ps --filter name=hiclaw-manager --format '{{.Names}}' 2>/d
     error "hiclaw-manager container is not running"
 fi
 
+# Ensure /tmp/import exists before copying files into container
+${CONTAINER_CMD} exec hiclaw-manager mkdir -p /tmp/import 2>/dev/null || true
+
 # ============================================================
 # Copy YAML files and referenced packages into container
 # ============================================================
@@ -73,9 +76,6 @@ for arg in "$@"; do
 
     ARGS+=("${arg}")
 done
-
-# Ensure /tmp/import exists
-${CONTAINER_CMD} exec hiclaw-manager mkdir -p /tmp/import 2>/dev/null || true
 
 # ============================================================
 # Forward to hiclaw CLI inside container


johnlanni (approved):
LGTM

PR #542, opened 2026-04-03
feat: add qwen3.6-plus and set as default model

Jing-ze:
## Summary

- Add `qwen3.6-plus` to known models with 1M context window and 64K max tokens (image supported)
- Change the default model from `qwen3.5-plus` to `qwen3.6-plus` across all install scripts, CLI, configs, shell scripts, docs, tests, and CI
- Retain `qwen3.5-plus` as an available model option (case branches, known-models list, config templates)

## Test plan

- [x] Go build passes (`go build ./...`)
- [x] Go tests pass (`go test ./...`)
- [x] Python tests (`copaw/tests/test_bridge.py`) — needs pytest environment
- [x] Verify install script menu shows `qwen3.6-plus` as option 1
- [x] Verify `hiclaw apply worker --name test` defaults to `qwen3.6-plus`
- [x] Verify switching to `qwen3.5-plus` still works via model-switch skill


johnlanni (commented):
coding plan还没有支持qwen3.6-plus

PR #558, opened 2026-04-04
fix(install): allow modification of preset values on upgrade

nillikechatchat:
## Summary

Previously, several step_* functions would skip prompts when values were already set during upgrade scenarios. This prevented users from modifying settings by pressing 'b' to go back and re-enter values.

## Fixed Functions

| Function | Variable | Fix |
|----------|----------|-----|
| `step_admin()` | `HICLAW_ADMIN_PASSWORD` | Always call prompt_optional for admin password |
| `step_network()` | `HICLAW_LOCAL_ONLY` | Add upgrade branch for network access mode |
| `step_volume()` | `HICLAW_DATA_DIR` | Add upgrade branch for data volume |
| `step_workspace()` | `HICLAW_WORKSPACE_DIR` | Add upgrade branch for workspace directory |

## Added i18n Messages

- `port.local_only.title_short` (网络访问模式 / Network Access Mode)
- `port.local_only.val_local` (仅本机 / Local only)
- `port.local_only.val_external` (允许外部访问 / External access)
- `data.title_short` (数据持久化 / Data Persistence)
- `workspace.title_short` (Manager 工作空间 / Manager Workspace)

## Test Plan

1. Run `./install/hiclaw-install.sh` with existing configuration (upgrade mode)
2. Press 'b' to go back to admin, network, volume, or workspace step
3. Verify that users can modify the previously set values
4. Confirm that new installations still work correctly

## Related

This fix follows the pattern established by `step_runtime()`, `step_e2ee()`, `step_docker_proxy()`, and `step_idle()` which already handle upgrade scenarios properly.


johnlanni:
Reviewed the changes — LGTM 👍

**Summary of changes:**
This PR fixes the installer upgrade flow so users can modify previously-set values when going back through steps. Four functions are updated:

- **`step_admin()`**: Removes the early-return bypass for preset passwords; `prompt_optional` is now always called so the user can change the admin password on upgrade.
- **`step_network()`**: Adds an upgrade branch that displays the current network access mode and lets the user re-select local-only vs. external access.
- **`step_volume()`**: Adds an upgrade branch that shows the current data volume and accepts a new value, defaulting to keeping the existing one.
- **`step_workspace()`**: Same pattern as `step_volume()` for the workspace directory.

All four functions follow the same upgrade pattern already established in `step_runtime()`, `step_e2oe()`, etc. The new i18n keys for short titles and display values are properly added in both zh/en. No bugs found — the logic is consistent and the back-button handling (`STEP_RESULT="back"`) is preserved in every branch.

shiyiyue1102:

---

shiyiyue1102:
Thank you for working on the upgrade usability issue. The installer navigation and state routing have since been redesigned around AGENTTEAMS_* variables, clear_step_vars, explicit upgrade modes, and current installer tests. This branch targets the old HICLAW_* step flow and cannot be applied safely now. We are closing it as obsolete. If the current installer still prevents changing a preset value, please open a fresh focused PR with a current reproduction and test.

PR #601, opened 2026-04-11
feat(install): show masked asterisks when typing secret input

cr7258:
## Description

Previously, entering the LLM API Key during installation showed nothing at all — users couldn't tell if their input was being registered.

This PR replaces `read -s` (silent, no visual feedback) with a custom `read_secret` function that displays `*` for each keystroke and supports backspace to delete characters, so users can tell their API Key input is being registered during installation.

image


cr7258:
@johnlanni

johnlanni (approved):
LGTM

PR #630, opened 2026-04-15
fix(controller): auto-recreate pod when deleted while CR is still Running

Erica177:
# fix(controller): auto-recreate pod when deleted while CR is still Running

## Problem

When a Worker or Manager pod is deleted externally (e.g. `kubectl delete pod`, node eviction, OOM kill) while the CR status is still `Running`, the reconciler skips reconciliation because `observedGeneration == generation` and `phase == desired`. The pod is never recreated, leaving the CR in a stale `Running` state with no backing pod.

## Fix

In `reconcileDesiredState` for both `WorkerReconciler` and `ManagerReconciler`, when `phase == desired == "Running"`, check the backend for actual pod/container status. If the pod is missing, trigger recreation via the existing `ensureWorkerRunning` / `ensureManagerRunning` methods which already handle pod recreation.

Detection runs on the existing 5-minute `reconcileInterval`, so a deleted pod is recreated within 5 minutes.

## Files Changed

- `hiclaw-controller/internal/controller/worker_controller.go` — add `ensurePodExists`, call it from `reconcileDesiredState`
- `hiclaw-controller/internal/controller/manager_controller.go` — add `ensureManagerPodExists`, same pattern

### DIFF
--- a/hiclaw-controller/internal/controller/manager_controller.go
+++ b/hiclaw-controller/internal/controller/manager_controller.go
@@ -345,10 +345,11 @@ func (r *ManagerReconciler) failManagerUpdate(ctx context.Context, m *v1beta1.Ma
 // --- Desired state reconciliation ---
 
 func (r *ManagerReconciler) reconcileDesiredState(ctx context.Context, m *v1beta1.Manager, desired string) (reconcile.Result, error) {
-	// If current phase already matches desired state, nothing to do.
-	// This also avoids calling backend Status/Stop/Delete for Managers
-	// in Docker mode, where the container name ("hiclaw-manager") doesn't
-	// include the backend's worker prefix ("hiclaw-worker-").
+	// When phase matches desired "Running", verify the pod/container still exists.
+	if m.Status.Phase == desired && desired == "Running" {
+		return r.ensureManagerPodExists(ctx, m)
+	}
+
 	if m.Status.Phase == desired {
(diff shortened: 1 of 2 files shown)


Jing-ze:
旧 reconciler 的 `reconcileDesiredState` 在 `phase == desired` 时直接短路返回，不检查 pod 是否真的还在。

目前 `fix-reconciler` 分支正在对 Worker reconciler 做结构性重构，改为声明式收敛模式。重构后的 reconcile 流程是：

```
reconcileNormal → reconcileInfrastructure → reconcileConfig → reconcileContainer
```

其中 `reconcileContainer` → `ensureContainerPresent` 每次 reconcile 都会查询 backend 的实际状态：

```go
result, err := wb.Status(ctx, w.Name)
switch result.Status {
case StatusRunning, StatusStarting, StatusReady:
    // pod 在，检查是否需要 recreate（spec 变更）
case StatusNotFound:
    return r.createWorkerContainer(ctx, s, wb)  // pod 不在，重建
default:
    // 未知状态，delete + recreate
}
```

不再有"phase == desired 就跳过"的短路逻辑，所以 pod 被外部删除（kubectl delete、node eviction、OOM 等）后会自动重建。集成测试 `TestWorkerPodDeleted_Recreates` 也已经覆盖了这个场景。

重构合并后这个 PR 就不需要了。

PR #637, opened 2026-04-16
fix: restore manager-workspace from MinIO on k8s mode pod restart

Jing-ze:
## Summary

- Fixed manager MinIO policy missing `manager/` prefix read/write access, which caused all workspace sync operations to silently fail in incluster mode
- Added `mc mirror` restore step for manager-workspace on k8s mode pod startup, matching the existing aliyun mode behavior

## Test plan

- [x] Unit tests pass (`TestMinIOAdminClient_BuildManagerPolicy`)
- [x] Verified on local kind cluster: wrote test file → synced to MinIO → deleted pod → new pod restored file from MinIO


johnlanni (approved):
LGTM

PR #662, opened 2026-04-20
fix: remove dead HICLAW_CONTROLLER_API_KEY code

Jing-ze:
## Summary

Remove `HICLAW_CONTROLLER_API_KEY` related code which is effectively dead.

## Background

The controller authenticates requests exclusively via `TokenReviewAuthenticator` — it sends bearer tokens to the K8s TokenReview API for validation against real ServiceAccount tokens. There is no static API key validation logic in the controller.

The Helm chart auto-generates a random 32-char string as `HICLAW_CONTROLLER_API_KEY` and the manager's `container-api.sh` uses it as a third-priority fallback bearer token. However, this random string can never pass K8s TokenReview validation, so it is dead code in all scenarios.

The actual working auth paths are:
1. `HICLAW_AUTH_TOKEN` — directly specified SA token
2. `HICLAW_AUTH_TOKEN_FILE` — projected SA token from volume mount (auto-rotated by kubelet)

## Changes

- `helm/hiclaw/values.yaml` — remove `controllerApiKey` field
- `helm/hiclaw/templates/secrets/runtime-env.yaml` — remove `HICLAW_CONTROLLER_API_KEY` secret generation block
- `manager/scripts/lib/container-api.sh` — remove API key fallback branch

## Test plan

- [ ] `helm template` renders without errors
- [ ] Manager pod starts and authenticates to controller via SA token as before


johnlanni (approved):
LGTM

PR #669, opened 2026-04-21
ci(test-integration): fix integration-tests skipped when build-openclaw-base is skipped

Jing-ze:
## Summary

When a PR does not modify `openclaw-base/**`, the `build-openclaw-base` job is correctly skipped via its `if` condition. However, GitHub Actions propagates the `skipped` status through the transitive dependency chain:

```
build-openclaw-base (SKIPPED) → build-images (SUCCESS) → integration-tests (SKIPPED)
```

Even though `build-images` already handles this with `if: always() && ...` and completes successfully, `integration-tests` lacks the same guard — so it gets skipped by default. This means **integration tests never run on PRs that don't touch `openclaw-base/`**, which is the majority of PRs.

## Fix

Add `if: always() && needs.build-images.result == 'success'` to the `integration-tests` job, matching the pattern already used by `build-images` to handle the optional `build-openclaw-base` dependency.


johnlanni (approved):
LGTM

PR #670, opened 2026-04-21
fix(copaw): adapt bridge to CoPaw's config.json + agent.json split

maplefeng-a:
## Summary

Adapt the CoPaw bridge to the new CoPaw config layout, which splits settings into two files:

- `.copaw/config.json` — global singleton (security guards, runner-level settings)
- `.copaw/workspaces/<agent>/agent.json` — per-agent identity, channels, prompts

The old bridge only wrote `agent.json`, so a stale `config.json` silently overrode per-agent security settings (forcing tool-approval prompts). It also wrote every field from openclaw.json on every restart, wiping out user-owned fields like `env`, console settings, and custom keys.

## What changed

- **Two-stage bridge**:
  1. On first boot, install `config.json` / `agent.json` from templates in `copaw/src/copaw_worker/templates/`.
  2. On every restart, overlay only controller-owned fields (Matrix `homeserver` / `user_id` / `access_token` / `encryption` / `history_limit`, `running.max_input_length`, `running.embedding_config`, heartbeat). Allowlists merge as union; `channels.matrix.groups` is deep-merged. Everything else (identity, language, `channels.console`, `env`, user-custom keys) is owned by the agent/user and survives restart.
- **New API**: `bridge_controller_to_copaw(cfg, working_dir, profile="worker"|"manager", agent="default")`. CLI: `python -m copaw_worker.bridge --openclaw-json <p> --working-dir <wd> [--profile manager]`. Old `bridge_openclaw_to_copaw` kept as a thin alias.
- Remove `manager/scripts/init/bridge-manager-config.py`; Manager now uses the same CLI with `--profile manager`.
- `start-copaw-manager.sh` ships a one-shot migration: archive any legacy `config.json` with `security.tool_guard.enabled=true` as `config.json.legacy-<ts>` so the template re-seeds with security off.
- Update `worker.py` docs, `tests/test-01-manager-boot.sh` assertions (check `agent.json` instead of legacy `config.json`), and rewrite `copaw/tests/test_bridge.py` around template install + user-edit preservation.

## Why

- Fixes tool-approval prompts on every call (stale global `config.json` overriding `agent.json`).
- Fixes user edits (e.g. `env` field, console toggle, custom keys in `agent.json`) being silently lost on restart.
- Fixes Manager-only fields (`user_id`, `require_mention`, `TOOLS.md`, heartbeat) never applying because the old Manager wrapper post-processed a file the new bridge no longer writes.

## Test plan

- [x] `cd copaw && pytest tests/test_bridge.py`
- [x] `tests/test-01-manager-boot.sh`
- [x] Manual: restart Manager/Worker, verify user-added `env` in `agent.json` survives and tool calls no longer prompt for approval.


johnlanni (approved):
LGTM

PR #671, opened 2026-04-21
Fix DingTalk link and update WeChat group image

max-wc:
Updated the DingTalk group link and replaced the WeChat group image.


johnlanni (approved):
LGTM

PR #673, opened 2026-04-21
fix(copaw): emit Element-style visible mentions in MatrixChannel

maplefeng-a:
## Summary

- `MatrixChannel._apply_mention` now emits all three layers of a mention in one shot: `body` (plain-text MXID fallback), `formatted_body` (with a `matrix.to` anchor), and `m.mentions.user_ids` per MSC3952. This satisfies openclaw >= 2026.4.x's mention monitor, which requires BOTH the metadata AND a visible mention — previously CoPaw-issued messages only wrote the metadata and were silently dropped with `reason: "no-mention"`, so the CoPaw ↔ OpenClaw cross-runtime path was effectively broken.
- `send()` / `send_media()` now accept `meta["mention_user_ids"]` so callers can target specific workers explicitly. Priority order: **explicit list > `meta["sender_id"]` / `meta["user_id"]` fallback > regex scan of `body`**. The agent's own MXID is filtered out of targets to prevent self-mention loops.
- New `copaw/tests/test_channel_mention.py` (8 cases) pins the three-layer invariant across explicit / sender-id / body-scan / multi-target / self-filter / media-event / no-target paths.

## Background

Follow-up to the problem surfaced by `test(test-15)` in #654: openclaw 2026.4.x tightened `extensions/matrix/src/matrix/monitor/mentions.ts:resolveMentions` to require a visible mention (a `matrix.to` link in `formatted_body`, or a regex match on identity) on top of `m.mentions.user_ids`. A Worker created from a minimal SOUL (no custom identity regex) never woke up on messages from a CoPaw agent under the old behaviour.

## Test plan

- [x] `copaw/tests/test_channel_mention.py` unit tests covering the three mention-source paths, multi-target anchors, self-mention filtering, and media-event formatted_body synthesis
- [x] Full local integration suite: 21/21 PASS (test-01 through test-21-team-project-dag, including test-15-import-worker-zip and test-06-multi-worker which exercise cross-worker mentions)
- [ ] CI green

## Notes

`copaw/src/matrix/` is vendored into **both** `copaw-worker` and `hiclaw-manager-copaw` (see `copaw/AGENTS.md` pitfall #1). Both images need to be rebuilt on release; otherwise Manager and Worker behaviour will drift.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,6 +4,8 @@ Record image-affecting changes to `manager/`, `worker/`, `openclaw-base/` here b
 
 ---
 
+- fix(copaw): `MatrixChannel._apply_mention` now emits Element-style visible mentions (`matrix.to` anchor in `formatted_body` + MXID in `body` + `m.mentions.user_ids`) so openclaw >= 2026.4.x workers actually wake up instead of dropping the event as `no-mention`; `send()` / `send_media()` also accept `meta["mention_user_ids"]` for explicit targets.
+
 - feat(manager): add Team Leader heartbeat and worker lifecycle builtins for team-scoped sleep/wake coordination (unreleased)
 
 - fix(manager): make find-skills use deterministic script paths in worker/copaw SKILL.md, render canonical install/search commands from `hiclaw-find-skill.sh`, and treat "import/install xxx skill from market" as a direct install trigger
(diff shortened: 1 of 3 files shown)


johnlanni (approved):
LGTM

PR #681, opened 2026-04-23
fix(controller): preserve AI-route allowedConsumers across controller restart

Jing-ze:
## Problem

Every time the controller boots, `initializer.initGatewayRoutes` calls `gateway.EnsureAIRoute` with a freshly-declared `AllowedConsumers: []string{}`. `EnsureAIRoute` unconditionally `POST`s to `/v1/ai/routes`, which Higress treats as an upsert and rewrites `authConfig.allowedConsumers` to `[]`. Between that `POST` and the Manager/Worker reconcilers calling `AuthorizeAIRoutes` again, Manager and every Worker fail LLM calls with HTTP 403 from key-auth.

Three supporting flaws make the symptom harder to observe and harder to recover from:

1. `modifyAIRoutes` swallows every per-route `GET` / `PUT` error (silent `break` then `return nil`), so a broken `PUT` never surfaces as a reconcile error.
2. `manager_reconcile_infra.reconcileManagerInfrastructure` logs `EnsureManagerGatewayAuth` failures as `(non-fatal)`, so controller-runtime never re-queues them with backoff.
3. The worker restart path in `ReconcileMemberInfra` has no equivalent of the Manager's defensive re-auth, so a worker stays 403-locked until the next spec-change event.

Symptom in the wild: after restarting `hiclaw-controller`, Manager + all Workers flap on 403 until the next reconcile that happens to touch auth. In K8s this can be tens of seconds to minutes depending on queue state; in embedded mode the flap window also covers any time the `/system/init` race moves.

This PR restores the intent of [#644](https://github.com/agentscope-ai/HiClaw/pull/644) (which no longer cleanly applies to current `main`), re-implemented against the current codebase and extended with a worker-side defensive re-auth that #644 did not cover.

## Solution

Establish a clear **ownership split** for AI-route state:

- **Initializer** owns the AI-route *skeleton* (name, path, upstream, key-auth framework).
- **Manager / Worker reconcilers** own `authConfig.allowedConsumers` via `AuthorizeAIRoutes` / `DeauthorizeAIRoutes`.

Concrete changes:

### Route skeleton / auth ownership

- `gateway.AIRouteRequest` drops the `AllowedConsumers` field entirely.
- `gateway.EnsureAIRoute` is rewritten **GET-first**:
  - Route exists → no-op. Logs a `WARN` only if `pathPredicate` / upstream provider diverge. **Never** writes auth state.
  - Route absent → `POST` body enables the key-auth framework but deliberately omits `allowedConsumers`, so Higress defaults it to `[]` exactly once at creation and the reconcilers populate it afterwards.
- `initializer.initGatewayRoutes` no longer passes any auth data to `EnsureAIRoute`.

### Error propagation

- `modifyAIRoutes` captures the first per-route error via a `firstErr` / `lastErr` pattern and returns it, so controller-runtime re-queues with backoff instead of silently succeeding.
- `manager_reconcile_infra.reconcileManagerInfrastructure` returns `EnsureManagerGatewayAuth` errors instead of logging them as `(non-fatal)`.

### Worker-side defensive re-auth (new, not in #644)

- New `Provisioner.EnsureWorkerGatewayAuth` mirrors the Manager method, using `consumer = "worker-<name>"`.
- Added to `service.WorkerProvisioner` interface and to `test/testutil/mocks.MockProvisioner` (with `EnsureWorkerGatewayAuthFn` hook and `Calls.EnsureWorkerGatewayAuth` record).
- `controller.ReconcileMemberInfra` calls it after `RefreshCredentials` in the existing-member branch, so a restarted worker is re-bound to AI routes on the first reconcile instead of waiting for the next spec-change event. Errors propagate.

### Security hardening (same PR to keep the auth surface consistent)

A separate concern surfaced by a security scan — the hard-coded `admin`/`admin` fallback credential — was fixed alongside because it touches the exact same auth code path:

- `gateway.NewHigressClient` no longer rewrites empty `AdminUser` / `AdminPassword` to `"admin"`.
- `config.LoadConfig()` no longer supplies `"admin"` as the default for `HICLAW_ADMIN_USER` / `HICLAW_ADMIN_PASSWORD` (affects `HigressAdminUser`, `HigressAdminPassword`, `MatrixAdminUser`, `MatrixAdminPassword`, and `WorkerEnv.AdminUser`).
- The install / deploy layer is now the **sole source** of these values. A missing env var fails fast at login instead of silently running as `admin`/`admin`.
- The embedded-mode one-shot `admin`/`admin` bootstrap path inside `HigressClient.ensureSession` (guarded by `AllowDefaultAdminFallback`) is intentionally preserved — it is the legitimate mechanism for re-keying a freshly-booted Higress instance and is already protected by `AdminUser != "admin" || AdminPassword == "admin"`.

## Tests

New cases in `internal/gateway/client_test.go` lock in the contract:

- `TestEnsureAIRoute_ExistingSkipsPOST` — never writes when the route exists.
- `TestEnsureAIRoute_MissingCreatesSkeletonWithoutAllowedConsumers` — `POST` body enables key-auth but omits `allowedConsumers`.
- `TestAuthorizeAIRoutes_PUTFailsPropagatesError` — non-2xx / non-409 `PUT` surfaces as an error instead of being swallowed.

## Verification

- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./internal/gateway/... ./internal/config/... ./internal/matrix/... ./internal/agentconfig/... ./internal/service/... ./internal/controller/...` ✅
- All three new unit tests pass.


johnlanni (approved):
LGTM

PR #682, opened 2026-04-23
feat(controller): align cloud-mode STS defaults with embedded MinIO and add team member accessEntries

Jing-ze:
## Background

HiClaw has two parallel storage permission paths depending on `HICLAW_STORAGE_PROVIDER`:

| Deployment | StorageProvider | Permission path | Reads `spec.accessEntries`? |
|---|---|---|---|
| Local all-in-one (`embedded`) | `minio` | `oss.MinIOAdminClient.buildWorkerPolicy` → `mc admin policy` | ❌ |
| Helm / in-cluster + bundled MinIO (`incluster`) | `minio` | same as above — `mc admin policy` | ❌ |
| Cloud (`incluster` + external OSS) | `oss` | `accessresolver` → `credprovider` sidecar → STS | ✅ |

On the MinIO path, `buildWorkerPolicy` grants every team member RW on `agents/<name>/*`, `shared/*`, **and** `teams/<team>/*`; Manager additionally gets `manager/*`.

On the cloud / STS path things diverged in two ways:

1. `accessresolver` only knew how to look up `Worker` / `Manager` CRs. For a team member it fell through `resolveWorker`, missed the `Team` CR entirely, and the default `AccessEntry` set contained no `teams/<team>/*` at all. A team worker in cloud mode could not read/write the shared team directory.
2. `LeaderSpec` / `TeamWorkerSpec` had no `accessEntries` field, so operators had no way to customize team-member scopes per deployment the way they already could on `WorkerSpec`.

As an orthogonal but related annoyance, `DefaultEntriesForWorker` granted `shared/*` as read-only while the embedded MinIO policy has always given it RW, so a standalone cloud-mode Worker couldn't write to the shared area by default.

## Solution

Bring the cloud / STS path up to parity with the embedded MinIO policy, and add the missing customization knob on team members.

### API / CRD

- `api/v1beta1`: add `AccessEntries []AccessEntry` to `LeaderSpec` and `TeamWorkerSpec` with the same semantics as `WorkerSpec.AccessEntries`. Regenerate `zz_generated.deepcopy.go`.
- `config/crd/teams.hiclaw.io.yaml` + `helm/hiclaw/crds/teams.hiclaw.io.yaml`: add a matching `accessEntries` schema block under `spec.leader.properties` and `spec.workers.items.properties`. The block is a copy of the one already on `workers.hiclaw.io.yaml` with the description extended to mention `${self.team}`.

### Defaults (`accessresolver/defaults.go`)

- `DefaultEntriesForWorker` collapses `agents/<name>/*` + `shared/*` into a single RW entry. Previously `shared/*` was R-only; embedded has always had it RW.
- New `DefaultEntriesForTeamMember` covers `agents/<name>/*` + `shared/*` + `teams/<team>/*`, all RW — mirrors `oss.MinIOAdminClient.buildWorkerPolicy` when `teamName != ""`. Leader and team workers share the exact same scope; **no cross-member elevation for the leader**.

### Resolver (`accessresolver/resolver.go`)

- `templateCtx` gains a `team` field and `expand` handles `${self.team}`.
- `ResolveForCaller` routes to a new `resolveTeamMember(name, teamName)` whenever `caller.Team != ""`. `caller.Team` is already populated by `auth.IdentityEnricher` via two field indexers (`spec.leader.name` / `spec.workerNames`), so no new lookup infrastructure is needed here.
- `resolveTeamMember` fetches the `Team` CR, matches `name` against the leader or one of the workers, picks up that member's `AccessEntries` (or falls back to `DefaultEntriesForTeamMember` when empty), and returns a `WorkerSessionName(name)` session — the pod's ServiceAccount is still `hiclaw-worker-<name>` regardless of whether the member is a leader or worker.
- Team CR not-found / CRD-missing does **not** hard-fail the in-flight STS request; we fall back to defaults, matching the behaviour of `resolveWorker` / `resolveManager` under the same conditions.

### Tests

- New `accessresolver` cases: `TestResolveTeamLeader_DefaultEntries`, `TestResolveTeamWorker_DefaultEntries`, `TestResolveTeamMember_CustomEntries` (verifies `\${self.team}` expansion), `TestResolveTeamMember_TeamCRMissing`.
- `TestResolveWorker_DefaultEntries` updated to the new single-entry RW shape.
- `internal/credentials/sts_test.go::TestIssueForCaller_WorkerDefaultEntries` updated to the same shape.

## Intentional non-goals

- **Embedded / in-cluster + MinIO still does not honor `spec.accessEntries`.** `oss.MinIOAdminClient.buildWorkerPolicy` is unchanged. CR-level customization today is a cloud-only feature; unifying the two paths (making MinIO also consume the resolved `AccessEntry` set) is a follow-up.
- **Leader is not granted cross-worker access.** Team collaboration flows through `teams/<team>/*`, consistent with the embedded policy.
- No change to the AI-Gateway default set; team members that need it can declare it explicitly in `accessEntries`.


johnlanni (approved):
LGTM

PR #685, opened 2026-04-23
refactor(controller): restructure HumanReconciler with service layer + lazy Matrix login

Jing-ze:
## Problem

`HumanReconciler` is the last reconciler in `hiclaw-controller` still following the pre-refactor shape — monolithic `human_controller.go`, imperative `handleCreate` / `handleUpdate` branches, and a direct dependency on `matrix.Client`. Worker / Manager / Team were all migrated to the service-layer + declarative convergence pattern in earlier PRs.

Naively adopting that pattern would have introduced a **Matrix device-session leak**: the pattern uses a 5-minute periodic requeue, and `POST /_matrix/client/v3/login` without `device_id` never reuses sessions. A Human with a stable spec would accumulate ~288 new device sessions per day on Tuwunel — the same class of regression Worker/Manager previously fixed via cached `WorkerCredentials.MatrixToken`. A Human has no equivalent credential store, so we need a different mitigation.

## Solution

**Structural refactor** (aligns Human with Worker/Manager/Team):

- Introduce `service.HumanProvisioner` (implemented by `*service.Provisioner` in `provisioner_human.go`) so the reconciler talks to an interface, enabling envtest-free unit tests.
- Split `human_controller.go` into phase files: `human_scope.go` + `human_reconcile_{infra,rooms,legacy,delete}.go`.
- Replace `handleCreate` / `handleUpdate` with a single declarative desired-vs-observed convergence loop; unify status writes behind a `defer` + `client.MergeFrom` + `Status().Patch`.
- Stop calling `EnsureUser` on steady-state reconciles so `!admin users reset-password` no longer silently clobbers user-initiated password changes in Element.

**Lazy Matrix login** (prevents the device-session leak):

- `reconcileHumanInfra` no longer issues `Login` on steady-state reconciles. It only calls `EnsureHumanUser` on first-time provisioning (which returns a usable access token for free).
- New `ensureUserToken` helper in `reconcileHumanRooms` fires `Login` **only when we actually need to `/join` a new room**, and caches the result across invite/join iterations in the same reconcile.
- Steady-state reconciles (`desired == observed`) issue zero Matrix writes. Pure-revoke reconciles issue zero Logins. New-room reconciles issue exactly one Login, matching the pre-refactor call pattern.
- Login failure (stale password) is non-fatal: we log and fall through to admin-only invite, never fall back to `EnsureUser` (which would reset the user's chosen password).

## Tests

Adds `test/testutil/mocks/MockHumanProvisioner` and `human_controller_test.go` with 7 cases — all pass without envtest:

- `Create_HappyPath` — single reconcile converges, finalizer added, zero `Login` calls
- `Update_AddRoom` — exactly 1 `Login` + 1 invite/join for the new room only
- `Update_RevokeRoom` — kick only, no `Login`
- `Update_PendingResource` — Worker with no `RoomID` is skipped without error
- **`SteadyState_NoLogin`** — locks in the device-bloat fix: already-converged Human triggers 0 Matrix writes
- `Delete_FinalizerCleanup` — `ForceLeaveRoom` for all rooms, finalizer removed
- `Login_StalePassword` — admin-only invite fallback, no `EnsureUser` retry, no error

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/controller -run TestHumanReconciler -count=1` — 7/7 pass
- [x] `go test ./... -count=1 -short` — all packages green
- [ ] Embedded-mode smoke test: create Human, verify Matrix account + room membership + `humans-registry.json` entry
- [ ] Incluster-mode smoke test: same as above without the registry check

## Backward compatibility

- No CRD changes.
- No new password-management features (`InitialPassword` semantics unchanged).
- For embedded/incluster operators: identical create / update / delete / revoke behavior on the happy path. The only observable differences vs the pre-refactor controller are the two bug fixes called out above (no periodic reset-password clobber, no device-session leak).


johnlanni (approved):
LGTM

PR #687, opened 2026-04-23
feat(controller): propagate CR metadata.labels + new spec.labels to managed Pods

Jing-ze:
## Background

Two things on the Hiclaw controller had been asked for by operators and kept biting us in multi-tenant / multi-instance deployments:

1. **Controller identity was not uniformly stamped on managed objects.** `hiclaw.io/controller`, `hiclaw.io/runtime`, and `hiclaw.io/role` were only applied to some Pods and not at all to ServiceAccounts or credential Secrets. Two `hiclaw-controller` instances sharing the same namespace could therefore cross-watch each other's Pods through the label-scoped informer cache and clobber each other's reconcile state. The Pod informer cache itself was not scoped to `hiclaw.io/controller=<name>` either, so even after fixing the labels the watch stream was still shared.
2. **There was no way for users to put their own labels on managed Pods.** `metadata.labels` on a `Worker` / `Manager` / `Team` CR was ignored, and there was no `spec.labels` field — anyone wanting to select Pods by team / environment / tenant had to fork the controller.

This PR addresses both in one stack, since the second one has to be built on top of the first (the CR `metadata.labels` / `spec.labels` merge has to sit *below* the controller's own identity labels, which means those identity labels have to be stable and consistently applied first).

### A. Identity label hardening (commits 1-2)

- Stamp `hiclaw.io/controller`, `hiclaw.io/runtime`, and `hiclaw.io/role` consistently on every Pod, ServiceAccount, and credential Secret created by the controller. `WorkerReconciler` / `ManagerReconciler` / `Provisioner` / `SecretCredentialStore` each grow a `ControllerName` field wired from `Config.ControllerName` in `app.go`.
- Manager Pods additionally gain `hiclaw.io/role=manager` and `hiclaw.io/runtime=<resolved>` so they match the Worker/Team Pod schema.
- The Kubernetes backend no longer defaults `app` / `hiclaw.io/worker` / `hiclaw.io/manager`. Callers own the full label set; the backend only stamps `hiclaw.io/runtime` (which depends on the post-resolution runtime value the reconciler doesn't know).
- Drop the unused `WorkerBackend.List(ctx)` interface method + `K8sBackend.List` / `DockerBackend.List` implementations — zero production callers and its `app=hiclaw-worker` selector would have mixed standalone / team-leader / team-worker Pods indistinguishably.
- Drop `hiclaw.io/type=worker-credentials` from credential Secrets and `hiclaw.io/created-by=controller` from Pods — both had no in-tree consumer.
- Scope the Pod informer cache (`opts.Cache.ByObject` in `startInCluster`) and `podLifecyclePredicates` to `hiclaw.io/controller=<ctrlName>` so two controllers in one namespace truly don't see each other's Pods. Defense-in-depth covered by new `pod_predicate_test.go`.
- Align the credential `Secret` \`app\` label with the Pod/SA value for the same worker via `ResourcePrefix` (was hardcoded `"hiclaw"`).

### B. User-defined labels feature (commits 3-5)

- New `spec.labels` (\`map<string,string>\`) on `Worker`, `Manager`, and per-member on `Team.spec.leader.labels` / `Team.spec.workers[].labels`. All four fields carry \`omitempty\` so existing Teams keep hash stability via `hashMemberSourceSpec`.
- CR `metadata.labels` is now honored as well.
- Final Pod label set is layered low-to-high:

  ```
  pod-template overlay  <  CR metadata.labels  <  CR spec.labels  <  controller system labels
  ```

  Same-key collisions are silently overridden by the higher layer — users writing `hiclaw.io/controller: attacker` into their labels cannot spoof controller identity, the value is simply stamped over by the controller's own.
- For `Team`, `metadata.labels` fans out to the leader Pod and **every** worker Pod. Per-member `spec.leader.labels` / `spec.workers[i].labels` stay per-member and never leak across members.
- **\`metadata.labels\` changes do NOT recreate running Pods.** They don't bump \`Generation\`, so no spec-changed rollout is triggered; the running Pod keeps its label set until the next legitimate spec rollout. **\`spec.labels\` changes DO recreate** — they bump Generation and feed through the normal spec-changed recreate path. This was a deliberate choice to avoid cosmetic label edits causing worker rollouts.
- New shared \`mergeLabels\` helper (low-to-high overwrite, nil layers skipped, returns nil when every layer is empty), used by `workerMemberContext`, `createManagerContainer`, and `buildDesiredMembers`.
- `zz_generated.deepcopy.go` is manually extended for the four new `Labels` map fields (the repo has historically hand-maintained this file; no `make generate` target).

### C. CRD sync (commit 6)

- Copy `hiclaw-controller/config/crd/{workers,managers,teams}.hiclaw.io.yaml` into `helm/hiclaw/crds/` so the chart can consume the new field. This also incidentally aligns a few earlier drifts the helm CRDs had (e.g. `WorkerSpec.state`, richer `ManagerStatus.phase` enum).

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./api/... ./internal/... ./cmd/...` — all green
- [x] `make test-integration` (envtest-backed) — all green
- [x] New unit tests:
  - `api/v1beta1/types_test.go` — DeepCopy isolation across all four Specs + cascade through `TeamSpec`
  - `internal/controller/labels_test.go` — priority, nil/empty skipping, all-nil returns nil, no input mutation
  - extended `worker_controller_test.go` and `team_controller_test.go` — four-layer merge, reserved-key silent override, per-member isolation, team-wide metadata fan-out
  - new `manager_container_test.go` — captures backend `CreateRequest.Labels` through a mock backend to lock in the manager path
  - new `pod_predicate_test.go` — cross-controller rejection on Create/Delete/Update
  - new \`AppLabelHonorsResourcePrefix\` case in `credentials_test.go`
- [x] New integration tests in `test/integration/controller/`:
  - Worker/Manager/Team: labels fan out from `metadata.labels` + `spec.labels` into `backend.CreateRequest.Labels` with system labels winning on collision
  - Worker/Manager: changing `metadata.labels` does NOT increment `CreateCount` / `DeleteCount` (no Pod recreation)
  - Team: per-member labels don't leak across members

## Notes / non-goals

- **No Pod-label live-sync.** Changing `metadata.labels` on an existing CR does not retroactively patch the running Pod's labels. If that turns out to be desired, it can be added as a separate reconcile phase later.
- **No validating webhook for reserved keys.** We chose "silent override" over reject — putting `hiclaw.io/controller` into `spec.labels` is allowed but the controller's own value always wins.


johnlanni (approved):
LGTM

PR #688, opened 2026-04-23
refactor(controller): stamp CR-to-Pod controller OwnerReference instead of inheriting from controller Pod

Jing-ze:
## Problem

`hiclaw-controller` used a non-standard scheme to set `ownerReferences` on the agent Pods it creates: `K8sBackend` would look up *its own Pod* at Create time, cache the resulting `ownerReferences` (after stripping out `ReplicaSet` entries), and then stamp the same list onto every managed Pod via `PodOverlay.OwnerReferences`. In practice that meant the Pod backing a `Worker` / `Team` / `Manager` CR was owned by the controller's `Deployment` (or nothing, when run standalone), not by the CR that actually drives its lifecycle.

Two concrete problems fell out of this:

1. **GC did not match the declarative model.** Deleting a `Worker` / `Team` / `Manager` CR did not garbage-collect the corresponding Pod. Clean-up relied entirely on the finalizer explicitly calling `backend.Delete` — if a finalizer got patched off (stuck `Terminating`, controller unreachable, operator intervention), the Pod was orphaned. `helm uninstall` followed by a `Terminating` CR with its finalizer stripped was the typical way to reproduce the leak.
2. **Non-trivial code to maintain.** ~140 lines of `K8sBackend` were dedicated to `getCurrentPod` + `ownerRefsCache` + `controllerOwnerRefs` + `filterOutReplicaSetOwners` + `classifyAPIError`, plus a `PodOverlay.OwnerReferences` channel used by nobody but the backend itself. All of it is an attempt at something the `controller-runtime` library already does correctly in one line.

## Solution

Move to the standard controller-runtime pattern: **the owning CR is the Pod's controller owner**. The CR (`Worker` / `Team` / `Manager`) is threaded through a new `CreateRequest.Owner` field, and `K8sBackend.Create` calls `controllerutil.SetControllerReference(req.Owner, pod, k.scheme)` before submitting the Pod.

- `K8sBackend` gains a `*runtime.Scheme` constructor arg, wired through `buildWorkerBackends` from `App.scheme`.
- `CreateRequest` gets `Owner metav1.Object` (godoc'd as K8s-only; Docker backend ignores it).
- `MemberContext` gets `Owner metav1.Object`. `WorkerReconciler`, `TeamReconciler.buildDesiredMembers` (both leader and workers), and `ManagerReconciler.createManagerContainer` each populate it with the CR they're reconciling.
- Dead code removed: `PodOverlay.OwnerReferences`, `K8sBackend.ownerRefs` cache, `getCurrentPod`, `controllerOwnerRefs`, `filterOutReplicaSetOwners`, `classifyAPIError`, and their tests.

### Why keep the explicit `backend.Delete` in finalizers

I considered removing the finalizer's `wb.Delete(ctx, containerName)` calls entirely (``ReconcileMemberDelete`` + `reconcileManagerDelete`) and letting K8s GC handle all Pod cleanup. **That would silently break the embedded deployment.** Two independent reasons:

1. The embedded mode runs only `kube-apiserver` + `kine` (see `hiclaw-controller/internal/apiserver/embedded.go`). There is no `kube-controller-manager`, therefore no garbage-collector controller running — nothing watches `ownerReferences` to cascade deletes.
2. Even if GC were running, in embedded mode the "Pods" are actually **Docker containers created by `DockerBackend`**. They are never stored as `Pod` objects in the apiserver, so no K8s GC can ever reach them.

Semantics after this PR:

| mode | container cleanup on CR deletion |
|---|---|
| `incluster` (K8s backend) | finalizer's `backend.Delete` is the primary path (synchronous, surfaces errors in the controller log); `ownerReferences` is an **additive** safety net for the "finalizer got patched off" case |
| `embedded` (Docker backend) | finalizer's `backend.Delete` is the **only** path; `Owner` is ignored by `DockerBackend` |

So `wb.Delete` remains, with one small tweak: its error path now swallows `backend.ErrNotFound` because K8s GC may already have won the race, at which point the not-found error is a success signal, not something worth logging at `Error` level.

## Diff shape

```
 changelog/current.md                               |   2 +
 hiclaw-controller/internal/app/app.go              |   9 +-
 .../internal/backend/agent_pod_template.go         |  16 +-
 .../internal/backend/agent_pod_template_test.go    |  14 --
 hiclaw-controller/internal/backend/interface.go    |  10 ++
 hiclaw-controller/internal/backend/kubernetes.go   | 131 ++++------------
 .../internal/backend/kubernetes_test.go            | 164 +++++++--------------
 .../controller/manager_reconcile_container.go      |   1 +
 .../internal/controller/member_reconcile.go        |  23 ++-
 .../internal/controller/team_controller.go         |   2 +
 .../internal/controller/worker_controller.go       |   1 +
 11 files changed, 129 insertions(+), 244 deletions(-)
```

Net -115 lines, all controller-internal; no CRD / API / Helm chart changes.

## Test plan

- [x] `go build ./... && go vet ./...` (hiclaw-controller)
- [x] `go test ./internal/controller/... ./internal/backend/... -count=1`
- [x] New unit test `TestK8sCreate_SetsControllerReferenceFromOwner` asserts the stamped `ownerReferences` has `controller=true`, `blockOwnerDeletion=true`, and points at the CR passed via `CreateRequest.Owner`.
- [x] New unit test `TestK8sCreate_OwnerRequiresScheme` asserts a clear failure when `Owner` is set but the backend was built without a `Scheme`.
- [x] Obsolete tests removed: `TestK8sCreate_OwnerRefsInheritsFromControllerPod`, `TestK8sCreate_OwnerRefsCachedAcrossCreates`, `TestK8sCreate_OwnerRefsRetriesWhenLookupFails`, `TestFilterOutReplicaSetOwners`, `TestClassifyAPIError`, `TestApplyPodTemplate_OwnerReferencesFromOverlay`.
- [ ] Manual smoke on an `incluster` deployment: `kubectl delete worker <name>` → finalizer log shows one `backend.Delete`, Pod disappears within seconds. `kubectl patch` the finalizer off on a stuck CR → Pod still GC'd by the cluster GC via `ownerReferences`.
- [ ] Manual smoke on an `embedded` deployment: `hiclaw delete worker <name>` → Docker container disappears (same semantics as before the PR).


johnlanni (approved):
LGTM

PR #690, opened 2026-04-23
feat(hiclaw-controller): honor spec.env on Worker/Manager/Team CRs

Jing-ze:
## Problem

HiClaw operators have no way to inject custom environment variables into the agent containers managed by `hiclaw-controller`. Today the only env vars that reach the Worker / Manager / Team-member containers are the ones the controller's `WorkerEnvBuilder` / `ManagerEnvBuilder` build internally (matrix credentials, gateway URL, `HICLAW_*` system vars, etc.). Any operator who wants to:

- turn on debug flags for a specific runtime (e.g. `OPENCLAW_MATRIX_DEBUG=1`, `HERMES_YOLO_MODE=0`),
- pass app-level credentials or feature flags to a worker's skills/MCP servers (e.g. `GITHUB_TOKEN`, `LANG`, `HTTP_PROXY`),
- inject experimental env knobs without patching and rebuilding the controller image,

has to edit the controller source, fork the env builder, or shell into the container and set vars by hand — none of which survives a reconcile.

The underlying CR types (`Worker`, `Manager`, `Team.leader`, `Team.workers[]`) had no `env` field, the CRD schemas rejected it, and the `backend.CreateRequest.Env` path was only populated from the builders. This PR closes that gap.

## Solution

Add a new `spec.env` (`map[string]string`) field to `Worker`, `Manager`, `Team.leader`, and `Team.workers[]` CRs. User-declared env is merged with the controller-generated env at container-creation time, using a deterministic **system-wins** policy:

- If a key is produced by the env builder, the system value wins. The ignored user key(s) are logged **once** at `INFO` level with a sorted, deterministic list so operators can see exactly which keys were dropped.
- Otherwise the user value is written verbatim. Empty string values are allowed (common for feature toggles like `DEBUG=`).
- `spec.env` replaces the previous map wholesale on every reconcile, same semantics as the existing `spec.skills` / `spec.mcpServers`.

Where the merge runs:

- `createMemberContainer` — covers standalone `Worker` plus both `Team` roles via the existing `WorkerSpec` projection in `team_controller.go`.
- `createManagerContainer` — covers `Manager`.

The env builders themselves stay pure; merge is a single helper (`mergeUserEnv`) called immediately after `EnvBuilder.Build` / `BuildManager`.

### CRD schema

Each affected CRD adds:

- `env: { type: object, additionalProperties: { type: string } }`
- `propertyNames.pattern: ^[a-zA-Z_][a-zA-Z0-9_]*$`

so non-POSIX identifiers are rejected at admission time.

Helm CRDs under `helm/hiclaw/crds/` are **directly synced** from `hiclaw-controller/config/crd/`. This also incidentally fixes prior drift that was never caught: `managers.hiclaw.io.yaml` in Helm was missing `spec.state` and `status.phase` enum values, and `workers.hiclaw.io.yaml` had a minor description difference. Both are harmless no-ops for existing clusters (CRD updates are additive here), but they should have been there all along.

### Scope / intentionally deferred

- `hiclaw` CLI / REST API surface (`CreateWorkerRequest.Env` etc.) is **not** touched in this PR. Operators set env via `kubectl apply -f` directly. The CLI / REST surface can be added in a follow-up without API churn since the CR type and merge semantics are already locked in.
- Team controller still hardcodes `copaw` for team workers; this PR does not change runtime selection. `spec.env` works uniformly across `openclaw` / `copaw` / `hermes`.

## Testing

Unit tests (`go test ./internal/... ./cmd/... ./api/...` — all pass):

- `internal/controller/user_env_test.go` (new) — empty user env, non-conflict merge, conflict-keeps-system-value plus the INFO log content/order assertion, deterministic log output across shuffled input orders.
- `internal/controller/team_controller_test.go` — new `TestHashMemberSourceSpec_EnvChangeFlipsHash` proves `LeaderSpec.Env` / `TeamWorkerSpec.Env` changes flip the per-member spec hash so a Team CR edit correctly recreates only the affected members.

Integration tests (envtest, `make test-integration` — all pass, 42s):

`MockWorkerBackend` extended to record the full `backend.CreateRequest` (`CreateReqs`, plus `LastCreateReq` / `FindCreateReq` helpers) so tests can assert the exact `Env` map that reaches the backend.

Three new envtest-backed tests assert end-to-end propagation `CR.spec.env` → `backend.CreateRequest.Env`:

- `TestWorkerCreate_EnvPassesToBackend` — standalone Worker, including a `USER_EMPTY=\"\"` key to prove empty values survive, plus a colliding `HICLAW_WORKER_NAME` to prove system-wins.
- `TestManagerCreate_EnvPassesToBackend` — same assertions against the Manager path / `mockMgrBackend`.
- `TestTeam_MemberEnv_PassesToBackend` — asserts both `Team.spec.leader.env` and `Team.spec.workers[0].env` reach the backend on the respective member's `CreateRequest`, **and** that leader-only keys do not leak into the worker's env (and vice versa).

Sample log line from the integration run confirming the conflict-log path:

```
INFO user-defined env keys ignored (reserved by system) subject=standalone/test-env-5b9yz1 keys=[HICLAW_WORKER_NAME]
```

## Changelog

Added `feat(hiclaw-controller)` entry to `changelog/current.md` describing the schema change, merge policy, and the CLI / REST API deferral.

## Files touched

- `hiclaw-controller/api/v1beta1/types.go` + `zz_generated.deepcopy.go` — new `Env` field on 4 specs, deepcopy for map.
- `hiclaw-controller/config/crd/*.yaml` + `helm/hiclaw/crds/*.yaml` — new `env` schema block (+ Helm resync).
- `hiclaw-controller/internal/controller/user_env.go` (new) + `_test.go` — merge helper.
- `hiclaw-controller/internal/controller/manager_reconcile_container.go`, `member_reconcile.go`, `team_controller.go` — call `mergeUserEnv` after `EnvBuilder.Build*`.
- `hiclaw-controller/internal/controller/team_controller_test.go` — env-in-hash coverage.
- `hiclaw-controller/test/testutil/mocks/backend.go` — `CreateReqs` recording, `LastCreateReq` / `FindCreateReq` helpers.
- `hiclaw-controller/test/integration/controller/{worker,manager,team}_test.go` — 3 new propagation tests.
- `changelog/current.md` — feat entry.

## Test plan

- [x] `go test ./internal/... ./cmd/... ./api/...`
- [x] `make test-integration` (envtest, k8s 1.31) — all existing + 3 new tests pass
- [ ] Smoke: `kubectl apply -f` a Worker with `spec.env: { FOO: bar }`, `kubectl exec` into the container, confirm `echo $FOO == bar`
- [ ] Smoke: same Worker with a colliding `HICLAW_WORKER_NAME: wrong` — confirm container sees the system value and controller logs the INFO line listing the ignored key


johnlanni:
Reviewed the diff — this looks solid. Here's a summary of what this PR does:

**Adds `spec.env` (`map[string]string`) to Worker, Manager, Team.leader, and Team.workers[] CRDs**, letting users inject custom environment variables into containers.

Key design decisions:
- **System-wins precedence**: if a user-defined key collides with a controller/backend key (e.g. `HICLAW_*`, `OPENCLAW_*`), the system value wins and the collision is logged at INFO level with sorted keys for deterministic output.
- **CRD validation**: `propertyNames.pattern` enforces POSIX-compatible env var names (`^[a-zA-Z_][a-zA-Z0-9_]*$`).
- **Hash propagation**: `Env` is projected into `WorkerSpec` via `leaderWorkerSpec` / `teamWorkerSpecToWorkerSpec`, so env changes correctly trigger container recreation through `hashMemberSourceSpec`.
- **DeepCopy** is properly updated for all four new `map[string]string` fields.
- **Tests**: unit tests for `mergeUserEnv` (empty, non-conflict, conflict, sorted log), hash-change tests for leader/worker env edits, and integration tests for Worker, Manager, and Team that verify user env passes through, system env is preserved, and system wins on collision. The team test also checks that leader/worker envs don't leak into each other.

No bugs found. LGTM 👍

PR #694, opened 2026-04-23
refactor(api)!: restructure mcpServers on Worker/Manager/Team CRDs and drop controller-side MCP authorization

Jing-ze:
## Problem / Background

The current \`spec.mcpServers: []string\` field on \`Worker\`, \`Manager\`, and \`Team\` CRDs carries only an MCP server **name** and expects the controller to derive both the endpoint URL and the gateway-side consumer authorization. In practice this is broken in two different ways:

- **Local Higress mode.** Gateway-side authorization (binding a consumer to each MCP server's \`allowedConsumers\`) is already owned by \`manager/agent/skills/mcp-server-management\`, which talks to the local Higress Console directly. The controller's Higress-client MCP auth path is effectively dead code — whatever the CRD lists is silently ignored.
- **Cloud AI-Gateway mode.** \`gateway.AIGatewayClient.AuthorizeMCPServers\` / \`DeauthorizeMCPServers\` are **no-op stubs**. A \`Worker\` declaring \`mcpServers: [github]\` at install time never actually gains access to anything — \`mcporter-servers.json\` is written with a controller-guessed URL template, and the upstream endpoint has no record of the consumer.

On top of that, \`GenerateMcporterConfig\` relied on \`HICLAW_AI_GATEWAY_URL\` + hard-coded \`/mcp-servers/<name>/mcp\` path suffixes to reconstruct URLs, which forced every deployment into the same URL shape and made it impossible for operators to point at an external MCP provider.

## Solution

Narrow the CRD to declare exactly what the client (mcporter) needs, and drop controller-side MCP authorization entirely. The gateway-side authorization story is unchanged but is now explicitly out-of-band:

- Local Higress: handled by \`manager/agent/skills/mcp-server-management\`, as today.
- Cloud AI Gateway: handled by the cloud operator (console / IaC), as today.

### Breaking CRD change

\`spec.mcpServers\` switches from \`[]string\` to \`[]{name, url, transport}\`:

\`\`\`yaml
spec:
  mcpServers:
  - name: github              # keys the entry in mcporter-servers.json (tool calls use <name>.<tool>)
    url: https://apig.example.com/mcp-servers/github/mcp
    transport: http           # http (Streamable HTTP, default) | sse
\`\`\`

- \`name\` and \`url\` are \`required\`; \`transport\` defaults to \`http\`. Enforced at admission via CRD schema.
- The controller translates this slice directly into \`mcporter-servers.json\` and injects \`Authorization: Bearer <gatewayKey>\` (reusing the consumer key already issued for LLM access — no new credential to manage).

### Controller simplification

- \`gateway.Client\` interface, \`HigressClient\`, and \`AIGatewayClient\` lose \`AuthorizeMCPServers\` / \`DeauthorizeMCPServers\` and their helpers.
- \`service.Provisioner\` loses \`ReconcileMCPAuth\`, the \`McpServers\` / \`AuthorizedMCPs\` fields on every request/result, and the \`mcpServers\` parameter on \`DeprovisionManager\`.
- \`service.Deployer\` carries \`McpServers []v1beta1.MCPServer\` on the deploy request; reconcilers pass \`Spec.McpServers\` straight through.
- CLI \`hiclaw create/apply/update worker\` loses the \`--mcp-servers\` flag; users declare MCP servers in YAML manifests.

### Migration

\`migration/registry_migration.go::extractMCPServers\` is upgraded to parse both shapes of the legacy \`mcporter-servers.json\` — the wrapped \`{ "mcpServers": { ... } }\` and the bare flat map — so existing v1.0.9 deployments upgrade in place with \`url\` and \`transport\` recovered per server. Without this, the now-\`required: [name, url]\` schema would reject migrated specs. Covered by new unit tests.

## Summary of changes

- **API / CRDs**: new \`MCPServer\` type; CRDs in \`hiclaw-controller/config/crd/\` and \`helm/hiclaw/crds/\` (workers/managers/teams) enforce the object schema; deepcopy regenerated.
- **Controller**: \`agentconfig/mcporter.go\` rewritten; gateway / provisioner / deployer / reconcilers stripped of MCP auth.
- **Migration**: \`extractMCPServers\` handles both legacy shapes + \`url\`/\`transport\` recovery, with unit tests.
- **CLI**: \`--mcp-servers\` removed from \`create\`/\`apply\`/\`update\`.
- **Tests**: integration \`TestManagerUpdate_MCPServersChange_TriggersReauth\` replaced by \`...RewritesMcporterJSON\` (asserts the deploy request payload, zero gateway reauth calls); symmetric Worker test added.
- **Docs**: \`docs/declarative-resource-management.md\` (and zh-cn), \`docs/import-worker.md\` (and zh-cn), \`docs/k8s-native-agent-orch.md\` (and zh-cn), \`docs/design/team-worker-proposal.md\` updated to the new schema and CLI breaking change.

## Out of scope (intentional)

- \`manager/agent/skills/mcp-server-management\` and other shell scripts are **unchanged**. That skill remains the authority for local-Higress consumer binding; this PR only reshapes the controller/CRD contract.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./...\` green (unit + integration)
- [x] New unit tests: \`TestGenerateMcporterConfig_*\`, \`TestExtractMCPServers_*\`
- [x] New integration tests: \`TestManagerUpdate_MCPServersChange_RewritesMcporterJSON\`, \`TestWorkerUpdate_MCPServersChange_RewritesMcporterJSON\`
- [ ] Migration smoke test against a real v1.0.9 \`mcporter-servers.json\` snapshot
- [ ] Cloud AI-Gateway install-embedded E2E with an externally-configured MCP server


dev-vektr-ai (changes requested):
Thanks for the refactor — the direction makes sense, and I like the controller/deployer simplification plus the migration support for both wrapped and flat `mcporter-servers.json` shapes. I do think there is one API-safety issue that should be addressed before merge, plus a couple of follow-up gaps worth covering.

## Blocking: `transport` default only exists in the CRD schema, not in the Go API / JSON generation path
The new contract says `transport` defaults to `http`, but that currently relies on Kubernetes defaulting in the CRD. A few code paths can still construct `[]v1beta1.MCPServer` outside CRD admission/defaulting (HTTP server request types, tests, potential internal callers, future tooling), and in those cases the only place enforcing the default is `GenerateMcporterConfig`.

That becomes a migration / API consistency problem because:
- persisted/read-back objects may still contain empty `transport` in Go,
- non-CRD entry points can bypass schema defaulting,
- behavior is now split between schema defaulting and deploy-time normalization.

I’d strongly recommend making the default explicit in the typed API too, e.g. via kubebuilder default markers on `MCPServer.Transport` (and regenerating CRDs), or otherwise centralizing normalization close to the API boundary so every caller gets the same semantics.

## Follow-up gaps I noticed
1. **No Team coverage for the breaking `mcpServers` shape change**
   The PR updates Team CRD schema/docs, but I couldn’t find symmetric integration/unit coverage proving team leader / member flows rewrite `mcporter-servers.json` correctly after the shape change. Since Team is part of the breaking API surface, I think it deserves at least one coverage path.

2. **Migration test coverage is still narrow**
   `extractMCPServers` now supports wrapped and flat legacy JSON, which is great. But it would be good to also cover malformed / partial legacy entries (e.g. missing `url`, non-object values, empty object, unexpected `transport`) so we know upgrades fail soft in exactly the way we intend.

3. **Docs / API compatibility messaging**
   Because this is a breaking CRD change *and* the CLI `--mcp-servers` flag is removed, I think the docs/changelog should make the migration path extra explicit: users must now provide full `{name,url,transport}` records in YAML / API requests, and CLI-based worker creation no longer offers an equivalent shorthand.

If you address the defaulting/normalization issue, I’m happy with the overall direction.

johnlanni (approved):
LGTM

PR #703, opened 2026-04-24
refactor(helm): scope controller RBAC to a single namespace

Jing-ze:
## Background

The controller's RBAC footprint in the Helm chart is broader than what the controller actually does at runtime. Specifically:

- The controller only ever reads and writes inside a single namespace:
  - Manager cache is narrowed via `opts.Cache.DefaultNamespaces` when `HICLAW_K8S_NAMESPACE` is set (`hiclaw-controller/internal/app/app.go`).
  - Leader election Lease is placed in the same namespace (`LeaderElectionNamespace`).
  - HTTP handlers (`internal/server/...`) use `h.namespace` for every `Get/List/Create/Delete`.
  - The initializer (`internal/initializer/initializer.go`) uses `Config.Namespace()`.
  - The Kubernetes backend and Provisioner create Pods / Secrets / ServiceAccounts in `p.namespace`.
- The chart injects that namespace via Downward API:
  ```yaml
  - name: HICLAW_K8S_NAMESPACE
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
  ```
  so the runtime namespace is always equal to the Pod's (i.e. the release) namespace.

Despite that, the chart used to grant cross-namespace CRUD on `workers / teams / humans / managers` (and their `/status` subresources) via a ClusterRole + ClusterRoleBinding. Two issues:

- **Least-privilege violation**: the controller could touch HiClaw CRs in any namespace, but it never does. This matters when multiple tenants / releases coexist on the same cluster.
- **Conceptually inconsistent**: core resources (Pods, Secrets, Services, SAs, Events, Leases) were already scoped via a namespaced Role to the install namespace. The CR rules were the only asymmetric bit.

## Solution

Move the HiClaw CR rules into the namespaced Role and drop the `-crd` cluster objects. Keep `tokenreviews` in a ClusterRole because it is a cluster-scoped virtual resource that cannot be authorized by a Role.

Resulting controller RBAC:
- One `Role` + `RoleBinding` in the release namespace, covering:
  - `pods, pods/log, pods/exec` (core)
  - `configmaps, secrets, services, serviceaccounts, events` (core)
  - `leases` (coordination.k8s.io)
  - `workers, teams, humans, managers` + `*/status` (hiclaw.io)
- One `ClusterRole` + `ClusterRoleBinding` for `authentication.k8s.io/tokenreviews: create`.

All HiClaw CRDs are ``scope: Namespaced`` (see `hiclaw-controller/config/crd/*.yaml`), so they are authorizable via a namespaced Role.

## Verification

- `helm template release-name helm/hiclaw --show-only templates/controller/rbac.yaml --set credentials.llmApiKey=dummy --set gateway.publicURL=http://example.com` now renders:
  - A single `Role` containing both the existing core rules and the `hiclaw.io` rules.
  - The `-tokenreview` ClusterRole + ClusterRoleBinding unchanged.
  - No `-crd` ClusterRole / ClusterRoleBinding.
- `grep -r '{release}-hiclaw-controller-crd'` across the repo returns no references outside the old `rbac.yaml`; no install script, test, or doc depends on the removed names.
- Controller code is unchanged — the informer cache and all writers already target a single namespace.

## Upgrade / rollback

- `helm upgrade`: the old `{release}-hiclaw-controller-crd` ClusterRole and ClusterRoleBinding were part of the release, so Helm will prune them automatically.
- `helm rollback`: restores the old templates, which reinstates the removed cluster objects. No custom migration required.
- Non-Helm installs (if any) should manually delete `{release}-hiclaw-controller-crd` ClusterRole / ClusterRoleBinding. Functional impact: none.

## Out of scope (intentionally)

- No `controller.rbac.scope` values switch is introduced. If a future use case needs a controller that manages HiClaw CRs across namespaces, a follow-up can reintroduce a ClusterRole-based variant behind an opt-in.
- No controller code, informer scope, or leader election configuration is changed.


johnlanni (approved):
LGTM

PR #713, opened 2026-04-25
feat(hiclaw-controller):support nacos auth type for sts hiclaw

shiyiyue1102:
This pull request introduces support for the new `ai-registry` service in the access control and credential provider layers, and refactors the Nacos AI client to support multiple authentication methods and skill download workflows. It also adds configuration options for Nacos authentication and improves error handling and logging in the service layer.

**Access Control and Credential Provider Enhancements:**
- Added support for the `ai-registry` service in `AccessEntry`, including a new scope with `namespaceId` and `resources` fields. The resolver now handles `ai-registry` entries, supporting both default and custom resource patterns.
- Added comprehensive tests for `ai-registry` access entry resolution, covering default and custom resource scenarios.

**Nacos AI Client Refactoring and Authentication:**
- Refactored the Nacos AI client (renamed file and major code changes) to support multiple authentication methods, skill ZIP downloads, and extraction, with improved request preparation and token management.

**Configuration and Service Layer Updates:**
- Introduced the `NacosAuthType` configuration option, allowing selection of authentication methods for Nacos AI client connections, and propagated this config through the app and service layers.
**Logging Improvements:**
- Changed skill push failure logs from error to info level in both manager and member controllers, including the error in the log context for better observability. 

- Change-Id: I97cdd13d58be71141f4e0dbdb97a22e8679900d4

### DIFF
--- a/hiclaw-controller/api/v1beta1/types.go
+++ b/hiclaw-controller/api/v1beta1/types.go
@@ -20,7 +20,7 @@ const (
 const LabelController = "hiclaw.io/controller"
 
 // AccessEntry declares one cloud-permission grant under a logical
-// service. v1 supported services: "object-storage", "ai-gateway".
+// service. v1 supported services: "object-storage", "ai-gateway", "ai-registry".
 //
 // Scope is a schema-less JSON blob in the CR layer: it may reference
 // logical names (bucketRef: workspace, gatewayRef: default) and
--- a/hiclaw-controller/internal/accessresolver/resolver.go
+++ b/hiclaw-controller/internal/accessresolver/resolver.go
@@ -208,6 +208,12 @@ func (r *Resolver) resolveEntries(in []v1beta1.AccessEntry, tmpl templateCtx) ([
 				return nil, fmt.Errorf("entry[%d]: %w", i, err)
 			}
 			out = append(out, entry)
+		case credprovider.ServiceAIRegistry:
+			entry, err := r.resolveAIRegistry(e, tmpl)
+			if err != nil {
+				return nil, fmt.Errorf("entry[%d]: %w", i, err)
+			}
+			out = append(out, entry)
 		case "":
 			return nil, fmt.Errorf("entry[%d]: missing service", i)
 		default:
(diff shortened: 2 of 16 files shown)


johnlanni:
Reviewed the full diff — the credential abstraction is clean and well-structured. A few observations:

---

**1. `nacosUserPassCredential.Apply` uses write lock for a read-only operation**

```go
func (c *nacosUserPassCredential) Apply(req *http.Request) {
	c.mu.Lock()
	tok := c.accessToken
	c.mu.Unlock()
```

This should be `c.mu.RLock()` / `c.mu.RUnlock()` — the method only reads `c.accessToken`. Using `Lock()` serializes all concurrent `Apply` calls unnecessarily, which could become a bottleneck under high request concurrency. The `Refresh` method already uses `Lock()` for writes, so `RLock` in `Apply` would allow multiple readers while blocking only during token refresh.

---

**2. Hardcoded `DEFAULT_GROUP` in STS HMAC signing**

```go
func (c *nacosSTSCredential) Apply(req *http.Request) {
	// ...
	group := "DEFAULT_GROUP"
	signData := c.namespace + "+" + group + "+" + timestamp
```

The Spas signature always uses `DEFAULT_GROUP`, but the Nacos API calls in `nacos_ai_service.go` never explicitly pass a `group` parameter. This works today because Nacos defaults to `DEFAULT_GROUP`, but if any future API call adds a `group` query parameter with a different value, the signature won't match what the server expects. Worth a comment or making it configurable if multi-group support is on the roadmap.

---

**3. New `NacosAIClient` per skill in `PushOnDemandSkills`**

Each nacos:// skill creates a fresh client (TCP preflight + credential refresh). If a worker has multiple nacos skills, this means multiple round-trips for preflight + auth. Not a bug, but could be optimized by creating one client and reusing it across skills in the same call.

---

Everything else looks solid — the `nacosCredential` interface abstraction, the STS Spas signing implementation, the access resolver for `ai-registry`, and the skill ZIP download + OSS mirror flow. Tests are comprehensive. 👍

PR #715, opened 2026-04-26
feat(sync): local-first openclaw.json merge (bash + hermes)

johnlanni:
## Summary

Align `openclaw.json` merge behavior across the shell helper and Hermes worker: **local (Worker) is the base**, and MinIO/Manager only overlays the slices it owns.

## Behavior

- **Base**: local JSON; fields such as `tools`, `agents`, `mcp`, etc. stay from the Worker unless explicitly merged below.
- **From remote**: `models`, `gateway`, and `channels` (jq object multiply / Python deep-merge with **remote winning** leaf conflicts).
- **`channels.matrix.accessToken`**: always taken from **local** after the channel merge (Worker re-login).
- **`plugins`**: unchanged policy — `entries` merged with **local winning** shared keys; `load.paths` is the **union** of both sides.

## Files

- `shared/lib/merge-openclaw-config.sh`: jq pipeline starts from `$local`; failure no longer replaces the file with remote-only JSON.
- `hermes/src/hermes_worker/sync.py`: `_merge_openclaw_config` matches the script; `pull_all` docstring updated.
- `manager/agent/worker-agent/skills/file-sync/scripts/hiclaw-sync.sh`: comments + **fix** merged output to `${WORKSPACE}/openclaw.json` (third argument) so the result is not written only to `/tmp` and then deleted.
- `worker/scripts/worker-entrypoint.sh`: comments describing local-first merge for the periodic pull path.

## Testing

Manual jq + Python spot-checks with representative local/remote payloads; Hermes `_merge_openclaw_config` output matches `merge_openclaw_config` for the same inputs.

### DIFF
--- a/copaw/src/copaw_worker/sync.py
+++ b/copaw/src/copaw_worker/sync.py
@@ -48,40 +48,51 @@ def _deep_merge(base: dict, override: dict) -> dict:
 
 
 def _merge_openclaw_config(remote_text: str, local_text: str) -> str:
-    """Merge remote and local openclaw.json, preserving Worker additions.
+    """Merge remote and local openclaw.json (local-first, same as hermes_worker).
 
     Rules:
-      - plugins: deep merge entries (remote wins shared), union load.paths
-      - channels: deep merge (remote wins shared types, local-only preserved)
-      - channels.matrix.accessToken: local wins
-      - Everything else: remote as-is
+      - Base: local; tools, agents, mcp, and other keys not listed below stay local.
+      - models, gateway: replaced from remote when present.
+      - channels: deep merge with remote winning leaf conflicts; local-only keys kept.
+      - channels.matrix.accessToken: local wins (Worker re-login after restart).
+      - plugins.entries: deep merge with local winning on shared keys; load.paths union.
     """
     remote = json.loads(remote_text)
     local = json.loads(local_text)
-    merged = dict(remote)
+    merged: dict[str, Any] = dict(local)
 
-    # plugins: union arrays, deep merge entries — only touch fields that exist
-    r_plugins = remote.get("plugins", {})
-    l_plugins = local.get("plugins", {})
+    if remote.get("models") is not None:
+        merged["models"] = remote["models"]
+    if remote.get("gateway") is not None:
+        merged["gateway"] = remote["gateway"]
+
+    r_channels = remote.get("channels") or {}
+    l_channels = local.get("channels") or {}
+    if r_channels or l_channels:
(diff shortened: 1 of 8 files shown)


johnlanni:
## Analysis of Review Points (after source code verification)

### 1. `allowFrom` — Review point was partially incorrect

> Review claimed: "allowFrom only controls who can toggle elevated via chat commands, NOT who can execute commands"

**This is wrong.** After reading openclaw source code, `allowFrom` is a **hard gate** on all elevated functionality, including `elevatedDefault: "full"`.

Key evidence from `openclaw-hiclaw/src/auto-reply/reply/reply-elevated.ts`:
```ts
if (!rawAllow || rawAllow.length === 0) {
    return false;  // no allowFrom → elevated denied entirely
}
```

And in `get-reply-directives.ts`:
```ts
const resolvedElevatedLevel = elevatedAllowed
    ? (directives.elevatedLevel ?? elevatedDefault ?? "on")
    : "off";  // forced to "off" if not allowed
```

So `elevatedDefault: "full"` is **completely ignored** if `allowFrom` is empty or missing. On Matrix, there is no `allowFromFallback` (unlike Discord which falls back to `channels.discord.allowFrom`). Without explicit `tools.elevated.allowFrom.matrix`, elevated is effectively disabled.

**Conclusion: `allowFrom` is necessary and the current PR approach is correct.** If we wanted to simplify, we could use `["*"]` (wildcard), but the current admin+manager list is more secure.

### 2. `allowFrom.matrix` overwrite risk — Valid concern

The review point about upgrade overwrite is **correct**. The jq expression replaces the entire `matrix` array:

```jq
.tools.elevated.allowFrom |= ((. // {}) | .matrix = ["@\($admin_user):\($matrix_domain)"])
```

**Suggested fix:** Use union + dedup instead:
```jq
.tools.elevated.allowFrom.matrix = ((.tools.elevated.allowFrom.matrix // []) + ["@\($admin_user):\($matrix_domain)"] | unique)
```

### 3. jq failure handling — Acceptable as-is

No change needed. Keeping local on failure is the right tradeoff.

### 4. Hermes/Bash consistency confirmed ✅, but Copaw still inconsistent

Hermes `sync.py` and Bash `merge-openclaw-config.sh` are now aligned (local wins on plugins.entries conflicts). However, `copaw_worker/sync.py` still uses the old remote-first logic. Should this PR also fix Copaw, or is that a follow-up?

---

**Summary of needed changes:**
1. Fix `allowFrom.matrix` jq to use union instead of replace (both manager and worker upgrade paths)
2. Optionally: align `copaw_worker/sync.py` with the new local-first merge logic

PR #740, opened 2026-04-28
fix: Phantom removal of Humans from Team rooms after controller restarts

buiminhhuy:
# Fix: Human phantom removal from Team room and removal after controller restart

## Bug Description

Human with `accessibleTeams: ["team-name"]` is being kicked from Team room with message:
```
"You were removed from Team: team-name by admin 💕 Reason: removed from desired member set"
```

This happens:
1. After hiclaw-controller restart
2. When Team's `Status.TeamRoomID` is temporarily empty (team not yet provisioned)

### Issue 1: Humans not added to team invites (RBAC error)

`ProvisionTeamRooms` used `k8sClient.CoreV1().RESTClient().Get().Resource("humans")` which calls core API group (`/api/v1/...`), but RBAC only allows `hiclaw.io` group.

**Result:** "humans is forbidden" error → Humans never added to `teamInvites`

**Fix:** Use `k8sCRClient.List()` (controller-runtime client) with proper CRD API group.

### Issue 2: Human controller kicks human when TeamRoomID is empty

In `reconcileHumanRooms`, `buildDesiredHumanRooms` skips teams with empty `TeamRoomID`:
```go
if team.Status.TeamRoomID != "" {
    desired[team.Status.TeamRoomID] = struct{}{}
}
```

When Team's `TeamRoomID` is empty (not yet provisioned), the room is not in `desired`. The removal loop then kicks the human because the room "is not desired".

**Fix:** Check if room belongs to a team still in `accessibleTeams` before kicking, regardless of `TeamRoomID` status.

### 1. `provisioner.go` - Fix Human listing API
- Added `k8sCRClient client.Client` field to `Provisioner`
- Changed Human listing from REST client to controller-runtime client
- Humans with matching `accessibleTeams` are now properly added to team invites

### 2. `human_scope.go` - Add protection helper
- Added `isRoomFromAccessibleTeam()` function that returns `true` if:
  - Team is in `accessibleTeams` AND
  - `TeamRoomID` matches OR `TeamRoomID` is empty (team not provisioned yet)

### 3. `human_reconcile_rooms.go` - Fix kick protection
- Before kicking, check if room belongs to accessible team
- If API error occurs, skip kick for safety (transient error)
- Only kick if room is NOT in accessible teams

## Test Cases

1. Human with `accessibleTeams: ["team-that-doesnt-exist-yet"]` → should NOT be kicked
2. After team provisions and `TeamRoomID` is set → Human joins correctly
3. Remove team from `accessibleTeams` → Human IS kicked correctly
4. Restart controller → Human stays in team room

### DIFF
--- a/hiclaw-controller/internal/app/app.go
+++ b/hiclaw-controller/internal/app/app.go
@@ -391,6 +391,7 @@ func (a *App) initServiceLayer(_ context.Context) error {
 		OSSAdmin:          a.ossAdmin,
 		Creds:             credStore,
 		K8sClient:         a.k8sClient,
+		K8sCRClient:       a.mgr.GetClient(),
 		KubeMode:          cfg.KubeMode,
 		Namespace:         a.namespace,
 		AuthAudience:      cfg.AuthAudience,
--- a/hiclaw-controller/internal/controller/human_reconcile_rooms.go
+++ b/hiclaw-controller/internal/controller/human_reconcile_rooms.go
@@ -72,11 +72,11 @@ func (r *HumanReconciler) reconcileHumanRooms(ctx context.Context, s *humanScope
 	// Removals: in-place filter. A failed kick keeps the room so the
 	// next reconcile retries, matching pre-refactor behavior.
 	//
-	// Unlike additions which rely on status.TeamRoomID, removals are
-	// based on the spec: we only kick if the team was explicitly
-	// removed from accessibleTeams. This prevents phantom removals when
-	// a team hasn't finished provisioning yet (TeamRoomID is empty) or
-	// when status hasn't been updated.
+	// Unlike additions which rely on status.TeamRoomID/Worker.Status.RoomID,
+	// removals are based on the spec: we only kick if the team/worker was
+	// explicitly removed from accessibleTeams/AccessibleWorkers. This prevents
+	// phantom removals when a team/worker hasn't finished provisioning yet
+	// (TeamRoomID/RoomID is empty) or when status hasn't been updated.
 	kept := next[:0]
 	for _, rid := range next {
 		if _, ok := desired[rid]; ok {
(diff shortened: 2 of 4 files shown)


johnlanni:
Thanks for the fix. I found a logic bug in the removal section of `reconcileHumanRooms`:

**Bug: `continue` skips kick for teams removed from accessibleTeams**

In `human_reconcile_rooms.go`, when `findTeamNameByRoomID` returns a non-empty team name but that team is **not** in `h.Spec.AccessibleTeams`, the inner `for` loop completes without `break`ing, and then the outer `continue` statement executes — skipping the kick entirely:

```go
if teamName := findTeamNameByRoomID(ctx, r.Client, h.Namespace, rid); teamName != "" {
    for _, accessibleTeam := range h.Spec.AccessibleTeams {
        if accessibleTeam == teamName {
            kept = append(kept, rid)
            break
        }
    }
    continue  // ← always executes, even when team was NOT found in accessibleTeams
}
```

This means: if a human is in a team room, and that team is **removed** from the human's `accessibleTeams`, the human will **never be kicked** — the `continue` skips the kick logic every time.

**Suggested fix** — track whether the team was found in accessibleTeams, and only `continue` if it was:

```go
if teamName := findTeamNameByRoomID(ctx, r.Client, h.Namespace, rid); teamName != "" {
    found := false
    for _, accessibleTeam := range h.Spec.AccessibleTeams {
        if accessibleTeam == teamName {
            kept = append(kept, rid)
            found = true
            break
        }
    }
    if found {
        continue
    }
}
```

---

Aside from this, the overall approach (checking spec rather than status to avoid phantom kicks) makes sense, and the `findTeamNameByRoomID` helper and the human invitation in `ProvisionTeamRooms` look correct.

buiminhhuy:
the commit, not yet completely for human design objective. Maybe need to implement full feature as human design architecture.

## Human

A Human resource represents a real person. Upon creation, a Matrix account is automatically registered and the user is invited to the appropriate Rooms based on their permission level, enabling human-AI collaboration.

### Basic Configuration

```yaml
apiVersion: hiclaw.io/v1beta1
kind: Human
metadata:
  name: john
spec:
  displayName: John Doe
  email: john@example.com
  permissionLevel: 2
  accessibleTeams: [alpha-team]
  accessibleWorkers: []
  note: Frontend lead
```

### Field Reference

| Field | Type | Required | Default | Description |
|-------|------|----------|---------|-------------|
| `metadata.name` | string | Yes | — | User identifier, globally unique |
| `spec.displayName` | string | Yes | — | Display name |
| `spec.email` | string | No | — | Email for sending credentials |
| `spec.permissionLevel` | int | Yes | — | Permission level: 1, 2, or 3 |
| `spec.accessibleTeams` | []string | No | — | Accessible Team list (effective for L2) |
| `spec.accessibleWorkers` | []string | No | — | Accessible standalone Worker list (effective for L2/L3) |
| `spec.note` | string | No | — | Notes |

### Three-Level Permission Model

Permission levels are inclusive — higher levels include all permissions of lower levels.

**Level 1 — Admin Equivalent**

Can talk to all roles in the system, including Manager, all Team Leaders, and all Workers. `accessibleTeams` and `accessibleWorkers` fields are ignored.

Use case: CTO, VP of Engineering.

```yaml
spec:
  permissionLevel: 1
```

**Level 2 — Team-Scoped**

Can talk to specified Teams' Leaders and all their Workers, plus specified standalone Workers.

Use case: Product manager, team member.

```yaml
spec:
  permissionLevel: 2
  accessibleTeams: [alpha-team, beta-team]
  accessibleWorkers: [standalone-dev]
```

**Level 3 — Worker-Only**

Can only talk to specified Workers. `accessibleTeams` field is ignored.

Use case: External collaborator, specialized staff.

```yaml
spec:
  permissionLevel: 3
  accessibleWorkers: [alice, bob]
```

### How Permissions Work

Human permissions are enforced through two mechanisms:

1. **Room invitations**: The Human is invited to the corresponding Matrix Rooms
2. **groupAllowFrom**: The Human's Matrix ID is added to the `openclaw.json` config of the corresponding Agents — Agents only respond to @mentions from whitelisted users

| Level | groupAllowFrom Changes | Room Invitations |
|-------|----------------------|------------------|
| L1 | Added to Manager + all Leaders + all Workers | All Rooms |
| L2 | Added to specified Teams' Leaders + Workers + specified standalone Workers | Specified Team Rooms + Worker Rooms |
| L3 | Added to specified Workers | Specified Worker Rooms |

PR #752, opened 2026-05-02
Added the `ContainerManaged` field enables the management of non-containerized workers.

xiyuxiaoxiaohaijiao:
After the v1.1.0 upgrade, there is no mechanism to manage non-containerized workers. In fact, we sometimes need non-containerized workers for the following reasons:

1. Workers need to interact with GUI programs on the host machine, such as graphical browsers;

2. While containerized workers are designed for security isolation, this introduces some inconveniences. Runtimes like qwenpaw, openclaw, and Hermes have implemented built-in security measures, such as security toolbars, and offer non-containerized installation options if users install them separately (not forcing containerization);

3. Hiclaw still maintains containerized management by default (default --container-managed=true).

Finally, this pull request passed my tests, and I very much hope the official team will accept my request. Thank you very much!


johnlanni (approved):
LGTM

PR #755, opened 2026-05-03
fix(install): 拉镜像失败时增加重试 || fix(install): Add retry when pulling the image fails

googs1025:
## 问题

安装时 `docker pull` 遇到网络抖动会直接退出，导致整个安装流程中断。

## 改动

`_pull_image()` 失败后最多重试 3 次，每次间隔 5s，3 次全失败才报错退出。

---
## Question

During installation, `docker pull` will exit directly if it encounters network jitter, causing the entire installation process to be interrupted.

## Changes

`_pull_image()` will retry up to 3 times after failure, with an interval of 5s each time. It will exit with an error after all 3 failures.

### DIFF
--- a/install/hiclaw-install.sh
+++ b/install/hiclaw-install.sh
@@ -2609,7 +2609,19 @@ EOF
             fi
         fi
         log "$(msg "${_pull_key}" "${_img}")"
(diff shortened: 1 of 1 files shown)


johnlanni (approved):
LGTM

PR #757, opened 2026-05-03
docs(manager): include hermes in worker-create checklist and note install-default fallback

googs1025:
Manager was only mentioning openclaw and copaw when asking admin which worker to create — hermes was invisible even though the runtime is supported.

Adds a "Before You Create" checklist that covers all three runtimes (openclaw / copaw / hermes) and a reminder that the install-time `HICLAW_DEFAULT_WORKER_RUNTIME` is just a fallback, so Manager should always present the three options explicitly instead of silently using the default.


googs1025:
只有提到 openclaw 跟 copaw

image

johnlanni (approved):
LGTM

Issue #634, opened 2026-04-16
worker的SOUL文件总是被自动修改成系统自带模板


shutong1026:
我的worker团队工作、对话质量不稳定，一开始我不知道什么原因，我观察到worker的对话不是按我规定的SOUL文件进行的，于是查了一下，发现SOUL被重置，然后我又复制一遍，worker会按照我的要求进行工作，但是等下一次启动时又不行了，我再打开SOUL发现又被重置了。问了一下manager，给我的回答如图。

<img width="1780" height="1218" alt="Image" src="https://github.com/user-attachments/assets/7297f592-6dd2-4f67-8d83-223249330e8f" />

---
The quality of my worker team's work and dialogue was unstable. I didn't know the reason at first. I observed that the worker's dialogue was not conducted according to the SOUL file specified by me, so I checked and found that SOUL had been reset. Then I copied it again. The worker would work according to my requirements, but it didn't work the next time it was started. I opened SOUL again and found that it had been reset again. I asked the manager and he gave me the answer as shown below.

<img width="1780" height="1218" alt="Image" src="https://github.com/user-attachments/assets/7297f592-6dd2-4f67-8d83-223249330e8f" />

shiyiyue1102:
There are related changes on latest main in #826 and #848. Package/base files and Worker/Team Leader `SOUL.md` are now handled as seed-only: default content is written on first creation, and later reconciles should not overwrite an existing runtime file.

Please verify this with a latest main/image that includes both PRs. If `SOUL.md` is still reset to the system template after restart or reconcile, please share the trigger path (Worker / Team Leader / standalone), runtime type, and the before/after timestamp plus content source for `SOUL.md`.

shiyiyue1102:
Following up here: could you confirm whether this still reproduces with a latest main/image that includes #826 and #848? If it does, please share the trigger path (Worker / Team Leader / standalone), runtime type, and the before/after timestamp plus content source for `SOUL.md`. If there is no further update within 3 days, we may temporarily close this issue until more information is available.

shiyiyue1102:
Closing this temporarily because there has been no further update after the follow-up request. Please feel free to reopen or comment with the requested details if the issue still reproduces.

PR #763, opened 2026-05-05
fix(controller): honor TeamWorkerSpec.Runtime and HICLAW_DEFAULT_WORKER_RUNTIME for team workers

Sunrisea:
## Summary

Team workers have silently ignored `Team.Spec.Workers[].runtime` and `HICLAW_DEFAULT_WORKER_RUNTIME` since v1.1.0 — every team worker was forced to run on `copaw` regardless of configuration.

This PR restores the pre-v1.1.0 resolution chain that standalone `Worker` CRs already use, while keeping the `team_leader` role on `copaw` (which is an intentional design constraint, see below).

Regression introduced in #666.

## Root Cause

Two places hardcoded `Runtime: "copaw"` when projecting a `TeamWorkerSpec` into the downstream model:

1. **`internal/controller/team_controller.go`** — `teamWorkerSpecToWorkerSpec()` overwrote the per-member runtime during reconcile, so the container backend always received `copaw` as the effective runtime for team workers.
2. **`internal/server/resource_handler.go`** — `teamMemberToResponse()` also hardcoded `Runtime: "copaw"` in the API response path, so `hiclaw get worker` and Manager skills reported `copaw` for every team worker regardless of what was actually running.

The comment above `teamWorkerSpecToWorkerSpec()` already described the intended behavior ("policy merge"), but the implementation silently dropped the `Runtime` field. The `TeamWorkerSpec.Runtime` field has existed in the CRD since the Teams feature landed, so users had no way to know their `runtime: hermes` / `runtime: openclaw` declarations were being ignored.

## Fix

- `teamWorkerSpecToWorkerSpec()`: pass `w.Runtime` through verbatim. An empty string is intentional and means "no per-member preference" — `backend.ResolveRuntime(spec.Runtime, fallback)` resolves it against `TeamReconciler.DefaultRuntime` (sourced from `HICLAW_DEFAULT_WORKER_RUNTIME` via `MemberDeps.DefaultRuntime` → `backend.CreateRequest.RuntimeFallback`), falling back to `RuntimeOpenClaw` if neither is set. This matches the resolution chain used by standalone `Worker` CRs.
- `teamMemberToResponse()`: split the branch by role — `team_leader` stays `copaw` (mirrors `leaderWorkerSpec()`), `worker` passes `wk.Runtime` through (matches `ListWorkers` / `GetWorker`, which return `w.Spec.Runtime` verbatim for standalone workers).
- Leader runtime is **intentionally left hardcoded to `copaw`**: `LeaderSpec` has no `runtime` field today, and only the `copaw` `team-leader-agent` template ships in the image. Changing that is out of scope for this regression fix.

## Upgrade Impact

- **New or edited Teams**: pick up the fix immediately on the next reconcile.
- **Existing team pods**: keep their current runtime until the user edits the Team spec or deletes the pod. This is safe because `hashMemberSourceSpec()` hashes the source `TeamWorkerSpec` (which this patch does not change), so `SpecChanged` stays `false` for untouched members and no pod churn is triggered.
- No CRD changes. No API breaking changes. Controller behavior change only.

## Tests

- `TestBuildDesiredMembers_LeaderAndWorkers`: assertion split by role — leader must be `copaw`, worker must be `""` (pass-through from the fixture, which leaves `Runtime` unset).
- `TestTeamWorkerSpecToWorkerSpec_RuntimePassthrough` (new): covers four cases — `hermes`, `openclaw`, `copaw`, and empty — to lock in verbatim pass-through and prevent future regressions.
- `go build ./...` and `go test ./internal/controller/... ./internal/server/...` pass locally.

## Changelog

```
- fix(controller): honor `TeamWorkerSpec.Runtime` and `HICLAW_DEFAULT_WORKER_RUNTIME` for team workers in `teamWorkerSpecToWorkerSpec()` and `teamMemberToResponse()` (regression from #666 that silently forced team workers to `copaw`). Leader runtime remains `copaw` by design. Existing team pods keep their current runtime until the user edits the Team spec or deletes the pod; new/edited teams pick up the fix immediately
```

## References

- Regression source: #666
- Related work: #690 (same family — `spec.env` dropped in the same projection layer)
- Touches the pattern discussed in #634 (team member fields silently overwritten)

---

cc @Jing-ze @googs1025 — you've been in adjacent PRs on this projection layer, would appreciate a look.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -8,4 +8,5 @@ Record image-affecting changes to `manager/`, `worker/`, `openclaw-base/` here b
 - fix(manager): document runtime-aware Worker dispatch (avoid @worker text in admin DM only); update task-management references, AGENTS.md, HEARTBEAT.md, channel-management skill
 - fix(manager): separate runtime-specific AGENTS/HEARTBEAT for OpenClaw vs CoPaw; remove cross-runtime references from manager agent docs
 - refactor(api)!: restructure `spec.mcpServers` on Worker/Manager/Team CRDs to `[]{name,url,transport}`; drop controller-side MCP gateway authorization; `mcporter-servers.json` is written from the CRD (see `docs/declarative-resource-management.md`)
+- fix(controller): honor `TeamWorkerSpec.Runtime` and `HICLAW_DEFAULT_WORKER_RUNTIME` for team workers in `teamWorkerSpecToWorkerSpec()` and `teamMemberToResponse()` (regression from #666 that silently forced team workers to `copaw`). Leader runtime remains `copaw` by design. Existing team pods keep their current runtime until the user edits the Team spec or deletes the pod; new/edited teams pick up the fix immediately ([af52054](https://github.com/agentscope-ai/HiClaw/commit/af52054))
 
--- a/hiclaw-controller/internal/controller/team_controller.go
+++ b/hiclaw-controller/internal/controller/team_controller.go
@@ -858,6 +858,13 @@ func leaderWorkerSpec(t *v1beta1.Team) v1beta1.WorkerSpec {
 //   - leader is always on the worker's groupAllow
 //   - team admin (if any) is on the worker's groupAllow
 //   - if Team.Spec.PeerMentions is true (default), all peers are groupAllow too
+//
+// Runtime is passed through from TeamWorkerSpec. An empty value is intentional
+// and means "no per-member preference" — backend.ResolveRuntime resolves it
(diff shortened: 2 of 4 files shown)


googs1025 (commented):
overall LGTM.  

Nice regression fix. Root cause is correctly localized to two hardcoded "copaw" — teamWorkerSpecToWorkerSpec() (reconcile path) and teamMemberToResponse() (API response path)

PR #778, opened 2026-05-07
feat(controller): improve credential to minio for non-container worke…

xiyuxiaoxiaohaijiao:
The worker-manager's skills still retain the ability to read environment variables containing Minio credentials. However, since version 1.1.0, some credentials are only stored in the /data/worker-creds directory within the hiclaw-controller container, and the manager cannot directly access them. Currently, some credentials are synchronized to the hiclaw-storage/agents/<worker-name>/credentials directory within Minio. This is safe because this directory is only accessible to the manager and its associated worker, and it is also deleted when a worker is deleted.
  Finally, this pull request passed my tests, and I very much hope the official team will accept my request. Thank you very much!


flystar32:
1. Which non-containerized worker types are currently supported, and how do they integrate?

2. How do the workers retrieve these sensitive credentials?

3. Regarding writing sensitive credentials to MinIO: for users that don't need this feature, enabling it by default might not be desirable. Should we add a feature flag?

I don't think this is really needed right now.

xiyuxiaoxiaohaijiao:
1: Currently, I have deployed a non-containerized worker of type qwenpaw on my own machine (this was already implemented in hiclaw versions greater than or equal to 1.0.4 and less than 1.1.0). Here is the article explaining the local host method worker at that time: https://mp.weixin.qq.com/s/JG4QiKsUys6j1rXL1VAuMw. However, when the version came to v1.1.0, because the worker management was separated into the hiclaw-controller container, the original worker-manager skills no longer apply to the --remote parameter (local host worker mode). To ensure my local workers could continue working (without forcing the use of containers), I previously submitted a PR (#752).

2: In my previous PR (#752), I added the `--container-managed` parameter to the hiclaw CLI (default `--container-managed=true`). However, the built-in worker-manager skills haven't yet been adapted for the non-containerized workers in the new version (the `--remote` parameter and the `get-worker-install-cmd.sh` script are still present). The `get-worker-install-cmd.sh` script reads some credentials from environment variables for human administrators to manually execute on the target machine. However, since version 1.1.0, some credentials are only stored in the `/data/worker-creds` directory within the hiclaw-controller container, and the manager cannot directly access them. Currently, some credentials are synchronized to the `hiclaw-storage/agents//credentials` directory within Minio. 3: I would like to gradually restore HiClaw's support for non-containerized workers of the qwenpaw type (Workers need to interact with GUI programs on the host machine, such as graphical browsers), but I don't know what the official thoughts are on non-containerized workers in the future. I look forward to the official opinion on non-containerized workers. Thank you very much! @flystar32

flystar32:
@xiyuxiaoxiaohaijiao 
Non-containerized workers are a significant part HiClaw. However, I have some concerns about saving secrets directly to MinIO. Maybe we could add a feature flag first, so this behavior is only enabled when explicitly turned on.

flystar32:
@xiyuxiaoxiaohaijiao  Hello, and welcome to this DingTalk group for communicating with HiClaw contributors!
钉钉群号： 185280007541

xiyuxiaoxiaohaijiao:
@flystar32 Yes, writing credentials to minio is not necessary. I would be very happy to join the DingTalk group for communicating with HiClaw contributors.

PR #792, opened 2026-05-11
fix: preserve AI route provider config on Manager restart

### LINKED ISSUE #507: 清理文件锁，或者清理worker都会导致api提供商的配置从自定义URL到官方OpenAI的 URL，导致大模型请求超时 || Clearing file locks or cleaning workers will cause the api provider's configuration to change from a custom URL to the official OpenAI URL, causing large model request timeouts.
### Bug Description

清理文件锁，或者清理worker都会导致api提供商的配置从自定义URL到官方OpenAI的 URL，导致大模型请求超时

### Steps to Reproduce

1. 配置好api提供商，阿里云的coding-plan地址。
2.让manager清理文件锁，或者清理worker。
3. manager超时无响应。
4. 发现api提供商的配置从自定义URL到官方OpenAI的 URL，导致大模型请求超时

### AI Analysis

问题根因找到了！                                                                                                                                          
  关键问题代码                                                                                                                                              
  setup-higress.sh 第 262 行：                                                                                                                              
                                                                                                                                                            
  | .upstreams[0].provider = "'"${LLM_PROVIDER}"'"  # 强制覆盖 provider！                                                                                   
                                                                                                                                                            
  问题流程                                                                                                                                                  
  1. 用户使用 openai-compat provider（自定义 URL）                                                                                                          
  2. 清理文件锁或清理 worker 触发 Manager 重启（PID 从 3209 变为 6358）                                                                                     
  3. Manager 启动时运行 setup-higress.sh                                                                                                                    
  4. 第 262 行强制将 provider 改为 ${LLM_PROVIDER}（默认 qwen）                                                                                             
  5. AI route 被路由到 qwen provider，但用户没有配置 qwen API key                                                                                           
  6. 所有 LLM 请求超时/失败                                                                                                                                 
                                                                                                                                                            
  日志证据                                                                                                                                                  
  05:40:39 - session file locked (PID=3209)  ← 第一次锁问题                                                                                                 
  05:42:42 - session file locked (PID=6358)  ← PID 变化 = Manager 重启了！                                                                                  
  05:47:37 - HTTP 503 错误                   ← LLM 服务不可用                                                                                               
                                                                                                                                                            
  根本原因                                                                                                                                                  
  setup-higress.sh 每次 Manager 启动都会执行：                                                                                                              
                                                                                                                                                            
  LLM_PROVIDER="${HICLAW_LLM_PROVIDER:-qwen}"  # 默认 qwen                                                                                                  
                                                                                                                                                            
  # 无条件覆盖 AI route 的 provider                                                                                                                         
  patched=$(echo "${existing_route_resp}" | jq '                                                                                                            
      .data                                                                                                                                                 
      | .upstreams[0].provider = "'"${LLM_PROVIDER}"'"  # 问题在这！                                                                                        
      ...                                                                                                                                                   
  ')                                                                                                                                                        
                                                                                                                                                            
  即使配置已经正确，也会强制改为 HICLAW_LLM_PROVIDER 的值。如果用户使用自定义 provider 但没有设置这个环境变量，配置就会被覆盖为 qwen。                      
                                                                                                                                                            
  修复方案                                                                                                                                                  
  只保留现有的 provider 配置，不要强制覆盖：                                                                                                                
                                                                                                                                                            
  # 修复：保留现有 provider，只在首次创建时设置                                                                                                             
  patched=$(echo "${existing_route_resp}" | jq --argjson domains "${AI_ROUTE_DOMAINS}" '                                                                    
      .data                                                                                                                                                 
      | .domains = $domains                                                                                                                                 
      # 移除这行: | .upstreams[0].provider = "..."                                                                                                          
      | .headerControl.enabled = true                                                                                                                       
      ...                                                                                                                                                   
  ')                                                                                                                                                        
                                                                                                                                                            
  或者只在 provider 不存在时才设置默认值。                                     

### Relevant Logs

```shell

```

### Component

Manager Agent

### Version / Commit

_No response_
<!--This is a translation content dividing line, the content below is generated by machine, please do not modify the content below-->
---
### Bug Description

Clearing file locks or cleaning workers will cause the api provider's configuration to change from a custom URL to the official OpenAI URL, causing large model request timeouts.

### Steps to Reproduce

1. Configure the API provider and Alibaba Cloud's coding-plan address.
2. Let the manager clear the file lock or clean the worker.
3. The manager times out and does not respond.
4. It was found that the configuration of the api provider changed from a custom URL to the official OpenAI URL, causing large model request to time out.

### AI Analysis

The root cause of the problem was found!                                                                                                                                          
  key question code
  setup-higress.sh line 262:
                                                                                                                                                            
  | .upstreams[0].provider = "'"${LLM_PROVIDER}"'" # Force coverage of provider!                                                                                   
                                                                                                                                                            
  problem flow
  1. User uses openai-compat provider (custom URL)
  2. Clear file locks or clean workers to trigger a Manager restart (PID changes from 3209 to 6358)
  3. Run setup-higress.sh when Manager starts
  4. Line 262 forces the provider to be changed to ${LLM_PROVIDER} (default qwen)
  5. AI route is routed to qwen provider, but the user does not configure qwen API key
  6. All LLM requests time out/failed
                                                                                                                                                            
  Log evidence
  05:40:39 - session file locked (PID=3209) ← First lock problem
  05:42:42 - session file locked (PID=6358) ← PID change = Manager restarted!                                                                                  
  05:47:37 - HTTP 503 Error ← LLM service unavailable
                                                                                                                                                            
  root cause
  setup-higress.sh will be executed every time Manager starts:
                                                                                                                                                            
  LLM_PROVIDER="${HICLAW_LLM_PROVIDER:-qwen}" #Default qwen
                                                                                                                                                            
  # Unconditionally override the AI route provider
  patched=$(echo "${existing_route_resp}" | jq '
      .data
      | .upstreams[0].provider = "'"${LLM_PROVIDER}"'" # Here's the problem!                                                                                        
      ...
  ')
                                                                                                                                                            
  Even if the configuration is correct, it will be forced to change to the value of HICLAW_LLM_PROVIDER. If the user uses a custom provider but does not set this environment variable, the configuration will be overwritten to qwen.                      
                                                                                                                                                            
  Fix
  Just keep the existing provider configuration, don't force an overwrite:
                                                                                                                                                            
  # Fix: Keep existing providers and only set them when first created
  patched=$(echo "${existing_route_resp}" | jq --argjson domains "${AI_ROUTE_DOMAINS}" '
      .data
      | .domains = $domains
      # Remove this line: | .upstreams[0].provider = "..."
      | .headerControl.enabled = true
      ...
  ')
                                                                                                                                                            
  Or only set the default value if the provider does not exist.                                     

### Relevant Logs

```shell

```

### Component

Manager Agent

### Version/Commit

_No response_

nillikechatchat:
## Summary
Fix for Issue #507: Clear file lock or clean worker causes API provider configuration to change from custom URL to official OpenAI URL.

### Problem
When mc mirror completes file synchronization, there is a race condition where file reads can occur before the file is fully written to disk. This causes `UnicodeDecodeError` when reading multi-byte UTF-8 characters at the end of files.

### Root Cause
`setup-higress.sh` line 262 unconditionally overwrites `upstreams[0].provider` with the current `LLM_PROVIDER` env var on every Manager restart. When `HICLAW_LLM_PROVIDER` is not set, it defaults to `qwen`, overwriting any custom provider configuration set via Higress Console.

### Solution
Remove the unconditional provider override in the jq patch. The AI Gateway route already stores the correct provider - we should only update domains and headerControl on restarts, preserving the provider setting.

### Files Changed
- `manager/scripts/init/setup-higress.sh` - Removed provider override line

### Verification
After this fix:
1. User configures custom provider (e.g., openai-compat with custom URL)
2. Manager restarts (file lock cleared or worker cleaned)
3. Provider configuration is preserved correctly

Fixes #507

### DIFF
--- a/manager/scripts/init/setup-higress.sh
+++ b/manager/scripts/init/setup-higress.sh
@@ -259,7 +259,6 @@ if [ -n "${HICLAW_LLM_API_KEY}" ]; then
         # Preserve all other fields (especially authConfig.allowedConsumers and version).
         patched=$(echo "${existing_route_resp}" | jq --argjson domains "${AI_ROUTE_DOMAINS}" '
             .data
-            | .upstreams[0].provider = "'"${LLM_PROVIDER}"'"
             | .domains = $domains
             | .headerControl.enabled = true
             | .headerControl.request.add = [{"key":"user-agent","value":"HiClaw/'"${HICLAW_VERSION}"'"}]


flystar32:
"After removing this line of code, will users still be able to update the provider via environment variables? For example, if a user updates the LLM Provider during a HiClaw version upgrade, will it succeed?"

删除了这行代码之后，用户能够通过环境变量更新 provider ？比如用户在升级 HiClaw 版本的过程中更新了 LLM Provider，是否能成功？

nillikechatchat:
> "After removing this line of code, will users still be able to update the provider via environment variables? For example, if a user updates the LLM Provider during a HiClaw version upgrade, will it succeed?"
> 
> 删除了这行代码之后，用户能够通过环境变量更新 provider ？比如用户在升级 HiClaw 版本的过程中更新了 LLM Provider，是否能成功？

升级的过程中？还是升级的时候修改服务商

flystar32:
> > "After removing this line of code, will users still be able to update the provider via environment variables? For example, if a user updates the LLM Provider during a HiClaw version upgrade, will it succeed?"
> > 删除了这行代码之后，用户能够通过环境变量更新 provider ？比如用户在升级 HiClaw 版本的过程中更新了 LLM Provider，是否能成功？
> 
> 升级的时候修改服务商

johnlanni:
I agree this fixes the restart overwrite problem described in #507: once `default-ai-route` already exists, Manager restart should not blindly reset `upstreams[0].provider` back to `${LLM_PROVIDER}` and accidentally break a provider that the user configured from the Higress Console.

However, I think this change also introduces a behavior change that we should clarify before merging.

After removing the provider assignment, an existing `default-ai-route` will no longer follow `HICLAW_LLM_PROVIDER` on subsequent Manager restarts/upgrades. That means if a user intentionally changes the provider through environment variables during an upgrade or reconfiguration, the route may keep using the old provider forever unless the user manually edits the route in Higress Console or deletes/recreates it.

So the source of truth changes from:

- old behavior: `HICLAW_LLM_PROVIDER` wins on every Manager startup
- new behavior: existing Higress route provider wins; env only affects initial route creation

I think preserving user Console changes is the right direction, but we need to make this semantic explicit and avoid surprising users who expect env-based provider changes to take effect during upgrade.

Possible ways to handle this:

1. Document the new behavior clearly: env controls initial route creation only; existing route config is preserved on restart.
2. Add an explicit force-sync switch, for example `HICLAW_FORCE_SYNC_LLM_PROVIDER=true`, for users who intentionally want env to overwrite the route provider.
3. Only skip the overwrite when the existing provider appears to be user-customized, and still allow controlled migration from old default values.

Also, the update log message still says `provider=${LLM_PROVIDER}` even though this PR no longer updates the provider, which may be misleading during troubleshooting.

For now I would not approve this yet until the intended provider source-of-truth behavior is confirmed.

shiyiyue1102:
Thank you for raising the restart overwrite problem. The current setup flow intentionally treats the configured AGENTTEAMS LLM provider as the source of truth during controlled restart or upgrade, while preserving the rest of the existing route. Simply removing the provider assignment would make environment-based provider changes silently stop working, which was the unresolved behavior concern in this PR. The surrounding script and variable contract have also changed. We are closing this implementation as outdated; a future change would need an explicit preserve or force-sync policy and tests for both paths.

PR #798, opened 2026-05-12
fix(install): make error() multi-line safe by splitting exit into die()

johnlanni:
## Summary
- Install scripts' `error()` printed its message and then immediately `exit 1`, which silently swallowed every subsequent line in multi-line error blocks (e.g. the embedded-image-not-found block in `hiclaw-install.sh` printed only its first line before terminating).
- Split into `error()` (prints only, multi-line safe) and `die()` (prints + exits). Single-line callers that relied on the exit are switched to `die()`; the existing multi-line blocks already had an explicit `exit 1` after them and now correctly print all lines first.
- Same change applied to `install/hiclaw-apply.sh`.

## Test plan
- [x] `bash -n install/hiclaw-install.sh` and `bash -n install/hiclaw-apply.sh` pass
- [x] Trigger a multi-line error path (e.g. set `HICLAW_VERSION` to a tag with no embedded image) and confirm all lines render before the script exits
- [x] Trigger a single-line error path (e.g. `hiclaw-apply.sh` with no manager container) and confirm exit-1 behavior is unchanged

### DIFF
--- a/install/hiclaw-apply.sh
+++ b/install/hiclaw-apply.sh
@@ -21,6 +21,10 @@ log() {
 
 error() {
     echo -e "\033[31m[HiClaw Apply ERROR]\033[0m $1" >&2
+}
+
+die() {
+    error "$1"
     exit 1
 }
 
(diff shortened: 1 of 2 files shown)


CH3CHO (approved):
LGTM

PR #807, opened 2026-05-13
fix(copaw): use display name instead of MXID in mention body (#174)

### LINKED ISSUE #174: 助手和管理员之间@对方没反应 || There is no response between the assistant and the administrator @ each other.
助手和管理员之间@对方的时候都会带上:matrix-local.hiclaw.io:18080，导致没反应，不会回复信息

---
When the assistant and the administrator @ each other, they will bring: matrix-local.hiclaw.io:18080, resulting in no response and no reply to the message

grllll:
## Summary                                                                                                                                         
  - Fixes #174                                                                                                                                       
  - `_apply_mention()` now resolves display names for the plain-text body field instead of inserting raw MXIDs (`@user:server`)                      
  - The full MXID is preserved in `matrix.to` anchors (`formatted_body`) and `m.mentions` metadata                                                   
  - Updated tests to assert display-name-based body content                                                                                          
                                                                                                                                                     
  ## Test plan                                                                                                                                       
  - [x] All 8 mention tests pass (`pytest copaw/tests/test_channel_mention.py`)                                                                    
  - [x] Verified end-to-end in local HiClaw instance: body shows `admin` instead of `@admin:matrix-local.hiclaw.io:18080`                            
  - [x] Confirmed mention detection still works (agents respond to @-mentions)

### DIFF
--- a/copaw/src/matrix/channel.py
+++ b/copaw/src/matrix/channel.py
@@ -1992,10 +1992,12 @@ class MatrixChannel(BaseChannel):
             html_body = html.escape(body).replace("\n", "<br>\n")
 
         for mxid in targets:
-            if mxid not in body:
-                body = f"{mxid} {body}" if body else mxid
-            mxid_enc = urllib.parse.quote(mxid, safe="")
             display = self._resolve_display_name(mxid, room_id) or mxid
+            if mxid in body:
+                body = body.replace(mxid, display, 1)
+            elif display not in body:
+                body = f"{display} {body}" if body else display
+            mxid_enc = urllib.parse.quote(mxid, safe="")
             anchor = (
                 f'<a href="https://matrix.to/#/{mxid_enc}">'
                 f"{html.escape(display)}</a>"
--- a/copaw/tests/test_channel_mention.py
+++ b/copaw/tests/test_channel_mention.py
@@ -42,7 +42,9 @@ def test_apply_mention_explicit_user_ids_prefixes_body_and_adds_anchor():
     )
 
     assert content["m.mentions"] == {"user_ids": ["@worker-a:hs.local"]}
-    assert content["body"].startswith("@worker-a:hs.local ")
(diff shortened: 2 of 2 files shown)


grllll:
## Verification                                                                                                                 
                                                                                                                                                     
  Screenshot after the fix — both mention styles now work:                                                                                           
                                                                                                                                                     
  1. `@manager` (display name mention) ✅                                                                                                            
  2. `@manager:matrix-local.hiclaw.io:18080` (full MXID mention) ✅                                                                                
                                                                                                                                                     
  Before the fix, only the full MXID style (2) triggered a response.
<img width="1468" height="710" alt="截屏2026-05-14 10 56 59" src="https://github.com/user-attachments/assets/563a590b-ab63-4c52-81fc-90af55497ebd" />

johnlanni:
I reviewed this change and the direction looks right. The issue is not just cosmetic: exposing the full MXID in the plain-text body makes mentions noisy and confusing, and it can also make Manager/Worker interactions feel broken when users naturally expect display-name style mentions.

The fix keeps the important Matrix mention layers intact: `m.mentions.user_ids` still carries the full MXID, and `formatted_body` still uses a `matrix.to` anchor pointing to the correct user. Only the plain-text `body` is changed to use the resolved display name / localpart fallback, which is the right tradeoff for readability while preserving protocol-level targeting.

The updated tests cover explicit mention IDs, fallback sender mentions, rewriting existing MXIDs, and media-event formatted body synthesis. Together with the reported local E2E verification, this gives me enough confidence that both display-name mentions and full-MXID mentions continue to work.

LGTM.

maplefeng-a (approved):
LGTM — clean, well-scoped fix. Display name resolution logic is correct, tests cover the key cases, and m.mentions/formatted_body remain untouched. Minor theoretical edge cases (same display name across targets, substring false positives) are acceptable trade-offs for real-world usage.

PR #810, opened 2026-05-15
fix(auth): cap and sweep TokenReview cache

googs1025:
TokenReview 的 cache map 之前没有上限，token 持续变化时会一直涨（TODO 也提到过）。

加一个 1000 entry 的 cap，插入时先清过期项再淘汰最旧的；同时起一个 1min 的后台 sweeper，跟 app ctx 一起退出。

---
The cache map of TokenReview had no upper limit before, and it will keep increasing when the token continues to change (TODO also mentioned).

Add a cap of 1000 entries. When inserting, first clear the expired items and then eliminate the oldest ones; at the same time, set up a 1min background sweeper and exit together with the app ctx.

### DIFF
--- a/hiclaw-controller/internal/app/app.go
+++ b/hiclaw-controller/internal/app/app.go
@@ -335,7 +335,7 @@ func (a *App) initFieldIndexers(ctx context.Context) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


johnlanni (approved):
LGTM

PR #811, opened 2026-05-16
chore(controller): graceful shutdown for HTTP server and background goroutines

googs1025:
之前 SIGTERM 之后 mgr.Start 返回 main 直接退出，HTTP server 还卡在 ListenAndServe 里没人 Shutdown，后台两个 goroutine（leader-elected init、file watcher）也没人 Wait。

改成用 WaitGroup 跟踪所有后台 goroutine，main 收到信号后再用 10s 超时 ctx 调 `app.Stop()`：先 `server.Shutdown` 优雅排空请求，再 `wg.Wait` 等后台退出。HTTP server 改用 `http.Server` 结构体，`Start` 把 `ErrServerClosed` 吞掉避免误报。带 server 生命周期单测。

---
After SIGTERM, mgr.Start returns to main and exits directly. The HTTP server is still stuck in ListenAndServe and no one is shutting down. There are two goroutines (leader-elected init, file watcher) in the background and no one is waiting.

Change to use WaitGroup to track all background goroutines. After main receives the signal, use 10s timeout ctx to call `app.Stop()`: first `server.Shutdown` to drain the request gracefully, and then `wg.Wait` to wait for the background to exit. HTTP server uses the `http.Server` structure instead, and `Start` swallows `ErrServerClosed` to avoid false positives. Single test with server life cycle.

### DIFF
--- a/hiclaw-controller/cmd/controller/main.go
+++ b/hiclaw-controller/cmd/controller/main.go
@@ -6,12 +6,15 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/hiclaw/hiclaw-controller/internal/app"
 	"github.com/hiclaw/hiclaw-controller/internal/config"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
+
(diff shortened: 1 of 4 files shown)


johnlanni (approved):
LGTM

PR #817, opened 2026-05-18
feat(install): optimize container runtime socket detection for rootless podman

NaCodermer:
### Description
This PR fixes and optimizes the `detect_socket` function in the installation script to better support **Rootless Podman** environments (especially on modern distributions like Debian 13) and multi-runtime systems.

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Details
1. **Flattened Check Structure:** Replaced the deeply nested `if-else` blocks with sequential independent `if` statements. This prevents the `command -v docker` wrapper/alias from shadowing the actual `podman` logic on systems where both exist.
2. **Early Returns Added:** Every successful socket match now triggers an immediate `return 0`. This completely eliminates the potential multi-line output bug where multiple runtime socket strings could accidentally concat into one variable.
3. **Active Fallback Activation:** Since rootless Podman sockets are often lazy-loaded on demand (meaning the `.sock` file won't exist upon fresh boots), an active fallback was introduced to try to run `systemctl --user enable --now podman.socket` before doing the final file check.
4. **User Transparency (Stderr Notification):** Because `systemctl --user enable --now` mutates system state, an informative notice is printed to `stderr`. This alerts the user to the write operation without polluting the `stdout` buffer used for variable assignment.

### DIFF
--- a/install/hiclaw-install.sh
+++ b/install/hiclaw-install.sh
@@ -1403,22 +1403,41 @@ generate_key() {
 
 # Detect container runtime socket on the host
 detect_socket() {
-    # Check for Podman
+    # 1. First priority: Check if rootless Podman socket is ALREADY running/existing
+    if [ -S "${XDG_RUNTIME_DIR}/podman/podman.sock" ]; then
+        echo "${XDG_RUNTIME_DIR}/podman/podman.sock"
+        return 0
+    fi
+
+    # 2. Check for system-wide rootful Podman socket
     if [ -S "/run/podman/podman.sock" ]; then
         echo "/run/podman/podman.sock"
-    # Check for standard Docker socket
-    elif [ -S "/var/run/docker.sock" ]; then
+        return 0
+    fi
+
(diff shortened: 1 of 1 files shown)


flystar32:
"systemctl --user enable --now" is a write operation that enables the podman.socket service. 
Silently executing it in an install script may exceed user expectations. Consider printing an info message (e.g "Enabling rootless Podman socket via systemd...") before the call so the user is aware of the state change.

NaCodermer:
Good point! I've added an informational message before the systemctl call to ensure full transparency. To avoid polluting the stdout buffer captured by the caller (CONTAINER_SOCK=$(detect_socket)), the notice is safely redirected to stderr (>&2).

flystar32 (approved):
LGTM

NaCodermer:
Thank you so much for your time and the approval, @flystar32! Excited to see this merged. Let me know if there's anything else needed on my end.

PR #821, opened 2026-05-19
fix(helm): disable Tuwunel default displayname suffix

shiyiyue1102:
## What changed

- Exposes `matrix.tuwunel.newUserDisplayNameSuffix` in Helm values.
- Passes `CONDUWUIT_NEW_USER_DISPLAYNAME_SUFFIX` to managed Tuwunel.
- Sets the same default in the local manager startup script.

## Why

Tuwunel appends a built-in display name suffix by default. HiClaw identities should keep Matrix display names equal to the configured Manager, Worker, and Human names.

## Compatibility

Default value is empty, which disables the suffix. Existing deployments that want a suffix can set `matrix.tuwunel.newUserDisplayNameSuffix` explicitly.

## Validation

### DIFF
--- a/helm/hiclaw/templates/matrix/tuwunel-statefulset.yaml
+++ b/helm/hiclaw/templates/matrix/tuwunel-statefulset.yaml
@@ -44,6 +44,10 @@ spec:
               value: "true"
             - name: CONDUWUIT_DB_POOL_WORKERS_LIMIT
               value: "32"
+            # Keep Matrix profile display names equal to HiClaw identities by
+            # default; Tuwunel otherwise appends its built-in "💕" suffix.
+            - name: CONDUWUIT_NEW_USER_DISPLAYNAME_SUFFIX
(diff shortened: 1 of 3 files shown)


johnlanni:
LGTM from the compatibility/generalization pass.

I checked both Helm and embedded startup paths, and both now set the display name suffix explicitly:

```yaml
- name: CONDUWUIT_NEW_USER_DISPLAYNAME_SUFFIX
  value: {{ .Values.matrix.tuwunel.newUserDisplayNameSuffix | default "" | quote }}
```

```sh
export CONDUWUIT_NEW_USER_DISPLAYNAME_SUFFIX="${CONDUWUIT_NEW_USER_DISPLAYNAME_SUFFIX:-}"
```

This looks straightforward and compatible with embedded mode. I don't see hacky logic here.

johnlanni (approved):
LGTM

PR #822, opened 2026-05-19
fix(bootstrap): propagate observability and stream timeout env

shiyiyue1102:
## What changed

- Passes `HICLAW_AI_STREAM_IDLE_TIMEOUT_SECONDS` from the local installer into the controller container.
- Passes CMS/ARMS observability environment variables into embedded controller startup.
- Aligns CoPaw manager OpenTelemetry GenAI semantic convention settings.

## Why

In embedded mode the controller creates Manager and Worker containers, so bootstrap-only environment values must be available to the controller process. Without this, observability and long-stream timeout settings do not propagate to managed components.

## Compatibility

All new installer environment values are optional. Existing local installs keep the same behavior when these variables are unset.

## Validation


johnlanni:
I found a couple of compatibility issues worth fixing before merge.

1. Embedded Windows/PowerShell installer still does not propagate the CMS env vars.

The shell installer now passes the CMS/ARMS settings into `hiclaw-controller`, which is necessary in embedded mode because the controller spawns the manager/workers:

```sh
_ctrl_env_args+=(-e "HICLAW_CMS_TRACES_ENABLED=${HICLAW_CMS_TRACES_ENABLED:-false}")
_ctrl_env_args+=(-e "HICLAW_CMS_SERVICE_NAME=${HICLAW_CMS_SERVICE_NAME:-hiclaw-manager}")
_ctrl_env_args+=(-e "HICLAW_CMS_METRICS_ENABLED=${HICLAW_CMS_METRICS_ENABLED:-false}")
```

But the PowerShell embedded installer's `docker run hiclaw-controller` args still only pass the existing base envs, e.g.:

```powershell
"-e", "HICLAW_CONTROLLER_URL=http://hiclaw-controller:8090",
"-e", "HICLAW_DOCKER_NETWORK=hiclaw-net",
"-e", "HICLAW_WORKSPACE_DIR=$($config.WORKSPACE_DIR)",
"-e", "HICLAW_HOST_SHARE_DIR=$($config.HOST_SHARE_DIR)",
"-e", "HICLAW_MANAGER_ENABLED=true",
```

This means embedded installs launched via `install/hiclaw-install.ps1` will still lose CMS trace/metric settings while the `.sh` path works. Please mirror the new `HICLAW_CMS_*` propagation into the PowerShell installer as well.

2. The OTEL semconv opt-in change may drop existing HTTP semantics.

This PR changes both manager and CoPaw worker from:

```sh
"OTEL_SEMCONV_STABILITY_OPT_IN": "http",
```

to:

```sh
"OTEL_SEMCONV_STABILITY_OPT_IN": "gen_ai_latest_experimental",
```

If the intent is to add GenAI semantic conventions, this should probably be additive instead of replacing the previous HTTP opt-in, for example `http,gen_ai_latest_experimental`. Otherwise existing CMS/ARMS HTTP span semantics may change unexpectedly for users who already enabled tracing.

shiyiyue1102:
@johnlanni Thanks, fixed in 1c7822d.

Addressed both compatibility points:

1. Mirrored the embedded CMS/ARMS env propagation into `install/hiclaw-install.ps1`, so the PowerShell installer now passes `HICLAW_CMS_TRACES_ENABLED`, `HICLAW_CMS_SERVICE_NAME`, `HICLAW_CMS_METRICS_ENABLED`, and optional CMS endpoint/license/project/workspace settings into `hiclaw-controller` just like the shell installer.
2. Kept the existing HTTP semconv opt-in and added GenAI additively by changing `OTEL_SEMCONV_STABILITY_OPT_IN` to `http,gen_ai_latest_experimental` for both the CoPaw manager and CoPaw worker paths.

Verified locally with:
- `bash -n install/hiclaw-install.sh copaw/scripts/copaw-worker-entrypoint.sh manager/scripts/init/start-copaw-manager.sh`
- PowerShell parser check for `install/hiclaw-install.ps1`
- `git diff --check`

johnlanni (approved):
LGTM

PR #823, opened 2026-05-19
feat(controller): support Nacos remote skills with STS auth

shiyiyue1102:
## What changed

- Adds `remoteSkills` to Worker and Team member CRDs for Nacos-backed skill sources.
- Adds Nacos client auth modes: `nacos`, `sts-hiclaw`, and `none`.
- Supports `nacos://...?authType=...` package URIs without a global `HICLAW_NACOS_AUTH_TYPE` switch.
- Adds `ai-registry` STS access scope and UUID STS session names.
- Pins the bundled Nacos CLI in the CoPaw worker image.

## Why

Remote skill/package access needs to be declared per source and per package URI. This keeps Nacos authentication explicit, supports credential-provider based STS access, and avoids leaking global auth behavior across unrelated package sources.

## Compatibility

- Existing built-in `skills` continue to work unchanged.
- Existing `package` URIs continue to work; Nacos URIs may opt into `authType` explicitly.
- The default MCP schema from open-source main is preserved.

## Validation

- `go test ./internal/accessresolver ./internal/executor ./internal/credentials ./internal/service ./internal/auth ./cmd/hiclaw`


johnlanni:
I found a few issues around generality, compatibility, and security boundaries.

1. Remote Nacos skill ZIP extraction should validate archive entries.

`GetSkill` downloads a ZIP from a remote registry and extracts it directly into `outputDir` via `unzip -o`:

```go
// Extract the ZIP directly into outputDir/. The ZIP is expected to contain
// a top-level directory named after the skill (e.g. baidu-search/SKILL.md),
// so the result will be outputDir/{name}/SKILL.md — matching what the caller
// reads back via filepath.Join(outputDir, name).
if err := os.MkdirAll(outputDir, 0o755); err != nil {
    return fmt.Errorf("failed to create skill directory: %w", err)
}

cmd := exec.CommandContext(ctx, "unzip", "-q", "-o", tmpPath, "-d", outputDir)
if out, err := cmd.CombinedOutput(); err != nil {
    return fmt.Errorf("failed to extract skill ZIP for %s: %s: %w", name, string(out), err)
}
```

Since this ZIP comes from a remote registry, entries with absolute paths, `../`, or unexpected symlinks can write outside the temp dir in the controller container depending on unzip behavior. Please avoid shelling out to `unzip` here or validate every archive entry before extraction (reject absolute paths, path traversal, and symlinks unless explicitly intended). That would also make this less dependent on a specific ZIP layout/tool behavior.

2. `sts-hiclaw` remote skill credentials are hard-coded to namespace-wide read/write.

For `remoteSkills.authType=sts-hiclaw`, the STS request is built with hard-coded permissions and resources:

```go
tm := credprovider.NewTokenManager(client, credprovider.IssueRequest{
    SessionName: sessionName,
    Entries: []credprovider.AccessEntry{
        {
            Service:     credprovider.ServiceAIRegistry,
            Permissions: []string{"read", "write"},
            Scope: credprovider.AccessScope{
                NamespaceID: namespace,
                Resources:   []string{"agentSpec/*", "skill/*"},
            },
        },
    },
})
```

This makes the newly added `ai-registry` access-entry/resource model much less useful for remote skills: users cannot narrow the scope from the CR, and fetching skills gets write permission even though it should only need read/list. This is also less compatible with deployments that enforce least-privilege credential-provider policies.

Could this path reuse resolved `ai-registry` access entries, or at least request only the minimal read/list scope for the requested source/resources instead of hard-coding namespace-wide read/write?

3. `remoteSkills.authType` default is inconsistent with Nacos package URI behavior.

For remote skills, empty authType is mapped to `nacos`:

```go
func mapRemoteSkillAuthType(raw string) (string, error) {
    authType := strings.TrimSpace(raw)
    if authType == "" {
        authType = "nacos"
    }
```

But Nacos package URIs auto-detect auth (`userinfo/env credentials => nacos`, otherwise `none`):

```go
if authType == "" {
    if username != "" && password != "" {
        authType = nacosAuthTypeNacos
    } else if username != "" || password != "" {
        return nil, fmt.Errorf("both username and password are required in nacos URL or env ...")
    } else {
        authType = nacosAuthTypeNone
    }
}
```

This inconsistency can surprise users and makes unauthenticated/public Nacos sources fail unless they explicitly set `authType: none`. For compatibility with the existing package resolver behavior, please either use the same auto-detection default for `remoteSkills`, or document clearly why remote skills intentionally default to username/password auth.

shiyiyue1102:
Thanks, fixed in d5e0e9c. Summary: (1) skill ZIP extraction now uses Go archive/zip with entry validation and rejects path traversal, absolute paths, symlinks, and special files; (2) sts-hiclaw remote skill credentials now request read/list only and scope STS resources to the requested skills; (3) remoteSkills.authType now uses the same empty-value auto-detection as nacos package URIs, and the CRD default was removed. Verified with go test ./... and make check-crd-sync.

johnlanni (approved):
LGTM

PR #826, opened 2026-05-20
fix(controller): preserve runtime package files on reconcile

shiyiyue1102:
## Summary
- make package/base files seed-only during controller reconcile so runtime-mutated storage files are not overwritten
- keep controller-owned outputs (`openclaw.json`, credentials, mcporter config, builtins) explicitly overwritten as before
- preserve existing Nacos remote skill/package behavior from current `main`

## Validation
- `go test ./internal/executor ./internal/service` from `hiclaw-controller/`
- `git diff --check`


shiyiyue1102:
@johnlanni CI is green now. Could you help review this when you have time?

shiyiyue1102:
@johnlanni The latest checks have all passed after the rebase. Could you take a look when convenient?

johnlanni:
I reviewed this with a focus on generality, compatibility, and embedded-mode behavior.

The direction makes sense: package/base files should not blindly overwrite runtime-mutated storage state during reconcile. This is especially important in embedded mode, where local agent files and object storage are synchronized and can otherwise fight each other.

However, I think the current implementation is still too broad and may introduce compatibility problems.

1. Package updates can no longer update existing package config files

`DeployToMinIO` now treats package config files as seed-only whenever the object already exists. That protects runtime-mutated files, but it also means an explicit package upgrade or package URI change will not apply updated `config/*` defaults/fixes if the object already exists in storage.

So the behavior changes from “package deploy updates package files” to “package deploy only creates missing files”. That seems risky for package bug fixes, security fixes, or runtime package version upgrades. For example, if a worker switches from package A to package B, newly introduced files from B will be seeded, but same-path config files from B will be skipped, and files that existed only in A will remain. That can leave the worker in a partial mixed-package state.

I think we need a way to distinguish ordinary reconcile from an intentional package update, or track package revision/hash and only preserve runtime-mutated files when the package itself did not change.

2. Embedded local files are still overwritten unconditionally

Even though storage writes are now seed-only, Phase 2 still writes the package files into `/root/hiclaw-fs/agents/<worker>` unconditionally:

- top-level config files are written with `os.WriteFile(...)`
- config subdirectories are copied with `cp -r ...`

In embedded mode, that local agent directory is part of the live runtime/sync path. So a reconcile can still overwrite runtime-mutated local files with package templates, while storage preserves the runtime-mutated version. That can create local-vs-storage divergence, and depending on sync timing, may still expose the overwrite behavior this PR is trying to avoid.

I think the seed-only policy should be applied consistently to the local embedded path as well, or the code should avoid touching local package/base files during reconcile unless this is a first seed or an explicit package update.

A smaller concern: the current local copy path still shells out to `cp -r` and ignores `CombinedOutput()` errors. Since this PR is about preserving compatibility-sensitive file state, it would be better to use Go file-copy logic with explicit error handling rather than relying on best-effort shell copies.

I would not approve this yet until the package-update semantics and embedded local-file behavior are clarified/fixed.

shiyiyue1102:
@johnlanni Thanks for the detailed review. I fixed the local embedded-file path so package files are also seed-only locally, and replaced the best-effort `cp -r` path with Go file copy plus explicit error handling. The intended scope for this PR is that a package is an initialization layer for the same Worker/Manager; runtime-mutated files are preserved on reconcile, while controller-owned generated files are still overwritten by the existing DeployWorkerConfig writers. For a different package, we currently expect a different worker rather than treating package URI changes as an in-place package upgrade.

shiyiyue1102:
I rechecked the package-layer semantics and added one follow-up fix: package `skills/` and package cron jobs are now seed-only in storage as well, so the package path is consistently an initialization layer on both local and storage sides. Controller-owned outputs such as `openclaw.json`, credentials, mcporter config, builtin skills, and CR-declared skills still use their existing dedicated overwrite paths.

shiyiyue1102:
@Jing-ze The latest checks are passing now. Could you help review this when you have time?

maplefeng-a (approved):
approve

PR #828, opened 2026-05-20
feat(harness): add harness worker runtime for external CLI agents

buiminhhuy:
## Summary

- Add `harness-remote` CLI for running HiClaw worker locally (outside cluster)
- Add `--use-subscription` flag: use Claude.ai OAuth instead of AI gateway (reads `~/.claude/.credentials.json`)
- Add `--model` flag to override model in subscription mode (needed when `openclaw.json` has gateway-only model)
- Fix stale gateway env keys (`ANTHROPIC_BASE_URL`, `ANTHROPIC_*_MODEL`) not cleaned from `settings.json` when switching to subscription mode
- Add Matrix media download: `m.image`/`m.file`/`m.video`/`m.audio` attachments are saved to workspace before passing to Claude
- Fix asyncio stream buffer limit (64KB → 32MB) to handle large image base64 tool results
- Package `hiclaw-common` and `hiclaw-harness-worker` as pip-installable PyPI libraries
- Add `Makefile` targets: `install-remote`, `install-remote-dev`, `publish-common`, `publish-harness`
- Add `HarnessType` field to `WorkerSpec` CRD: `claude | gemini | opencode | codex`
- Add `ContainerManaged` field to `WorkerSpec`: skip pod reconciliation for remote/pip workers
- Update quickstart doc with subscription mode example, `--model` flag, and external domain guide

## Test plan

- [ ] `pip install hiclaw-harness-worker` then `harness-remote --help`
- [ ] Run with `--use-subscription --model claude-sonnet-4-6` and verify no `ANTHROPIC_BASE_URL` in `settings.json`
- [ ] Send Matrix image message and verify file saved to workspace
- [ ] Worker CR with `containerManaged: false` skips pod creation


maplefeng-a (commented):
Thanks for this contribution! The idea of delegating agent execution to external CLI tools (Claude Code, Gemini CLI, etc.) is a great direction — it lets HiClaw leverage mature toolchains without reinventing them.

A few observations and suggestions:

**Bugs / Issues:**
1. **Session resume is broken**: `matrix_relay.py:123` always passes `None` as `session_id` to `on_invoke`, so the resume logic in harness adapters (e.g., `claude.py`'s `--resume` flag) is never actually triggered.
2. **`_load_session()` is defined but never called** in `worker.py`, which means session state is lost on restart.

**Architecture concern:**
The current approach introduces a 4th `RuntimeType` with its own FileSync, MatrixRelay, and policies — most of which are duplicated from the existing hermes runtime (the code comments note "copied verbatim"). This creates a maintenance burden: every cross-cutting capability (anti-loop, team coordination, session management) now needs to be kept in sync across 4 separate codepaths.

We're actually exploring internally a **plugin-based approach** where "team capabilities" (file sync, messaging relay, tool access) are decoupled from runtime types and injected as a shared plugin layer. Under that model, harness/CLI delegation would be an **execution adapter** within existing runtimes rather than a standalone new runtime type. This would avoid the duplication issue and make it easier to add new CLI backends in the future.

The design is still being finalized, but I'd suggest holding off on merging this as-is. Happy to sync up once we have more details to share — the harness adapter work you've done here would fit well as a component in that architecture.

What do you think?

buiminhhuy:
Thanks @maplefeng-a 
I’m still running experiments for now to avoid blocking ongoing work, and waiting for your finalized architecture — hopefully it comes soon. If not, once the experiments are done, I’ll align everything back to the correct design based on the plugin-based architecture.

buiminhhuy:
Thanks @maplefeng-a, for the feedback — both bugs and the duplication concern are addressed. A few notes on the architectural direction we landed on:

Fixed session persistence and resume issues

Fixed a bug where matrix_relay.py always passed None as session_id to on_invoke, preventing harness adapters (e.g., Claude CLI) from triggering their `--resume` flow.
Fixed worker session recovery by properly loading persisted session state on startup. Previously, `_load_session()` was never invoked in `worker.py`, causing session information to be lost after worker restarts.

On duplication: Rather than keeping per-runtime copies, shared capabilities (`policies`, `sync`, Matrix transport) are extracted into a common leaf package under `shared/python/hiclaw_common`. The key architectural point: harness converges onto mautrix (matching the direction hermes already took), which is what makes a shared Matrix client viable across both runtimes without subclassing.

On `RuntimeHarness` as a separate type: The subprocess/stream-JSON/`--resume` execution model is fundamentally different from hermes's long-lived gateway loop — they can't share a runtime without significant abstraction cost. Keeping it separate and thin is the right call until the plugin layer design is settled. `hiclaw_common` is intentionally a leaf utility with no contracts, so plugging it into whatever interface upstream lands on should be straightforward.

shiyiyue1102:

---

shiyiyue1102:
Thank you for the extensive harness implementation. The adapter idea remains useful, but the current architecture has moved away from adding another standalone runtime that duplicates Matrix, file sync, policy, and session layers. Those capabilities now live in the current runtime and plugin boundaries, and the Worker and Team APIs have also changed substantially since this 49-file branch. We are closing this PR because it is not compatible with the accepted architecture. A future CLI-agent integration should start as a design issue and reuse the current shared capability and execution-adapter interfaces.

PR #833, opened 2026-05-21
fix(copaw): require slash-prefixed control commands

shiyiyue1102:
## Summary
- require CoPaw runtime control commands to keep a leading slash instead of accepting bare words like `stop`
- normalize Element-style double-slash commands such as `//stop` to `/stop`
- keep group-room control commands behind the normal mention gate

## Verification
- `python3 -m py_compile copaw/src/matrix/channel.py copaw/tests/test_channel_mention.py`
- `uv run --project copaw pytest copaw/tests/test_channel_mention.py` could not reach tests locally because `python-olm==3.2.16` failed to build in this environment: missing `cmake`/`gmake` fallback and a libolm compile error.


shiyiyue1102:

maplefeng-a (approved):
LGTM. 安全加固逻辑正确，测试覆盖关键路径。两个小建议：1) 测试 fake 的 is_control_command 加个空字符串保护；2) if/elif + 独立 if 之间加空行或注释区分两段决策。不阻塞合并。

PR #835, opened 2026-05-21
fix(copaw): seed worker heartbeat interval

shiyiyue1102:
## Summary
- seed the CoPaw worker agent template with a 10-minute heartbeat interval
- record the image-affecting template change in the unreleased changelog

## Validation
- jq empty copaw/src/copaw_worker/templates/agent.worker.json


flystar32 (approved):
LGTM

PR #839, opened 2026-05-21
fix(copaw): skip mc alias setup in k8s mode

shiyiyue1102:
## Summary
- Skip static `mc alias set` for CoPaw workers when `HICLAW_RUNTIME=k8s`
- Keep k8s workers on wrapper-provided `MC_HOST_hiclaw` credentials, avoiding short access-key rejection for worker names like `tt`
- Add a focused regression test and changelog entry

## Validation
- python3 -m py_compile copaw/src/copaw_worker/sync.py copaw/tests/test_worker_sync.py
- git diff --check

Note: direct pytest is not available in the system Python, and uv-based pytest is blocked locally by python-olm/libolm build prerequisites (cmake/gmake).

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,6 +4,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `hermes/`, `o
 
 ---
 
+- fix(copaw): skip static mc alias setup for k8s workers that use wrapper-provided credentials.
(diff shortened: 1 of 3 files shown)


shiyiyue1102:

maplefeng-a (approved):
LGTM — minimal additive fix, good test coverage, doesn't affect existing cloud/local paths.

PR #840, opened 2026-05-21
fix(controller): preserve default object-storage access entries

shiyiyue1102:
## Summary
- preserve the default object-storage access entry when a CR configures custom non-object-storage entries
- keep ai-gateway / ai-registry entries alongside the default storage scope
- add focused accessresolver regression coverage and changelog entry

## Validation
- go test ./internal/accessresolver


shiyiyue1102:

maplefeng-a (approved):

PR #841, opened 2026-05-21
fix(copaw): suppress missing MinIO object warnings

shiyiyue1102:
## Summary
- make CoPaw FileSync._cat treat missing optional MinIO objects as debug-only misses
- keep non-missing mc cat failures as warnings
- add focused worker sync tests and changelog entry

## Validation
- git diff --cached --check
- PYTHONPATH=copaw/src python3 -c '<direct FileSync._cat missing-object/non-missing failure checks>'

Note: pytest target was attempted with uv, but local dependency setup failed while building python-olm because cmake/gmake are unavailable in this environment.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,6 +4,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `hermes/`, `o
 
 ---
 
+- fix(copaw): suppress noisy warnings when optional MinIO objects do not exist.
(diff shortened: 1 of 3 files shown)


shiyiyue1102:
@flystar32 @maplefeng-a Checks are passing now. Could you review this when you have time?

maplefeng-a (approved):

PR #842, opened 2026-05-21
feat(controller): propagate skills API defaults to workers

shiyiyue1102:
## Summary
- add controller WorkerEnv defaults for SKILLS_API_URL and NACOS_AUTH_TYPE
- propagate those defaults into worker and manager container environments
- cover the propagated env values in worker env tests and add a changelog entry

## Validation
- go test ./internal/config ./internal/service
- git diff --check

Note: the earlier CoPaw model-preflight notification candidate was not submitted because it depends on the broader CoPaw health/preflight framework that is not yet on open-source main.


shiyiyue1102:
@flystar32 @maplefeng-a This is ready for review from the CI side. Could you take a look when convenient?

maplefeng-a (approved):

PR #849, opened 2026-05-22
feat(providers): add health checker and router for multi-AI provider …

nillikechatchat:
…support

- Add ModelSlotConfig, ProviderConfig, ModelSwitchConfig models
- Implement ProviderHealthChecker with circuit breaker pattern
- Add ModelRouter for intelligent provider selection and failover
- Support session sticky, load balancing strategies (priority/round_robin/weighted)
- Enable automatic fallback chain on provider failures
- Track provider statistics (success rate, latency, consecutive failures)
- Integrate with Higress AI Gateway for reliable provider switching

This addresses issues with frequent failures when switching providers during conversations by:
1. Real-time health monitoring of all providers
2. Circuit breaker to prevent cascading failures
3. Automatic failover to healthy providers
4. Session continuity through sticky sessions
5. Configurable retry and recovery timeouts


shiyiyue1102:
Thank you for exploring provider health checks and failover. This PR adds a standalone 900-line router under the old CoPaw source layout but does not integrate it into the request path. In the current architecture, provider configuration, routing, and failover belong at the Higress AI Gateway boundary so all runtimes share the same behavior. Maintaining a second CoPaw-only router would split ownership and observability. We are closing this PR as architecturally superseded. A new proposal should target the current gateway provider APIs and include end-to-end routing tests.

PR #851, opened 2026-05-23
docs(copaw): add CredAgent config reference

shiyiyue1102:
## Summary
- add a CredAgent configuration reference for CoPaw workers
- document protected credential paths, output sanitization rules, hot reload, and prompt hardening behavior

## Tests
- git diff --cached --check

### DIFF
--- a/docs/credagent-config.md
+++ b/docs/credagent-config.md
@@ -1,0 +1,158 @@
+# CredAgent 凭证保护配置
+
+## 概述
+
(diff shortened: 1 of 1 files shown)


maplefeng-a (approved):
LGTM. Clean and comprehensive CredAgent configuration reference — covers file guard, output sanitization, prompt hardening, and hot reload clearly.

PR #852, opened 2026-05-23
fix(install): pass stream idle timeout on Windows

shiyiyue1102:
## Summary
- pass HICLAW_AI_STREAM_IDLE_TIMEOUT_SECONDS through to the controller container in the PowerShell installer
- keep Windows installer behavior aligned with the controller stream idle timeout setting

## Tests
- git diff --cached --check
- pwsh Parser::ParseFile install/hiclaw-install.ps1


maplefeng-a (approved):
LGTM. Follows the same env-var passthrough pattern as the surrounding code. CI metrics variance is expected noise — installer change doesn't affect LLM behavior.

PR #853, opened 2026-05-23
test(controller): cover team leader ready auth

shiyiyue1102:
## Summary
- add coverage for a Team Leader reporting readiness for its own worker resource

## Tests
- git diff --cached --check
- go test ./internal/auth (from hiclaw-controller/)


maplefeng-a (approved):
LGTM. Good coverage addition — Team Leader should indeed be authorized for ActionReady on its own worker resource.

PR #876, opened 2026-05-28
fix(copaw): honor runtime path env defaults

shiyiyue1102:
## Summary
- honor COPAW_INSTALL_DIR when no CoPaw install dir is passed explicitly
- honor COPAW_WORKING_DIR when no FileSync local dir is passed explicitly
- document the image-affecting CoPaw runtime path fallback in changelog/current.md

## Validation
- python3 -m py_compile copaw/src/copaw_worker/config.py copaw/src/copaw_worker/sync.py


shiyiyue1102:
@maplefeng-a @flystar32 The conflict is resolved and the latest checks are green now. Could you take a look when convenient?

maplefeng-a (changes requested):
Thanks for keeping this small. I think the FileSync fallback needs one more adjustment before merge.

`COPAW_WORKING_DIR` currently means the CoPaw runtime directory, e.g. `<install>/<worker>/.copaw`, not the outer FileSync root. The existing tool path reflects that by passing `local_dir=working_dir.parent` when it derives `working_dir` from `COPAW_WORKING_DIR`.

With this PR, constructing `FileSync(..., worker_name="alice")` without an explicit `local_dir` while `COPAW_WORKING_DIR=/tmp/hiclaw/alice/.copaw` produces `/tmp/hiclaw/alice/.copaw/alice`. That puts the sync root inside the runtime directory, so later code that expects `<local_dir>/.copaw/workspaces/default` will look in the wrong place.

Could you adjust the fallback to preserve the existing directory contract, and add a small regression test for the `COPAW_WORKING_DIR` fallback?

shiyiyue1102:
Thanks for catching this. Fixed by preserving the existing `COPAW_WORKING_DIR` contract: FileSync now falls back to the runtime directory parent as `local_dir`, and I added a regression test for the env fallback.

Validation run:
- `python3 -m compileall -q copaw/src`
- `git diff --check`
- targeted equivalent Python assertion for the `COPAW_WORKING_DIR=/.../alice/.copaw` fallback

Commit pushed: 2b46965

maplefeng-a (approved):
LGTM. Thanks for addressing the previous review feedback. I rechecked the updated path and the old blocker is covered by the latest commit.

PR #880, opened 2026-05-28
feat(install): add Podman Quadlet autostart support via systemd

NaCodermer:
- Add interactive `step_podman_autostart` with `.env` persistence.
- Automate `.kube` Quadlet generation and systemd service activation.
- Implement smart privilege mapping for Root/Rootless Quadlet paths.
- Add robust `XDG_RUNTIME_DIR` fallback for rootless socket detection.
- Enable user linger for rootless container persistence across sessions.
- Ensure safe systemd handover with graceful failure recovery.
- Add comprehensive uninstaller cleanup for Quadlet files and services.
- Update i18n dictionaries for Podman configurations.

### DIFF
--- a/install/hiclaw-install.sh
+++ b/install/hiclaw-install.sh
@@ -653,6 +653,41 @@ msg() {
         "docker_proxy.registries_prompt.en") text="Additional allowed image sources (press Enter to skip)" ;;
         "docker_proxy.registries_label.zh") text="额外放行的镜像来源" ;;
         "docker_proxy.registries_label.en") text="Additional allowed image sources" ;;
+        # --- Podman Autostart (Quadlet) ---
+        "podman.autostart.title.zh") text="--- Podman 驻留与开机自启 (Quadlet) ---" ;;
+        "podman.autostart.title.en") text="--- Podman Linger & Autostart (Quadlet) ---" ;;
+        "podman.autostart.desc.zh") text="配置为 systemd 服务，支持开机自动启动及后台驻留。" ;;
+        "podman.autostart.desc.en") text="Configure as a systemd service for automatic startup on boot." ;;
+        "podman.autostart.enable.zh") text="启用（推荐）" ;;
+        "podman.autostart.enable.en") text="Enable (recommended)" ;;
+        "podman.autostart.disable.zh") text="不启用（容器随当前会话关闭而停止）" ;;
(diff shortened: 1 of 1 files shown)


NaCodermer:
I've investigated the recent CI failure (`test-06-multi-worker`), and it turns out the issues are related to test script flakiness and metrics tooling, rather than the `podman-quadlet-autostart` feature itself.

Here is the breakdown of the failures and the fixes I will be pushing to this branch shortly to get the CI pipeline green:

**1. Race Condition in Higress Consumer Check**

* **Cause:** The test script uses a hardcoded `sleep 30` before asserting the existence of the `worker-bob` Higress consumer. In the CI environment, the asynchronous provisioning process (pulling images, starting containers, and registering routes via the controller) can occasionally exceed 30 seconds, leading to a false negative.
* **Fix:** Replace the hardcoded `sleep` with a polling loop that checks the Higress API periodically with a reasonable timeout (e.g., 120s) until the consumer is successfully registered.

**2. SQLite Missing Table Error in Metrics Collection**

* **Cause:** The metrics extraction script (`lib/agent-metrics.sh`) attempts to query the `sessions` table for the newly created `Hermes` worker (Bob). Since Bob is freshly provisioned and hasn't processed any LLM calls yet, the SQLite database hasn't initialized the `sessions` table. This causes a Python `sqlite3.OperationalError`, which outputs a stack trace to stdout and breaks the subsequent `jq` JSON parsing.
* **Fix:** Wrap the SQLite query in a `try...except sqlite3.OperationalError` block to gracefully handle uninitialized databases and return a default fallback JSON (e.g., `{"llm_calls": 0, "tokens": 0}`).

NaCodermer:
> **Closing Comment:**
> I am closing this PR in favor of a simpler, more robust approach.
> While exploring the Quadlet (`.kube`) implementation, it became clear that wrapping the containers in Pods introduces unintended network namespace isolation. This breaks the existing `hiclaw-net` bridge network and DNS resolution (e.g., Manager being unable to reach Controller).
> Furthermore, the official installation script already leverages the `--restart unless-stopped` policy, indicating that the architecture embraces an eventual consistency/crash-loop recovery model rather than strict startup ordering.
> Therefore, I have opened a new PR using the native `podman-restart.service`, which perfectly respects the existing network topology and restart policies for both rootful and rootless setups without the overhead of Quadlets.
> Superseded by: *[#891]* *[#894]*

PR #889, opened 2026-05-30
fix(controller): Team dismiss 时保护共享 Worker 的凭据不被误删 || fix(controller): Protect shared Worker credentials from being accidentally deleted when Team dismisses

Libres-coder:
问题(#887)
删除 Team CR 时，handleDelete() 无条件清理所有 member 的凭据（MinIO + 本地 .env），不检查 Worker 是否仍被其他 Team 引用。多 Team 共享 Worker 场景下，一个 Team 被删会导致共享 Worker 凭据丢失。

修复
在 handleDelete() 中，对每个 team worker 调用 ReconcileMemberDelete() 前，通过已有的 spec.workerNames field indexer 做 refcount 检查：

•有其他 Team 引用 → 跳过清理，log 提示

•refcount 查询失败 → fail-closed，不删凭据，返回 error

•无其他引用 → 正常清理

变更
•team_controller.go +57 行：新增 listOtherTeamsReferencingWorker() + handleDelete() refcount 守卫

•team_controller_test.go +273 行：6 个测试（共享保护 / 非共享清理 / 删除中排除 / fail-closed）
测试

全部 50 个测试通过，无回归。

---
Issue(#887)
When deleting a Team CR, handleDelete() unconditionally clears the credentials of all members (MinIO + local .env), without checking whether the Worker is still referenced by other Teams. In a multi-Team shared Worker scenario, deletion of a Team will cause the shared Worker credentials to be lost.

Repair
In handleDelete(), before calling ReconcileMemberDelete() on each team worker, do a refcount check through the existing spec.workerNames field indexer:

•There are other Team references → skip cleaning, log prompt

•refcount query fails → fail-closed, does not delete credentials, returns error

•No other references → normal cleanup

change
•team_controller.go +57 lines: added listOtherTeamsReferencingWorker() + handleDelete() refcount guard

• team_controller_test.go +273 lines: 6 tests (shared protection / non-shared cleanup / exclude from deletion / fail-closed)
test

All 50 tests passed with no regressions.


Libres-coder:
test-02-create-worker 失败似乎是 CI 环境的已有问题，能否帮忙查看一下 @maplefeng-a 

---
The failure of test-02-create-worker seems to be an existing problem in the CI environment. Can you help check @mapplefeng-a

shiyiyue1102:
Thank you for protecting shared Worker credentials. The current Team and Worker model has since been decoupled: Team deletion detaches membership, while standalone Worker lifecycle and credentials are owned by the Worker controller. This branch patches only the old embedded-Team deletion path and relies on the legacy workerNames reference model, so it no longer fits the primary architecture. We are closing it as superseded by the decoupled ownership model. If a credential deletion bug still reproduces on current main, please report the exact legacy or decoupled path and we will address that focused case.

PR #891, opened 2026-06-01
feat(install): add Podman autostart support via podman-restart.service

NaCodermer:
This commit introduces native OS-level autostart support for Podman users by leveraging `podman-restart.service` and user linger.

Key changes:
- Detects the active container runtime (Docker vs. Podman).
- Prompts the user to enable autostart if Podman is detected.
- Safely configures `podman-restart.service` and Systemd Drop-in for Rootful setups.
- Implements robust Rootless support using `loginctl enable-linger` and Systemd Drop-in techniques to ensure the user-level `podman-restart.service` executes correctly on boot.

This ensures containers recover on reboot while maintaining the exact network namespaces, bridges, and aliases defined by the original `podman run` commands.

### DIFF
--- a/install/hiclaw-install.sh
+++ b/install/hiclaw-install.sh
@@ -653,6 +653,51 @@ msg() {
         "docker_proxy.registries_prompt.en") text="Additional allowed image sources (press Enter to skip)" ;;
         "docker_proxy.registries_label.zh") text="额外放行的镜像来源" ;;
         "docker_proxy.registries_label.en") text="Additional allowed image sources" ;;
+        # --- Podman Autostart ---
+        "podman.autostart.title.zh") text="--- Podman 驻留与开机自启 ---" ;;
+        "podman.autostart.title.en") text="--- Podman Linger & Autostart ---" ;;
+        "podman.autostart.desc.zh") text="配置为 systemd 服务，支持开机自动启动及后台驻留。" ;;
+        "podman.autostart.desc.en") text="Configure as a systemd service for automatic startup on boot." ;;
+        "podman.autostart.enable.zh") text="启用（推荐）" ;;
+        "podman.autostart.enable.en") text="Enable (recommended)" ;;
+        "podman.autostart.disable.zh") text="不启用（容器随当前会话关闭而停止）" ;;
+        "podman.autostart.disable.en") text="Disable (containers stop when session ends)" ;;
+        "podman.autostart.choice.zh") text="请选择 [1/2]" ;;
+        "podman.autostart.choice.en") text="Enter choice [1/2]" ;;
+        "podman.autostart.selected_enabled.zh") text="Podman 开机自启: 已启用" ;;
+        "podman.autostart.selected_enabled.en") text="Podman Autostart: enabled" ;;
+        "podman.autostart.selected_disabled.zh") text="Podman 开机自启: 已禁用" ;;
+        "podman.autostart.selected_disabled.en") text="Podman Autostart: disabled" ;;
+        "podman.autostart.title_short.zh") text="Podman 开机自启" ;;
+        "podman.autostart.title_short.en") text="Podman Autostart" ;;
+        "podman.autostart.val_enabled.zh") text="已启用" ;;
+        "podman.autostart.val_enabled.en") text="enabled" ;;
+        "podman.autostart.val_disabled.zh") text="已禁用" ;;
+        "podman.autostart.val_disabled.en") text="disabled" ;;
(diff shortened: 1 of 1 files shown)


NaCodermer:
> ### Description
> 
> 
> This PR adds robust, native OS-level autostart support for Podman environments by dynamically configuring `podman-restart.service`.
> ### Why this approach?
> 
> 
> (This supersedes the previous Quadlet approach proposed in *[#880]*).
> The Quadlet (`.kube`) approach implicitly wrapped containers in Pods, which altered network namespaces and broke the internal `hiclaw-net` DNS resolution.
> Since the official HiClaw installation script already provisions containers with the `--restart unless-stopped` policy, it relies on a crash-loop/eventual consistency model for recovery. `podman-restart.service` is the standard, official mechanism to honor these flags on boot. It restores the containers exactly as they were created—preserving all network bridges, aliases, and mounts—without the overhead or isolation issues of Quadlets.
> ### Changes Included
> 
> 
> * **Runtime Detection:** Identifies if the user is running Podman.
> * **Interactive Prompt:** Asks the user if they wish to enable the autostart service during installation.
> * **Rootful Support:** Enables the system-wide `podman-restart.service`.
> * **Rootless Support (Crucial):** Automatically configures `loginctl enable-linger` for the current user and enables the user-level systemd service (`systemctl --user`), ensuring containers restart even if the user hasn't logged in after a reboot.
> 
> 
> ### How to Test
> 
> 
> 1. Run `./hiclaw-install.sh` in a Podman environment (Rootful or Rootless).
> 2. Accept the prompt to enable the Podman autostart service.
> 3. Verify the installation completes successfully.
> 4. Reboot the host machine.
> 5. Verify that the `hiclaw-manager`, `hiclaw-controller`, and any configured workers automatically start and successfully reconnect to the `hiclaw-net` network.
> 
>

instead of #880

NaCodermer:
---

### 🐛 Issue Summary: Environment Variable Disconnect

The CI test (`test-02-create-worker.sh`) is failing due to an environment variable mismatch between the CI host machine and the `hiclaw-manager` container.

1. The CI host sets `HICLAW_DEFAULT_WORKER_RUNTIME=copaw`.
2. The test prompt instructs the Manager Agent to `"use the install default (pick whatever the env says)"`.
3. However, `HICLAW_DEFAULT_WORKER_RUNTIME` is **not being passed into the Manager Agent's container**.
4. The Agent evaluates its local environment, finds the variable unset, and falls back to creating an `openclaw` worker.
5. The test script on the host assumes a `copaw` worker was created and attempts to hit CoPaw-specific HTTP probes (`/worker/livez`), resulting in `Connection refused`.

---

### 🛠️ Proposed Solutions & Comparison

#### **Option 1: Fix Environment Variable Passthrough (Recommended)**

**Approach:** Update the bootstrap/deployment logic (e.g., the Controller or `docker-compose`) to ensure `HICLAW_DEFAULT_WORKER_RUNTIME` is explicitly passed down into the Manager Agent container.

* **Pros:** * **Solves the root cause:** Addresses an actual deployment/product gap.
* **Preserves test integrity:** Keeps the original intent of the test (verifying that the Agent can autonomously read system defaults).
* **Validates CI Matrix:** Ensures that matrix testing across different runtimes actually works as intended.


* **Cons:** Requires digging into the core application/deployment code rather than just tweaking a test script.

#### **Option 2: Explicitly Inject Variable via Prompt (Quick Fix)**

**Approach:** Modify the test script to inject the host variable directly into the LLM prompt. Change `- runtime: use the install default...` to `- runtime: ${HICLAW_DEFAULT_WORKER_RUNTIME:-openclaw}`.

* **Pros:** * Fastest to implement (one-line change).
* Guarantees the Agent creates the exact runtime the host expects.


* **Cons:** * **Degrades test value:** It spoon-feeds the answer to the LLM, bypassing the test's intent to verify the Agent's ability to read environmental configurations.

#### **Option 3: Dynamic Test Assertion (Defensive Scripting)**

**Approach:** Modify the test script to fetch the actual created runtime (`ACTUAL_RUNTIME=$(hiclaw get worker alice | jq -r '.runtime')`) and only execute the CoPaw probes if it matches.

* **Pros:** * Makes the test highly robust to environmental inconsistencies.
* **Cons:** * **High risk of False Positives:** If the variable passthrough is broken (as it is now), the script will silently skip the CoPaw probes and report a "Pass". This would leave the CoPaw worker creation logic completely untested in CI without anyone noticing.

---

**Recommendation:** Proceed with **Option 1**. Since the test explicitly exists to verify the Agent's context-awareness of its environment, the environment must be correctly provided to the container.

NaCodermer:
move to #894

PR #894, opened 2026-06-01
feat(install): add Podman autostart support via systemd

### LINKED ISSUE #900: Feature(install): Support Podman container auto-restart across host reboots (Host-safe approach)
### Description
Currently, containers deployed via HiClaw on Podman do not automatically restart after a host system reboot. To provide a seamless experience similar to Docker's default behavior, we need an automated way to recover HiClaw containers on system boot.

### Proposed Solution
Instead of modifying the global, user-managed `podman-restart.service` (which poses risks to other non-HiClaw workloads on the same host), we will introduce a **dedicated, HiClaw-owned Systemd service** (`hiclaw-podman-restart.service`).

### Implementation Details
1. **Isolated Ownership:** Create a dedicated `hiclaw-podman-restart.service` unit file.
2. **Strict Scoping:** The service will use Podman filters (e.g., `--filter name=hiclaw`) to ensure it *only* manages HiClaw containers, avoiding interference with other host processes.
3. **Environment Awareness:** Support both Rootful (`/etc/systemd/system/`) and Rootless (`~/.config/systemd/user/`) paths seamlessly.
4. **Clean Uninstall:** The uninstall script will safely disable and remove this dedicated unit without touching the native Podman configurations.
5. **Separation of Concerns:** Separate the `detect_socket` (read-only) logic from `ensure_podman_socket` (state mutation) to prevent unexpected host-side mutations during simple checks.

### Tracking
This issue tracks the implementation proposed in PR #894.

NaCodermer:
Continues the work from #891

This PR introduces native OS-level autostart support for Podman environments,
ensuring containers recover automatically after host reboots.

Implementation

A dedicated `hiclaw-podman-restart.service` is deployed and owned entirely
by HiClaw, maintaining a safe host-side boundary per the direction of #900.

Key changes

Podman autostart (step + setup)
Adds `step_podman_autostart` to the interactive install wizard and
`setup_podman_autostart` for the actual systemd configuration.

The service is strictly scoped via `--filter name=hiclaw --filter
restart-policy=always --filter restart-policy=unless-stopped`, ensuring
only HiClaw containers with an explicit restart policy are managed.
No other containers on the host are affected.

Rootful: system unit written to /etc/systemd/system/.
Rootless: user unit written to ~/.config/systemd/user/ with graceful
loginctl enable-linger handling (native attempt first, sudo fallback).
The final success message is gated — emitted only when both the service
and user linger are confirmed active.

Upgrade flow preserves the existing HICLAW_PODMAN_AUTOSTART value and
allows the user to change it. NON_INTERACTIVE and QUICKSTART modes skip
the prompt; autostart is controlled entirely by the environment variable.

Socket mutation decoupled from detect_socket
`detect_socket()` is now a strictly read-only probe. All host-side socket
activation is delegated to a new `ensure_podman_socket()` function, which
is called lazily — only when HICLAW_MOUNT_SOCKET=1 and the runtime is Podman.

`ensure_podman_socket` activates the systemd socket unit (rootful or user-level)
and polls internally for socket file readiness (max 3s, early exit). Callers
proceed directly to detect_socket with no additional coordination needed.

This resolves stale socket issues caused by systemd idle timeouts without
contaminating the read-only discovery function.

XDG_RUNTIME_DIR baseline exported at entry
`check_container_runtime` now exports XDG_RUNTIME_DIR once, at the earliest
point, for non-root users. All downstream functions inherit the correct
rootless environment baseline without each needing their own guard.

podman-docker masquerade detection
`check_container_runtime` now inspects `docker --version` output for the
string "podman" to correctly set DOCKER_CMD=podman when the podman-docker
compatibility package is installed.

Docker Desktop socket paths added to detect_socket
Adds $XDG_RUNTIME_DIR/docker.sock (rootless Docker on Linux) and
~/.docker/run/docker.sock (Docker Desktop on macOS) to the Docker branch
fallback chain.

Clean uninstall boundary
Uninstall safely disables and removes `hiclaw-podman-restart.service`
without activating any sockets or modifying host-level state.
The native podman-restart.service is strictly untouched.

Closes #900


NaCodermer:
> ### Why this approach?
> 
> 
> (This supersedes the previous Quadlet approach proposed in *[#880]*).
> The Quadlet (`.kube`) approach implicitly wrapped containers in Pods, which altered network namespaces and broke the internal `hiclaw-net` DNS resolution.
> Since the official HiClaw installation script already provisions containers with the `--restart unless-stopped` policy, it relies on a crash-loop/eventual consistency model for recovery. `podman-restart.service` is the standard, official mechanism to honor these flags on boot. It restores the containers exactly as they were created—preserving all network bridges, aliases, and mounts—without the overhead or isolation issues of Quadlets.
> 
> 
> ### How to Test
> 
> 
> 1. Run `./hiclaw-install.sh` in a Podman environment (Rootful or Rootless).
> 2. Accept the prompt to enable the Podman autostart service.
> 3. Verify the installation completes successfully.
> 4. Reboot the host machine.
> 5. Verify that the `hiclaw-manager`, `hiclaw-controller`, and any configured workers automatically start and successfully reconnect to the `hiclaw-net` network.
> 
>

instead of #880 #891

Finally, use a HiClaw-owned unit hiclaw-podman-restart.service instead of overriding the global podman-restart.service.

NaCodermer:
我在 kubuntu26.04 上测试 podman rootful 和 rootless 安装成功， 且在主机重启后能正常拉起容器
自动启动默认开启，可配置为不开启，且支持 HICLAW_PODMAN_AUTOSTART 变量持久化
兼容升级和卸载

---
I tested podman rootful and rootless on kubuntu26.04. The installation was successful, and the container could be pulled up normally after the host was restarted.
Automatic startup is enabled by default, can be configured not to enable, and supports HICLAW_PODMAN_AUTOSTART variable persistence
Compatible with upgrade and uninstall

shiyiyue1102:
Thanks for the update. The Podman autostart direction makes sense, but before merging we should narrow the host-side behavior because this installer can modify system-level Podman/systemd configuration.

A few concerns from the current diff:

1. `install/hiclaw-install.sh` is changed from executable to non-executable (`100755 -> 100644`). The script and install docs still support direct `./hiclaw-install.sh` usage, so please keep the executable bit.
2. The PR writes a drop-in override for `podman-restart.service` and removes/disables that service during HiClaw uninstall. `podman-restart.service` is Podman-wide, not HiClaw-owned, so this can overwrite user-managed Podman systemd configuration or affect other containers with restart policies on the same host.
3. `detect_socket()` now starts/enables `podman.socket` as a fallback. Socket detection should ideally avoid host-side mutations, or the behavior should be explicitly scoped and confirmed by the install step.
4. This is a non-trivial install/lifecycle behavior change but does not link an issue or RFC. Please link the tracking issue or open one so we can agree on the expected host-side ownership and cleanup behavior.

My preferred shape is either:

- only enable the official `podman-restart.service` / user service when requested, without writing an override and without disabling it on HiClaw uninstall; or
- create a HiClaw-owned unit such as `hiclaw-podman-restart.service` that only manages HiClaw containers and can be safely removed by the installer.

Could you adjust the implementation around the service ownership / cleanup boundary, or share another safer approach for Podman reboot recovery?

NaCodermer:
@shiyiyue1102 

Thanks for the detailed review! I totally agree that modifying the global podman-restart.service is dangerous for host stability.
I will adopt your second suggestion: creating a dedicated hiclaw-podman-restart.service that only tracks HiClaw's own containers. I will also revert the file permission change and open an Issue to track this feature.
I'll update the PR shortly!

NaCodermer:
@shiyiyue1102 

https://github.com/agentscope-ai/HiClaw/commit/837b66e071870ecc7a3b86c79f603084244b1dfb

feat(install): add Podman autostart support via systemd

This commit introduces native systemd autostart support for Podman
environments, addressing the lack of automatic container recovery
after host reboots when using Podman as the container runtime.

Key changes and improvements:
1. Dedicated Systemd Service (Host-side Safety):
   - Created a HiClaw-owned unit `hiclaw-podman-restart.service` instead
     of overriding the global `podman-restart.service`.
   - Strictly scoped via `--filter name=hiclaw` to ensure it only manages
     HiClaw-related containers, leaving user-managed containers untouched.
   - Safely removes only this dedicated service during uninstall.

2. Rootless Support & Graceful Escalation:
   - Supports both rootful (system-wide) and rootless (user-level) deployments.
   - For rootless, dynamically attempts `loginctl enable-linger` natively
     first via Polkit. If privileges are required, it gracefully falls back
     to prompting for `sudo`.

3. Socket Mutation Decoupling (Stale Socket Fix):
   - Removed host-side mutation logic from `detect_socket()`, making it a
     purely read-only environment probe.
   - Introduced `ensure_podman_socket()` to explicitly handle socket
     activation.
   - Lazy Initialization: Socket activation is now strictly scoped to
     `install_manager` and triggered only when `HICLAW_MOUNT_SOCKET=1`.
   - Fixed a Stale Socket bug: Actively enables the podman socket before
     probing to prevent the socket from being deactivated by systemd's idle
     timeout.

4. Clean Uninstall Boundary:
   - Running `./hiclaw-install.sh uninstall` no longer triggers any socket
     activation or host-side state mutations.

Addresses host-side ownership concerns raised in previous reviews.

Ref: #900

NaCodermer:
Hi @shiyiyue1102 , just a quick heads-up: the CI is fully green now! ✅ All your previous concerns regarding host-side boundary, rootless fallback, and the lazy socket initialization have been addressed in the latest commits. Ready for your review whenever you have time. Thanks!

shiyiyue1102 (changes requested):
Thanks for the update. The host-side ownership boundary is much safer now: the PR creates a HiClaw-owned `hiclaw-podman-restart.service`, keeps `detect_socket()` read-only, and only removes the HiClaw-owned unit during uninstall. CI is green.

One remaining issue before approval: `setup_podman_autostart()` still logs `install.podman.success` unconditionally after the rootful/rootless `systemctl enable --now hiclaw-podman-restart.service` branch. If `systemctl` fails, or rootless linger cannot be enabled, the installer can still end with "Podman autostart successfully configured", even though reboot recovery is not actually guaranteed. The user-level failure hint also still points to `podman-restart.service` instead of the dedicated `hiclaw-podman-restart.service`.

Could you make the final status reflect the actual result, and update the manual recovery hint to the dedicated service name?

---

这版的 host-side 边界已经安全很多：现在是 HiClaw 自己的 `hiclaw-podman-restart.service`，`detect_socket()` 保持只读，卸载也只删除 HiClaw 自己的 unit。CI 也已经是绿色。

合并前还剩一个问题：`setup_podman_autostart()` 在 rootful/rootless 的 `systemctl enable --now hiclaw-podman-restart.service` 失败后，最后仍会无条件打印 `install.podman.success`。如果 `systemctl` 失败，或者 rootless linger 没启用成功，安装器仍可能显示“Podman autostart successfully configured”，但实际重启恢复并不能保证。另外 user-level 失败提示里仍然指向 `podman-restart.service`，也需要改成专用的 `hiclaw-podman-restart.service`。

请把最终状态改成反映真实执行结果，并同步更新手动恢复提示里的服务名。

NaCodermer:
Thanks for catching those cases! 

I've pushed an update addressing your feedback with the following fixes:

* **Conditional Success Logging:** I introduced an `_autostart_ok` flag to track the actual state of the service. The final success message is now strictly conditional on this flag resolving to `1`.

* **Strict Rootless Validation:** In rootless mode, the script now tracks `_linger_ok` alongside the service enablement. It only considers the autostart setup successful if *both* the user-level service and systemd linger are active. If the service enables but linger fails, it explicitly warns the user that containers won't auto-recover on reboot rather than logging a false success.

* **Accurate Service Names:** I updated all the UI status messages and manual fallback hints (in both English and Chinese) to correctly reference `hiclaw-podman-restart.service` instead of `podman-restart.service`.

The last commit is updated and ready for your review. Let me know if anything else stands out!

NaCodermer:
https://github.com/agentscope-ai/HiClaw/pull/894#issuecomment-4623410839

```text
=== Finished: 21-team-project-dag ===

========================================
  Test Summary
========================================
  Total:  27
  Passed: 26
  Failed: 1
========================================

Failures:
  - Leader did not respond within timeout


@dag-team-21249-dev: ⚠️ Model service check failed: model chat completion preflight failed: timed out
Provider: hiclaw-gateway, Model: qwen3.6-plus
```

Root Cause: The integration test test-21-team-project-dag failed due to a model service timeout. The Leader agent did not respond within the expected timeout period because the model service (qwen3.6-plus via hiclaw-gateway) failed its preflight check with a connection timeout.

NaCodermer:
The same set of code, CI passed before #909 was merged. The current error https://github.com/agentscope-ai/HiClaw/actions/runs/26991105018/job/79652326213 was not a problem during the last CI (see https://github.com/agentscope-ai/HiClaw/actions/runs/26959094841/job/79546569986). The infrastructure might be unstable, so I will restart the PR to trigger CI again.

NaCodermer:
Hi @shiyiyue1102 , just a quick heads-up: the CI is fully green now! ✅ 
Ready for your review whenever you have time. Thanks!

shiyiyue1102:
Thanks for the update. The implementation side now looks aligned with the safer host-side boundary: the PR uses the dedicated `hiclaw-podman-restart.service`, keeps the native `podman-restart.service` untouched, gates the final success message on the actual service/linger result, and the latest CI run is green.

The remaining item before merge is the PR description. It still says the implementation leverages `podman-restart.service` and configures Systemd drop-ins, while the current code and #900 use the HiClaw-owned `hiclaw-podman-restart.service` approach. Could you update the PR body to match the latest implementation? After that, this should be ready to merge.

---

感谢更新。实现侧现在已经符合更安全的 host-side 边界：当前 PR 使用专用的 `hiclaw-podman-restart.service`，不再触碰原生 `podman-restart.service`，最终成功日志也已经按 service/linger 的真实结果判断，最新一轮 CI 也是绿色。

合并前剩下的问题是 PR 描述。当前 body 仍写着通过 `podman-restart.service` 和 Systemd drop-in 实现，但最新代码和 #900 已经改为 HiClaw 自己的 `hiclaw-podman-restart.service` 方案。请把 PR body 更新成和最新实现一致；修完这个后，这个 PR 就可以合并。

NaCodermer:
@shiyiyue1102 
Thanks for the review! I have updated the PR description to accurately reflect the hiclaw-podman-restart.service implementation, the gated success logic, and other details from the latest commits. It should be good to merge now!

NaCodermer:
Tested podman rootful(also sudo -E) and rootless on Kubuntu 26.04. Everything works as expected.

All changes are complete and this is ready for final review.

shiyiyue1102 (approved):
LGTM.

The previous blocking points have been addressed: the implementation now uses the HiClaw-owned `hiclaw-podman-restart.service`, leaves the native Podman service untouched, keeps socket detection read-only, scopes socket activation to the explicit Podman socket path, gates the final autostart success message on the actual service/linger result, and the PR description now matches the current implementation. Latest CI is green.

PR #901, opened 2026-06-02
fix(copaw): handle static MinIO in k8s mode

buiminhhuy:
## Root Cause

Issue in sync.py:294-301 when running in Kubernetes mode with a static MinIO configuration:

_ensure_alias() skips mc alias set, assuming it will be handled by mc-wrapper.
mc-wrapper.sh only configures MC_HOST_hiclaw when an STS/bearer token is available (Aliyun mode).
This cluster uses an internal MinIO instance with static credentials, so the bearer token is empty and ensure_mc_credentials() becomes a no-op.
As a result, mc has no knowledge of the hiclaw alias and interprets hiclaw/hiclaw-storage/... as a local filesystem path, causing the below issue.

```
subprocess.CalledProcessError: Command '['/usr/local/bin/mc', 'mirror', 'hiclaw/hiclaw-storage/agents/hbleader/', '/root/hiclaw-fs/agents/hbleader/', '--overwrite', '--exclude', 'credentials/**']' returned non-zero exit status 1.
Failed to mirror from MinIO: Command '['/usr/local/bin/mc', 'mirror', 
'hiclaw/hiclaw-storage/agents/hbleader/', '/root/hiclaw-fs/agents/hbleader/', 
'--overwrite', '--exclude', 'credentials/**']' returned non-zero exit status 1.
```

##Fix

In Kubernetes mode, check whether MC_HOST_hiclaw is configured:

If MC_HOST_hiclaw is set (cloud Kubernetes + STS), keep the existing behavior and skip alias configuration.
If MC_HOST_hiclaw is not set (Kubernetes + static MinIO credentials), fall through to the standard mc alias set flow and configure the alias using the credentials from the runtime configuration.


maplefeng-a (approved):
CI is green and the CoPaw-scoped change looks good under the agreed review policy. Approving.

shiyiyue1102 (changes requested):
Thanks for the fix. The runtime change is aligned with the reported static-MinIO failure, but one blocking test mismatch remains.

`copaw/tests/test_worker_sync.py::test_ensure_alias_skips_static_alias_in_k8s_mode` currently runs with `AGENTTEAMS_RUNTIME=k8s` and no `MC_HOST_<alias>`, then asserts that `_mc` is never called. With this PR, that exact case intentionally falls through to `mc alias set`, so the existing unit-test contract now contradicts the implementation. The current CI matrix does not appear to execute this test, which is why all reported checks remain green.

Please:
1. Update the skip test to set `MC_HOST_<alias>` and keep asserting that no static alias is configured.
2. Add a separate K8s test with `MC_HOST_<alias>` unset and assert the expected `alias set` call.
3. Update the `_ensure_alias` docstring to mention static/self-hosted K8s as part of the static-credential path.

Once both credential branches are locked down, this should be ready to merge.

---

感谢修复。运行时代码与所描述的静态 MinIO 故障方向一致，但目前还有一个阻塞性的测试不一致。

现有的 `copaw/tests/test_worker_sync.py::test_ensure_alias_skips_static_alias_in_k8s_mode` 在 `AGENTTEAMS_RUNTIME=k8s` 且没有 `MC_HOST_<alias>` 时，断言不会调用 `_mc`。而这个 PR 正是要让该场景落到 `mc alias set`，因此现有单测契约已经与实现相矛盾。当前 CI 看起来没有执行到这条单测，所以所有检查仍显示为绿色。

请补充：
1. 在 skip 测试中设置 `MC_HOST_<alias>`，并继续断言不会配置静态 alias。
2. 新增一个未设置 `MC_HOST_<alias>` 的 K8s 测试，断言会执行预期的 `alias set`。
3. 更新 `_ensure_alias` 的 docstring，明确静态/自建 K8s 也走静态凭据路径。

两个凭据分支都有测试锁定后，这个 PR 就可以继续合并。

shiyiyue1102:
Follow-up after tracing the full storage credential path: please keep using the existing `storage.provider` / `AGENTTEAMS_STORAGE_PROVIDER`; no new user-facing configuration field is needed. However, the provider value currently reaches the Controller only and is not projected into Worker containers. Please propagate it through `WorkerEnvDefaults` and `WorkerEnvBuilder`.

Expected behavior:
- `minio`: do not depend on Controller STS; configure the static alias from `AGENTTEAMS_FS_ENDPOINT`, `AGENTTEAMS_FS_ACCESS_KEY`, and `AGENTTEAMS_FS_SECRET_KEY`.
- `oss`: require Controller-issued `MC_HOST_<alias>`; if it is missing or refresh fails, fail explicitly instead of falling back to a static alias.
- Provider absent, for mixed-version compatibility: prefer an existing `MC_HOST_<alias>`; otherwise retain the static-MinIO fallback.

This keeps deployment mode (`k8s`) separate from storage credential mode and avoids masking an OSS STS failure as a local-path error.

In addition to the unit-test and docstring updates requested in the previous review, please also:
1. Update `changelog/current.md`, because this changes the built CoPaw image.
2. Link the existing issue in the PR body, preferably with `Fixes #957`.

---

进一步梳理完整的存储凭据链路后，建议继续复用现有的 `storage.provider` / `AGENTTEAMS_STORAGE_PROVIDER`，不需要新增用户侧配置字段。不过这个 provider 当前只进入了 Controller，还没有透传到 Worker 容器，请通过 `WorkerEnvDefaults` 和 `WorkerEnvBuilder` 完成透传。

期望行为：
- `minio`：不依赖 Controller STS，使用 `AGENTTEAMS_FS_ENDPOINT`、`AGENTTEAMS_FS_ACCESS_KEY` 和 `AGENTTEAMS_FS_SECRET_KEY` 配置静态 alias。
- `oss`：必须拿到 Controller 签发的 `MC_HOST_<alias>`；如果缺失或刷新失败，应明确报错，不能降级到静态 alias。
- provider 未设置时，为兼容新旧 Controller/Worker 混用：优先使用已有的 `MC_HOST_<alias>`，否则保留静态 MinIO fallback。

这样可以把部署模式（`k8s`）与存储凭据模式分开，也能避免把 OSS STS 获取失败掩盖成错误的本地路径问题。

除上一次 review 已要求的单测和 docstring 修改外，还请补充：
1. 更新 `changelog/current.md`，因为这次修改会进入 CoPaw 构建镜像。
2. 在 PR body 中关联现有 issue，建议使用 `Fixes #957`。

oss-maintainer (commented):
### Summary
Clean, minimal fix for static MinIO in k8s mode. Verified the logic against the surrounding code:

- When `MC_HOST_<alias>` is already injected by the mc-wrapper, behavior is unchanged (skip alias setup).
- When it is absent (static MinIO in k8s), execution now falls through to the existing static alias path (`_mc("alias", "set", ...)` with `self.endpoint` / static credentials), which is the correct target; `os` is already imported and the `_cloud_mode` branch is unaffected.

One tiny (non-blocking) observation: `mc_host_set` is computed twice — once for the diagnostic log at the top of `_ensure_alias` and again inside the k8s branch; the inner recomputation could reuse the outer value.

Note: I could not verify CLA status for this repository (no `license/cla` commit-status context is configured), so this is a COMMENT rather than an APPROVE per policy. The code itself looks good to me.

---

Issue #902, opened 2026-06-03
[Proposal] Decouple Team and Worker CRs: separate organizational membership from runtime definition


Sunrisea:
## Summary

The current `Team` CRD bundles two distinct concerns into a single resource:

1. **Organizational relationships** — who is in the team, who is the leader, who is admin, peer-mention policy.
2. **Per-member runtime definition** — model, identity, soul, package, mcpServers, env, accessEntries, etc. for every leader/worker.

This bundling has produced concrete pain points: triplicated API types, an over-burdened `TeamReconciler`, no independent K8s identity for team members, and tight coupling between organizational changes and runtime state.

This issue proposes splitting the two concerns:

- `Team` CR keeps only **organizational/policy fields** (`workerMembers[]`, `admin`, `humanMembers`, `accessPolicy`, `channelPolicy`).
- Every team member becomes a regular `Worker` CR, owned by `WorkerReconciler` exclusively.
- `TeamReconciler` becomes a thin coordinator that only manages Matrix Room membership, MinIO `AGENTS.md` coordination injection, and status aggregation. It never touches Pods, ServiceAccounts, configs, or routes.
- Crucially, **`Worker` CRs carry no team annotation**. Team membership is expressed only by `Team.spec.workerMembers[].name`, queried via field indexer.

We have a working PoC of this design (controller + dual-format API + automatic migrator) running in our internal integration environment. End-to-end validation passed across legacy-format, decoupled-format, and mixed scenarios with no Pod rebuilds and no Matrix Room recreation. We would like to upstream this to HiClaw and are opening this issue first to gather feedback on the architecture before sending PRs.

---

### 1.1 The "all-in-one" Team CR

```go
type TeamSpec struct {
    Description   string             `json:"description,omitempty"`
    TeamName      string             `json:"teamName,omitempty"`
    Admin         *TeamAdminSpec     `json:"admin,omitempty"`
    HumanMembers  []TeamMemberSpec   `json:"humanMembers,omitempty"`
    Leader        LeaderSpec         `json:"leader"`              // 15 runtime fields inlined
    Workers       []TeamWorkerSpec   `json:"workers,omitempty"`   // 18 runtime fields inlined per member
    PeerMentions  *bool              `json:"peerMentions,omitempty"`
    ChannelPolicy *ChannelPolicySpec `json:"channelPolicy,omitempty"`
}
```

[...]

PR #910, opened 2026-06-03
feat: decouple Team and Worker CRs (org membership vs runtime definition)

Sunrisea:
> Proposal: agentscope-ai/HiClaw#902
>
> This PR is the reference implementation discussed in #902. It is large but composed of six self-contained commits aligned to the rollout stages described there. Each commit is independently buildable and testable; reviewing commit-by-commit is recommended.

## Why

`Team` CR today bundles **organizational relationships** (membership, leader, admin, channel policy) with **per-member runtime definition** (model, identity, soul, mcp servers, accessEntries, ...). This single resource is mutated for unrelated reasons and produces the pain points enumerated in #902:

1. Three near-duplicate API types (`WorkerSpec` / `TeamWorkerSpec` / `LeaderSpec`) — every new runtime field must be added in three places.
2. `TeamReconciler` is over 1000 LOC and owns Pod/config/expose/SA — a single member's `model` change triggers a Team-level reconcile.
3. Team members have no independent K8s identity (`kubectl get worker <team-member>` fails).
4. Organizational changes interfere with runtime state (adding a member rewrites every existing member's hash).
5. Standalone Worker and team-member Worker live on two completely different controller code paths.

Full background and design rationale are in #902.

## What this PR does

- Adds new fields to `Team` CR: `spec.workerMembers[]` (refs to existing Worker CRs) and `spec.accessPolicy.defaultEntries`.
- Strips team-annotation awareness from `WorkerReconciler` — `Worker` CRs are now fully team-agnostic.
- Adds a decoupled reconcile path on `TeamReconciler`: when `spec.workerMembers` is set, it only manages Team Room membership, MinIO `AGENTS.md` coordination injection, heartbeat config, and status aggregation. It no longer creates Pods, Matrix users, ServiceAccounts, configs, or routes for members.
- Adds a controller-internal auto-migrator that drives legacy Team CRs through `WorkerCRsCreated → StatusSeeded → PodsReparented → CoordinationInjected → TeamSpecPatched → Migrated` with no Pod rebuilds and no Matrix Room recreation.
- Adds a dual-format `POST /teams`: existing legacy clients keep working; new clients can submit `{members: [...]}`. Sending both forms is a 400.
- Pod `ownerReferences` is reparented from Team CR → Worker CR during migration so `kubectl delete team` no longer cascades into Pods.

## Design highlights (from #902)

- **`Worker` CRs carry no team annotation.** Membership is expressed only via `Team.spec.workerMembers[].name` and discovered by a `spec.workerMembers[*].name` field indexer (O(1)). This keeps the Team↔Worker relationship single-sourced.
- **Coordination delivery is out-of-band.** Team membership changes do not patch Worker CRs and do not trigger `WorkerReconciler`. The agent process learns about team context via Matrix Room invites + a hot-reloadable block in `AGENTS.md` synced by `mc mirror`.
- **Permissions are merged at issuance time.** `AccessResolver` returns `Worker.spec.accessEntries ∪ Team.spec.accessPolicy.defaultEntries`. The hard-coded `DefaultEntriesForTeamMember()` is preserved as a fallback when a Team has no explicit `accessPolicy`.
- **Migration is crash-safe via "move-then-write".** The Team CR `UPDATE` is the *last* step; legacy `TeamMemberStatus` fields (`Observed`/`SpecHash`/`RoomID`/`ExposedPorts`/`RuntimeName`) are pruned by K8s on that update, after Worker CRs already hold the data. A crash before the final UPDATE is fully recoverable.

## Backwards compatibility

- Existing Team CRs without `spec.workerMembers` continue to use the legacy reconcile path.
- The legacy `POST /teams` request shape (`{leader, workers}`) keeps working unchanged.
- `LeaderSpec` and `TeamWorkerSpec` are marked `// Deprecated` but **not removed** in this PR. Removal is proposed for a future major version.
- `WorkerStatus` is unchanged. `TeamStatus.Members` is augmented; legacy fields are pruned only after the migrator's final UPDATE on each Team.

## Migration safety

- Cluster-level flag `--enable-auto-team-migration=true` (default on).
- Per-Team opt-out: annotation `hiclaw.io/auto-migrate=disabled`. Reverse: `=rollback`.
- Health pre-check before each Team enters the state machine: `phase=Active` + every member has non-empty `MatrixUserID` and `RoomID`.
- Token bucket (1 QPS, concurrency 1 by default) gates new migration starts; in-flight migrations are not rate-limited so half-migrated states cannot linger.
- Leader-elected: only the leader controller advances the state machine.
- Finalizer `hiclaw.io/migration-in-flight` blocks Team deletion mid-migration.
- Events `MigrationPhaseAdvanced` / `MigrationPhaseFailed` and metrics `hiclaw_team_migration_phase{team,phase}` / `hiclaw_team_migration_failures_total`.

## Validation

End-to-end integration tests on an internal v1.1.2-based cluster:

| Phase | Scenario | Result |
|---|---|---|
| 1 | Baseline v1.1.2 legacy `POST /teams` reaches `Active` | Pass |
| 2a | Decoupled controller: Worker CRs have no team annotations; `Team.spec.workerMembers` references them | Pass |
| 2b | Decoupled `POST /teams` (`{members: [...]}`) creates two Worker CRs; leader replies to user | Pass |
| 3 | Legacy `POST /teams` against the new controller reaches `Active`, then auto-migrator advances to `Migrated` (workerMembers populated, Worker CRs created, no Pod rebuild, no Room recreation) | Pass |
| 4 | Conflict detection: duplicate worker name → 409; mixing both formats → 400; missing `team_leader` → validation error | Pass |
| 4 | Three Teams (legacy-baseline, decoupled, legacy-migrated) coexist and operate on the same cluster | Pass |

Unit tests added in this PR (Go):

- `internal/server/create_team_decoupled_test.go` — 276 lines covering format detection, validation, conflict cases, atomic rollback.
- `internal/controller/team_controller_test.go` — +519 lines covering the decoupled reconcile path.
- `internal/migration/team_migration_test.go` — 587 lines covering the migration state machine and crash-resume.

## Risks / known limitations

- A `Worker` is allowed to belong to at most one Team (enforced by the field indexer). Multi-Team is intentionally deferred — design does not block it.
- `mc mirror` cadence is the lower bound on coordination-context propagation latency (seconds-level today). Acceptable; agents also use Matrix room-join events as a secondary signal.
- Old agent versions that do not hot-reload `AGENTS.md` will pick up changes only on restart. Documented; consider a minimum agent version note in release notes.

## Open discussion items (mirrored from #902)

1. Annotation-free Worker CRs vs a thin `hiclaw.io/team` annotation for ops convenience.
2. Default of `--enable-auto-team-migration=true` vs requiring explicit opt-in.
3. Deprecation timeline for `LeaderSpec` / `TeamWorkerSpec` — propose one minor release of soft-deprecation, removal in next major.
4. Whether the Leader vs Worker coordination block content should remain differentiated (kept differentiated in this PR) or unified.

Happy to split this PR further if commit-by-commit review is preferred over a single review thread, or to gate the new path behind a feature flag if maintainers prefer a phased landing.

Closes / addresses: #902


shiyiyue1102:

---

shiyiyue1102:
Thank you for driving the Team and Worker decoupling design and for the extensive migration work. The core direction has now landed on main through #986, including WorkerMembers, the decoupled reconcile path, independent Worker lifecycle ownership, status projection, and the agentteams.io API migration. This branch now conflicts across 38 files and represents an earlier version of the same architecture, so it should not be merged on top of the landed implementation. We are closing it as superseded. The work and discussion here were valuable input to the final design.

PR #914, opened 2026-06-05
fix(copaw): suppress warning for missing MinIO objects

shiyiyue1102:
## Summary
- Treat missing MinIO objects from `mc cat` as debug-only in CoPaw file sync.
- Keep non-missing `mc cat` failures as warnings with redacted/previewed stderr.

## Validation
- `python3 -m py_compile copaw/src/copaw_worker/sync.py`
- `python3 -m pytest copaw/tests/test_sync.py` not run: local environment has no `pytest` module.

### DIFF
--- a/copaw/src/copaw_worker/sync.py
+++ b/copaw/src/copaw_worker/sync.py
@@ -192,11 +192,11 @@ def _mc(
         exc.cmd = redacted_cmd
         log = logger.warning if warn_on_error else logger.debug
         log(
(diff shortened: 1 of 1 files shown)


shiyiyue1102:

maplefeng-a (approved):
LGTM. I rechecked the CoPaw MinIO/mc cat path: missing objects remain debug-only, non-missing failures still warn with previewed stderr, and local py_compile plus test_sync passed.

PR #915, opened 2026-06-06
docs(teamharness): add runtime config contract

shiyiyue1102:
## Summary

- add the TeamHarness member runtime config contract
- add the staged TDD plan for TeamHarness runtime integration

## Validation

- `git diff --check`

## Sync Notes

This is a docs-only slice from the internal sync baseline. It is independent from PR #914 and does not touch runtime code.

### DIFF
--- a/docs/member-runtime-config-contract.md
+++ b/docs/member-runtime-config-contract.md
@@ -1,0 +1,117 @@
+# Member Runtime Config Contract
+
+This document defines the YAML config snapshot that the HiClaw controller writes
+to object storage for a managed runtime member. Managed runtime workers and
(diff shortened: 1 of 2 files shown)


shiyiyue1102:
@flystar32 @maplefeng-a The docs check is passing now. Could you review this when convenient?

maplefeng-a (approved):
LGTM. The runtime config contract and staged TDD plan are useful as a docs-only slice. Non-blocking note for follow-up: the current CRD/runtime compatibility docs still use `copaw` as the internal value for QwenPaw, so please keep the eventual controller/runtime implementation aligned with that compatibility boundary unless a new `qwenpaw` runtime value is introduced deliberately.

PR #916, opened 2026-06-06
fix(agent): update file sharing path guidance

shiyiyue1102:
## Summary

- update CoPaw Worker file-sharing guidance to use `/root/hiclaw-fs/agents/...`
- update Team Leader file-sharing guidance to use the same current runtime path
- record the image-affecting prompt change in `changelog/current.md`

## Validation

- `git diff --check`

## Sync Notes

This is a narrow agent-facing content slice from the internal sync baseline. It is independent from PR #914 and only updates path guidance.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,6 +4,8 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 ---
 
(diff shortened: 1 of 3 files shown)


shiyiyue1102:
@flystar32 @maplefeng-a The checks are passing now; could you help review this when you have time?

maplefeng-a (approved):
LGTM. The file-sharing skill text keeps agents on shared/... / global-shared/... and updates the internal absolute path that should not be exposed. Minor non-blocking nit: the changelog wording says "use /root/hiclaw-fs/agents/...", which can be read differently from the skill text's "do not expose absolute paths" guidance.

PR #917, opened 2026-06-08
fix(copaw): seed worker agent heartbeat

shiyiyue1102:
## Summary

- enable the CoPaw worker template heartbeat seed at a 10-minute interval
- add the required unreleased changelog entry for the CoPaw image content change

## Validation

- `git diff --check HEAD~1..HEAD`
- `node -e 'JSON.parse(require("fs").readFileSync("copaw/src/copaw_worker/templates/agent.worker.json", "utf8"))'`

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -10,4 +10,5 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 **Bug Fixes**
 
+- **CoPaw Worker heartbeat**: CoPaw worker templates now seed heartbeat at a 10-minute interval so Team Leader agents created from the worker template can run heartbeat turns without requiring an explicit Team CR heartbeat spec.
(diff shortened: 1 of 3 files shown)


maplefeng-a (changes requested):
Thanks for the fix. The behavior change looks reasonable: `agent.worker.json` is the seed for worker-profile agents, and enabling the default heartbeat there should cover Team Leader agents that do not have an explicit Team CR heartbeat spec.

Before merging, could you also update the matching bridge test? `copaw/tests/test_bridge.py::test_worker_template_seeds_default_heartbeat_when_openclaw_silent` still expects `{"enabled": false, "every": "10m"}`, which now contradicts the new template default. Please update that assertion to lock in the intended enabled-by-default behavior.

shiyiyue1102:
Thanks for catching this. Fixed the bridge test assertion so `test_worker_template_seeds_default_heartbeat_when_openclaw_silent` now locks in the enabled-by-default worker heartbeat behavior. I also verified the bridge path locally with a direct assertion; CI is rerunning on the pushed commit.

maplefeng-a (approved):
Thanks for the update. The requested bridge test assertion now matches the enabled-by-default worker heartbeat behavior, and the focused bridge test passes locally. This looks good to me.

PR #920, opened 2026-06-08
feat(controller): support model provider selection

shiyiyue1102:
## Summary
- add spec.modelProvider to Worker, Team leader/member, and Manager CRDs
- resolve Higress model providers and inject the selected gateway URL into runtime config
- authorize AI gateway consumers only on the selected provider route while preserving legacy all-route behavior

## Tests
- go test ./internal/gateway ./internal/agentconfig ./api/v1beta1 ./internal/controller

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -7,6 +7,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 - fix(agent): update file-sharing path guidance for CoPaw and Team Leader agents to use `/root/hiclaw-fs/agents/...` instead of the retired `/root/.hiclaw-worker/...` path.
 
 - **OpenHuman runtime**: OpenHuman added as the fourth Worker runtime with native Matrix support via `channel-matrix` feature flag; includes controller routing (K8s + Docker backends), Dockerfile, entrypoint script, agent template, Helm chart integration, and Makefile build targets.
+- **Multi model providers**: Worker, Team, and Manager specs can now select a Higress model provider via `spec.modelProvider`; the controller resolves the provider, injects the matching gateway URL into runtime config, and authorizes consumers only on the selected AI route.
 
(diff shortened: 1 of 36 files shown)


maplefeng-a (changes requested):
Thanks for the update. The overall direction makes sense, but I think the gateway authorization flow still needs one more fix before this can merge.

When `spec.modelProvider` is set, newly provisioned workers/managers still go through the legacy provisioning path first: `ProvisionWorker` / `ProvisionManager` call `AuthorizeAIRoutes(ctx, consumerName, "")`. In the Higress implementation an empty provider filter keeps the legacy behavior and updates all AI routes. The later provider-specific call only authorizes the selected provider route; it does not remove the consumer from routes that were already authorized by the earlier all-route call.

That means a new worker/manager with `modelProvider` can still remain authorized on every/default route, and changing the provider can leave stale authorization behind. Could you pass the resolved provider ID into the initial/restore gateway auth path, or otherwise deauthorize non-selected routes, and add a test covering the provision/restore path with `modelProvider`?

One smaller API gap: Team leader/member CRDs now have `modelProvider`, but the REST Team create/update request structs do not expose or map that field yet. If Team-level provider selection is meant to be usable through the API/CLI, please wire that path too.

maplefeng-a (approved):
Thanks for the update. The modelProvider auth path now scopes provisioning/restore to the resolved provider, the Higress route filter removes stale non-selected route authorization, and the Team/Manager REST mappings are covered by tests. The targeted controller/server/gateway tests pass locally. Looks good to me.

shiyiyue1102:
@qiacheng7 CI 已全部通过，这个 PR 也已经按 review 反馈更新过了，麻烦帮忙 review 下。

---
@qiacheng7 All CIs have been passed, and this PR has been updated according to review feedback. Please help to review it.

PR #930, opened 2026-06-14
fix(install): skip empty CONTAINER_SOCK mount on Windows+WSL

forcoder:
## Summary

The embedded controller's `podman run` unconditionally mounts `-v "${CONTAINER_SOCK}:/var/run/docker.sock"`. On Windows + WSL the installer cannot auto-detect a Unix socket path (no `systemctl`, no `/run/podman/podman.sock`), so `CONTAINER_SOCK` is empty and podman fails with "host directory cannot be empty" — blocking every install attempt on Windows.

- Wrap the socket mount (and matching `--security-opt label=disable`) in a check for non-empty `CONTAINER_SOCK`
- Add an INFO log when running embedded mode
- Add a WARNING log when `CONTAINER_SOCK` is empty, telling users Worker creation won't work until a real Docker/Podman socket is available

## Test plan

- [ ] On a Windows+WSL host, run `HICLAW_NON_INTERACTIVE=1 ./install/hiclaw-install.sh manager` — should no longer fail with "host directory cannot be empty"
- [ ] On a Linux host with a working podman socket, verify the socket is still mounted as before
- [ ] On a Linux host without a podman socket (e.g. `HICLAW_MOUNT_SOCKET=0`), verify the controller starts with a warning

### DIFF
--- a/install/hiclaw-install.sh
+++ b/install/hiclaw-install.sh
@@ -2840,8 +2840,15 @@ EOF
         CONTAINER_SOCK=$(detect_socket)
         if [ -n "${CONTAINER_SOCK}" ]; then
             log "$(msg install.socket_detected "${CONTAINER_SOCK}")"
-            SOCKET_MOUNT_ARGS="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
-        else
+            # Docker/Podman API socket -- only mount when detection succeeded.
+            # On Windows + WSL the installer cannot auto-detect a Unix socket
+            # path, so CONTAINER_SOCK may be empty.
+            local _socket_args=""
+            if [ -n "${CONTAINER_SOCK}" ]; then
+                _socket_args="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
+            fi
+
+            ${_socket_args} \n        else
             log "$(msg install.socket_not_found)"
             # Interactive confirmation when socket not found
             if [ "${HICLAW_NON_INTERACTIVE}" != "1" ]; then
@@ -3135,6 +3142,10 @@ CREDEOF
     if [ "${HICLAW_USE_EMBEDDED}" = "1" ]; then
         # ============================================================
         # New architecture: embedded controller + auto-created manager
+        log "INFO: using embedded controller architecture."
+        if [ -z "${CONTAINER_SOCK}" ]; then
+            log "WARNING: CONTAINER_SOCK is empty -- embedded controller will start but Workers cannot be created."
+        fi
         # ============================================================
 
         # Internal port: 8080 (Higress gateway inside the container).


install/hiclaw-install.sh
```diff
@@ -2840,8 +2840,15 @@ EOF
...
-        else
+            # Docker/Podman API socket -- only mount when detection succeeded.
+            # On Windows + WSL the installer cannot auto-detect a Unix socket
+            # path, so CONTAINER_SOCK may be empty.
+            local _socket_args=""
+            if [ -n "${CONTAINER_SOCK}" ]; then
+                _socket_args="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
+            fi
+
+            ${_socket_args} \n        else
```
oss-maintainer, line 2851: This line is corrupted: it contains a literal `\n` and has swallowed the `else` keyword as a command argument:

```bash
${_socket_args} \n        else
```

Consequences:
1. The outer `if [ -n "${CONTAINER_SOCK}" ]` loses its `else` branch — the "socket not found" path (log + interactive confirmation) now executes when the socket IS detected, and is skipped when it is not (the opposite of the intent).
2. `${_socket_args}` is unquoted in command position: when the socket is detected it word-splits into `-v /path:/var/run/docker.sock --security-opt label=disable` and bash tries to execute `-v` as a command → exit 127 → with `set -e` (line 50) the installer aborts on every Linux/macOS host where socket detection succeeds.

It looks like an intended structure such as:

```bash
CONTAINER_SOCK=$(detect_socket)
local _socket_args=""
if [ -n "${CONTAINER_SOCK}" ]; then
    log "$(msg install.socket_detected "${CONTAINER_SOCK}")"
    _socket_args="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
else
    log "$(msg install.socket_not_found)"
    ...
fi
```

Please re-apply this hunk and verify with `bash -n` plus a test run of both branches.

install/hiclaw-install.sh
```diff
@@ -2840,8 +2840,15 @@ EOF
...
         if [ -n "${CONTAINER_SOCK}" ]; then
             log "$(msg install.socket_detected "${CONTAINER_SOCK}")"
-            SOCKET_MOUNT_ARGS="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
-        else
+            # Docker/Podman API socket -- only mount when detection succeeded.
+            # On Windows + WSL the installer cannot auto-detect a Unix socket
+            # path, so CONTAINER_SOCK may be empty.
+            local _socket_args=""
+            if [ -n "${CONTAINER_SOCK}" ]; then
+                _socket_args="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
```
oss-maintainer, line 2848: This replaces the `SOCKET_MOUNT_ARGS="-v ..."` assignment with `_socket_args`, but `SOCKET_MOUNT_ARGS` is still the variable consumed by the container run command (~line 3698), and `_socket_args` is never consumed anywhere. Net effect: even when socket detection succeeds, the Docker/Podman API socket is no longer mounted — a regression on the primary platforms. Either keep assigning `SOCKET_MOUNT_ARGS` here or switch the run command to consume `_socket_args` (e.g. via an array, since word-splitting an unquoted string of flags is fragile).

install/hiclaw-install.sh
```diff
@@ -2840,8 +2840,15 @@ EOF
         CONTAINER_SOCK=$(detect_socket)
         if [ -n "${CONTAINER_SOCK}" ]; then
             log "$(msg install.socket_detected "${CONTAINER_SOCK}")"
-            SOCKET_MOUNT_ARGS="-v ${CONTAINER_SOCK}:/var/run/docker.sock --security-opt label=disable"
-        else
+            # Docker/Podman API socket -- only mount when detection succeeded.
+            # On Windows + WSL the installer cannot auto-detect a Unix socket
+            # path, so CONTAINER_SOCK may be empty.
+            local _socket_args=""
+            if [ -n "${CONTAINER_SOCK}" ]; then
```
oss-maintainer, line 2847: This inner `if [ -n "${CONTAINER_SOCK}" ]` is redundant: the enclosing branch (line 2841) already guarantees `CONTAINER_SOCK` is non-empty. The empty case (Windows+WSL) is handled in the `else` branch, so the fix likely belongs around where `SOCKET_MOUNT_ARGS` is consumed rather than here.

install/hiclaw-install.sh
```diff
@@ -3135,6 +3142,10 @@ CREDEOF
     if [ "${HICLAW_USE_EMBEDDED}" = "1" ]; then
         # ============================================================
         # New architecture: embedded controller + auto-created manager
+        log "INFO: using embedded controller architecture."
+        if [ -z "${CONTAINER_SOCK}" ]; then
```
oss-maintainer, line 3146: This warning fires whenever `CONTAINER_SOCK` is empty, including when socket mounting is intentionally disabled (`HICLAW_MOUNT_SOCKET != 1`), which would be a misleading warning. Consider guarding it with the mount flag, e.g. `if [ "${HICLAW_MOUNT_SOCKET}" = "1" ] && [ -z "${CONTAINER_SOCK}" ]; then`.

oss-maintainer (changes requested):
### Summary
This change intends to skip the Docker/Podman socket mount when `CONTAINER_SOCK` is empty (Windows+WSL), but the core edit is corrupted and breaks the installer on every platform, so I cannot approve it as-is. Two blocking issues are annotated inline: (1) the line `${_socket_args} \n        else` contains a literal `\n` and a misplaced `else` — a bash syntax/runtime failure (with `set -e` the installer aborts when the socket IS detected); (2) `SOCKET_MOUNT_ARGS` is no longer assigned anywhere while the container run command still consumes it, so the socket mount is silently dropped even when detection succeeds. A suggested corrected structure is in the inline note. Requesting changes — please re-apply this hunk and verify with `bash -n` plus a test run of both the socket-detected and socket-not-detected branches.

---

oss-maintainer:
⚠️ **Merge conflict detected**

This PR has conflicts with the `main` branch and cannot be merged. Please rebase or merge `main` into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix/windows-wsl-installer
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review after conflicts are resolved.

---
*Automated notification by github-manager-bot*

PR #932, opened 2026-06-15
fix(helm): default chart images to v1.1.2

googs1025:
## Summary
- Bump the Helm chart version/appVersion to 1.1.2 so default image tags resolve to v1.1.2.


shiyiyue1102:
Thank you for preparing the v1.1.2 chart defaults. The release line has moved on to v1.2.0 beta, and Helm release packaging now supplies chart and app versions from the release workflow rather than relying on this old static bump. Merging v1.1.2 now would move current release metadata backward. We are closing this PR as superseded by the current release process.

PR #933, opened 2026-06-16
feat(plugins): add AgentTeams plugin packaging CLI

maplefeng-a:
## Summary
- add a generic AgentTeams plugin package contract under `plugins/`
- add the `agentteams plugin` fallback CLI for install/list/update/uninstall flows
- add plugin validation/package scripts and CLI integration tests using a synthetic demo plugin fixture

## Boundaries
- this PR only introduces the generic plugin package/CLI layer
- it does not include TeamHarness, QwenPaw, Claude, or LoongSuite-specific plugin content; those will be split into follow-up PRs

## Tests
- `python3 -m compileall -q plugins/cli/src`
- `python3 plugins/tests/cli/test_agentteams_plugin_cli.py`
- `git diff --check`

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -42,6 +42,11 @@ Plugins are distributed as tarballs whose root is the plugin content itself:
     └── uninstall.sh
 ```
 
+`metadata.name` must use 1-63 lowercase letters, digits, `-`, or `_`, starting
+and ending with a letter or digit. Manifest-controlled paths (`package.include`,
+prompts, skills, MCP args, and adapters) must be relative paths inside the
+plugin root and must not contain `..` segments.
+
 The lifecycle entrypoints are the shared contract:
 
 - `scripts/install.sh`: detects supported local runtimes and dispatches to the
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -1,0 +1,70 @@
+# HiClaw Plugins
+
+This directory contains HiClaw runtime plugin packages and the local fallback
+CLI for installing them.
+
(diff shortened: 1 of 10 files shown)


shiyiyue1102:
Thanks for the update. The plugin packaging direction looks reasonable, but I think we need to block merge until plugin names and manifest-controlled paths are constrained.

Right now `metadata.name` / the CLI plugin name is used directly in `.agentteams/plugins/<name>`. A name such as `../../../escaped-plugin` can install content outside the project-local `.agentteams/plugins` tree and still report success. The same class of validation should also cover manifest paths such as `package.include`, prompts, skills, MCP args, and adapter paths so package contents cannot escape the plugin root.

Could you add a safe plugin-name pattern and reject absolute paths / `..` segments for manifest-controlled paths, plus regression tests for these negative cases? After this path-boundary issue is fixed, this should be ready to merge.

maplefeng-a:
Thanks for catching this. Fixed by adding a safe plugin-name pattern before install/update/uninstall paths are derived, and by rejecting absolute paths or '..' segments for manifest-controlled paths in package.include, prompts, skills, MCP args, and adapter paths. Added regression coverage for the negative cases as well.

shiyiyue1102 (approved):
LGTM. The path-boundary blocker is fixed: plugin names are validated before install/update/uninstall paths are derived, manifest-controlled paths reject absolute paths and '..' segments, and regression coverage now covers the negative cases. I also verified the current head locally with compileall, the plugin CLI test suite, and git diff --check.

PR #934, opened 2026-06-16
feat(manager): support STS auth in skill discovery

maplefeng-a:
## Summary
- add `NACOS_AUTH_TYPE=sts-hiclaw` support to the manager worker skill discovery wrappers
- request STS credentials from the HiClaw controller and pass them to the Nacos CLI
- extend the find-skill regression tests to cover worker, CoPaw worker, and Hermes worker scripts

## Scope
- independent from #933; this only changes manager skill discovery wrappers and their shell regression test
- does not include TeamHarness plugin, QwenPaw worker runtime, or external CLI harness runtime changes

## Tests
- `bash manager/tests/test-hiclaw-find-skill.sh`
- `bash -n manager/agent/worker-agent/skills/find-skills/scripts/hiclaw-find-skill.sh`
- `bash -n manager/agent/copaw-worker-agent/skills/find-skills/scripts/hiclaw-find-skill.sh`
- `bash -n manager/agent/hermes-worker-agent/skills/find-skills/scripts/hiclaw-find-skill.sh`
- `bash -n manager/tests/test-hiclaw-find-skill.sh`
- `git diff --check`

### DIFF
--- a/manager/agent/copaw-worker-agent/skills/find-skills/scripts/hiclaw-find-skill.sh
+++ b/manager/agent/copaw-worker-agent/skills/find-skills/scripts/hiclaw-find-skill.sh
@@ -69,6 +69,57 @@ get_skills_api_url() {
     printf '%s\n' "https://skills.sh"
 }
 
+resolve_controller_bearer() {
+    if [ -n "${HICLAW_AUTH_TOKEN:-}" ]; then
+        printf '%s' "${HICLAW_AUTH_TOKEN}"
+        return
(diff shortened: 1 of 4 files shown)


shiyiyue1102 (approved):
LGTM. Compatibility checked; existing Nacos auth paths are unchanged unless NACOS_AUTH_TYPE=sts-hiclaw is explicitly set, and the STS bearer path matches the existing controller credential helper. CI is passing.

PR #935, opened 2026-06-16
fix(shared): pass cluster id to STS credential refresh

maplefeng-a:
## Summary
- include `X-HiClaw-Cluster-ID` when shared OSS credential refresh requests controller STS credentials
- keep the existing request shape unchanged when `HICLAW_CLUSTER_ID` is empty
- add a standalone shell regression test with mocked `curl` and `jq`

## Scope
- independent from #933 and #934; this only changes the shared mc/OSS credential helper
- does not include QwenPaw runtime, TeamHarness plugin, install/build, or controller changes

## Tests
- `tests/test-oss-credentials.sh`
- `bash -n shared/lib/oss-credentials.sh`
- `bash -n tests/test-oss-credentials.sh`
- `git diff --check`

### DIFF
--- a/shared/lib/oss-credentials.sh
+++ b/shared/lib/oss-credentials.sh
@@ -59,9 +59,16 @@ _oss_refresh_sts_via_controller() {
         return 1
     fi
 
-    resp=$(curl -s -w "\n%{http_code}" -X POST "${_controller_url}/api/v1/credentials/sts" \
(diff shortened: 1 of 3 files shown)


shiyiyue1102 (approved):
LGTM. Compatibility checked; HICLAW_CLUSTER_ID only adds the controller header when set, and the empty-cluster path preserves the existing STS request shape. CI is passing.

PR #939, opened 2026-06-16
feat(teamharness): add TeamHarness plugin core

maplefeng-a:
## Summary
- add the TeamHarness plugin manifest, prompts, role skills, MCP server, and runtime adapters
- add QwenPaw and Claude Code adapter packaging/install hooks
- add TeamHarness contract, adapter, and MCP tool tests

## Tests
- bash plugins/tests/run-integration-tests.sh
- ruby plugins/tests/teamharness/mcp/tools/test-protocol.rb
- ruby plugins/tests/teamharness/mcp/tools/test-roomflow.rb

### DIFF
--- a/docs/teamharness-boundary-and-contracts.md
+++ b/docs/teamharness-boundary-and-contracts.md
@@ -1,0 +1,98 @@
+# TeamHarness v0.1 Boundary And Contracts
+
+This document defines the functional boundary for the current TeamHarness
+plugin package.
+
+TeamHarness v0.1 is the runtime-neutral team collaboration base. It packages
+stable prompts, team skills, MCP tools, lifecycle scripts, and runtime adapter
+entrypoints. It does not own worker lifecycle, controller reconciliation,
+worker desired-state apply loop, runtime hook behavior, or
+periodic workspace persistence.
+
+## Responsibilities
+
+TeamHarness owns:
+
+- Stable team collaboration prompts and role prompts.
+- Team collaboration skills for organization, communication, shared files,
+  projects, task delegation, and task execution.
+- Explicit MCP tools for team messages, shared files, project flow, task flow,
+  and plugin health.
+- A single plugin tarball installed through the HiClaw `agentteams` CLI by
+  default and compatible with LoongSuite `plugin-probe` for local runtimes.
+
+TeamHarness does not own:
+
+- Controller generation of `shared/runtime/members/{runtimeName}/runtime.yaml`.
+- Worker process lifecycle, pod restart, or runtime process supervision.
+- Worker desired-state parsing, polling, apply, or diagnostics.
+- Runtime-neutral top-level hooks.
+- Runtime hook trigger contracts, payload formats, and enforcement behavior.
+- Credential access enforcement that depends on runtime-specific file or tool
+  guard support.
+- AgentSpec package download, apply, rollback, or update inside the worker desired-state apply loop.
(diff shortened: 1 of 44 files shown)


shiyiyue1102:
Thanks for the TeamHarness update. The overall direction looks reasonable, but before merging could you double-check two compatibility paths?

1. `filesync` credentials in the QwenPaw MCP server: the `HICLAW_CONTROLLER_URL` + `HICLAW_AUTH_TOKEN_FILE` propagation path looks OK for the cloud/controller side, but the current `filesync` implementation still ultimately runs `mc` with either an existing `MC_HOST_hiclaw` or `HICLAW_FS_ENDPOINT` + `HICLAW_FS_ACCESS_KEY` + `HICLAW_FS_SECRET_KEY`. The QwenPaw adapter env does not appear to pass `HICLAW_FS_ACCESS_KEY`, `HICLAW_FS_SECRET_KEY`, or `MC_HOST_hiclaw` into the MCP client env. Could you confirm that QwenPaw MCP inherits those values from the parent environment, or add the missing propagation/refresh logic and a test so both cloud STS and local/static credential modes work?

2. The output sanitizer hook directly reads `QwenPawAgent._acting`. Since this is a private API, please add a small `hasattr`/fallback guard so the plugin can degrade safely if QwenPaw changes that method name after 1.1.7, instead of failing during plugin registration.

maplefeng-a:
Thanks for catching these. Fixed both compatibility paths:

- the QwenPaw adapter now passes HICLAW_FS_ACCESS_KEY, HICLAW_FS_SECRET_KEY, and MC_HOST_hiclaw through to the TeamHarness MCP client env, alongside the existing controller/STS refresh inputs, with adapter test coverage;
- the sanitizer hook now checks for QwenPawAgent._acting and degrades cleanly if that private hook is unavailable.

Re-ran:
- python3 -m pytest plugins/tests/teamharness/adapters/qwenpaw/test_adapter.py -q
- bash plugins/tests/run-integration-tests.sh
- ruby plugins/tests/teamharness/mcp/tools/test-protocol.rb
- ruby plugins/tests/teamharness/mcp/tools/test-roomflow.rb

shiyiyue1102:
The implementation looks aligned with the intended behavior. One remaining process blocker is that this PR changes `plugins/**`, but the current automatic PR CI path filters do not cover plugin-only changes, so GitHub does not report PR checks for this branch.

Could you add a dedicated lightweight plugin workflow, for example `.github/workflows/test-plugins.yml`, triggered by:

```yaml
on:
  pull_request:
    paths:
      - 'plugins/**'
      - '.github/workflows/test-plugins.yml'
```

It should run the existing plugin validations/tests, such as plugin metadata validation, TeamHarness contract tests, MCP tool tests, and the QwenPaw adapter pytest suite. That should give plugin PRs their own required checks without forcing the full image/integration workflow for every plugin-only change.

Once that CI coverage is in place and passing, this PR should be ready to move forward.

shiyiyue1102 (approved):
LGTM.

I re-checked the TeamHarness plugin core and the previously raised QwenPaw compatibility concerns. The MCP client now propagates the filesync credential envs, and the sanitizer hook now degrades safely when the private QwenPaw `_acting` hook is unavailable. The related adapter tests cover both paths.

GitHub still does not report checks for this plugin-only PR because the workflow path filter does not cover `plugins/**`; merging is based on the source review and maintainer confirmation.

PR #964, opened 2026-07-01
fix(controller): keep modelProvider auth in reconcilers

shiyiyue1102:
## Summary

- keep provider-specific AI route authorization in the Manager/Worker/Team reconcilers
- remove `ModelProviderID` from provisioning requests and gateway-auth restore interfaces
- leave provisioning on the default gateway authorization path so provider scoping is not duplicated across controller and service layers
- update mocks and focused assertions for the new boundary

## Tests

- `go test ./internal/controller ./internal/service ./test/testutil/mocks`
- `go test ./...`


shiyiyue1102:
@qiacheng7 CI is green now. Could you take a look when convenient?

qiacheng7 (approved):

PR #973, opened 2026-07-02
fix(config): pass custom model parameters to controller

### LINKED ISSUE #961: manager配置反复重置 || manager configuration is reset repeatedly
**背景：**
版本：HiClaw v1.1.2
Manager Runtime：OpenClaw
LLM 配置：【OpenAI兼容】百炼平台API，使用qwen3.7-plus

**问题：**
在安装时，已经配置最大上下文长度及最大输出长度，但是Manager的启动后并未带入该环境变量，导致其上下文长度为150K，触发压缩频繁，不管是通过修改openclaw.json或者通过IM修改MinIO及openclaw.json的配置，都会在几分钟内被重置。

**诉求：**
期望给出可实现修改方案

Image

---
**Background:**
Version: HiClaw v1.1.2
Manager Runtime:OpenClaw
LLM configuration: [OpenAI compatible] Bailian platform API, using qwen3.7-plus

**Question:**
During installation, the maximum context length and maximum output length have been configured, but the environment variable was not brought into the Manager after startup, resulting in a context length of 150K, triggering compression frequently. Whether it is through modifying openclaw.json or modifying the configuration of MinIO and openclaw.json through IM, it will be reset within a few minutes.

**Appeal:**
It is expected that a feasible modification plan will be provided

Image

RerankerGuo:
## Summary
- pass custom model context/max-token/reasoning/vision env vars into the embedded controller container
- expose the same model parameter overrides in Helm values and render them into the runtime Secret
- teach controller config to read reasoning/vision overrides and forward all model parameters to agentconfig

Fixes #961.

## Tests
- bash -n install/hiclaw-install.sh
- git diff --check

- Go 1.25.0: `go test ./internal/config`
Not run: `helm template ...` because Helm is not installed (`helm: command not found`).

## Temporary dependency

- This branch is temporarily stacked on #985 (`78e9385`), whose full CI passed all 10 integration matrices.
- Relative to #985, this PR retains only its original scoped changes; the shared test fix disappears from the diff after #985 merges.
- Fresh checks on `4fc9eaa4` passed every build and all 10 integration matrices.


RerankerGuo:
@shiyiyue1102 Hi! `SHARD_A copaw/hermes` failed on this PR's CI. Same pattern as #984 — in the copaw/hermes case the `hiclaw-worker-bob` session directory does not exist at all and Manager never invoked `hiclaw create worker --name bob` — Manager only created `alice`. Failure mode is a test-prompt interpretation flake, not the model-parameter env wiring in this PR.

This PR only changes Helm values + embedded install scripts + controller config to pass custom model context window / max tokens / reasoning / vision env vars. It doesn't touch the Manager agent runtime or worker-create code path. Could you re-run CI or merge if you agree the flake is unrelated?

shiyiyue1102:
Passing model context, token, reasoning, and vision overrides through every deployment path has product value, but the current PR is broad and conflicts with main. It also overlaps newer model-vision and QwenPaw work. Please rebase and verify which pieces are still missing, then either narrow this PR to one coherent model-parameter contract or split installer and controller changes if they can land independently. The updated branch should cover Helm, embedded Bash, PowerShell, config parsing, and nil versus explicit-false behavior with focused tests.

RerankerGuo:
Closing because this branch spans controller config, Helm, Bash, PowerShell, and integration setup while overlapping newer model-capability work such as #1103. Any remaining passthrough gap should be reproduced on current main and split into a coherent minimal contract patch.

PR #979, opened 2026-07-02
fix(copaw): retry prompt reads during worker startup

### LINKED ISSUE #711: [Bug] copaw worker can crash reading partially-written merged AGENTS.md during apply --zip update
## Summary

`copaw/src/copaw_worker/worker.py:123` reads `AGENTS.md` from
`self.sync.local_dir` via `read_text()` with no retry or size check.
On `hiclaw apply worker --zip` against an **existing** copaw worker
(update path), this read appears to race with the sync's merged-file
write, producing `UnicodeDecodeError` mid multi-byte character; the
container exits with code 1 and only recovers on a manual
`docker start`.

## Environment

- HiClaw v1.1.0 GA, embedded mode, macOS Docker Desktop
- Worker: copaw runtime, package contains `config/SOUL.md` + `config/AGENTS.md`
- Reproduces on `apply --zip` *update* (existing worker), not on initial create

## Reproduction

1. Have a copaw worker already running with package A
2. Build package B with a modified `config/AGENTS.md`
3. `docker cp B.zip hiclaw-controller:/tmp/`
4. `docker exec hiclaw-controller hiclaw apply worker --zip /tmp/B.zip --name <existing-worker>`
5. Within ~10s: `docker ps -a --filter name=hiclaw-worker-<name>` shows `Exited (1)`

## Observed traceback (excerpt)

```
File ".../copaw_worker/worker.py", line 123, in start
    (workspace_dir / name).write_text(src.read_text())
File ".../pathlib.py", line 1059, in read_text
    return f.read()
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 19964-19965:
  unexpected end of data
```

## Investigation (evidence, not proof)

The race-with-sync explanation is the most consistent reading of the
evidence, but I have not done a source-level confirmation that the two
writers/readers actually overlap. What we observed:

- The merged `agents/<worker>/AGENTS.md` in MinIO **is** valid UTF-8
  when fetched post-mortem (`mc cp ...` after the crash; complete +
  decodable end-to-end)
- File size at the error position (~19964 bytes) is suspiciously
  close to the *previous* merged file size, consistent with
  `worker.start()` reading partway into the new prefix during the
  merger's write
- `docker start <container>` recovers cleanly; second boot sees the
  fully-written file

## Suggested directions

1. **Atomic merge writes** — write the merged file to `*.tmp` then
   `os.replace`, so partial reads are structurally impossible. Most
   robust: addresses the root cause regardless of who reads when
2. **Bounded retry on `UnicodeDecodeError`** in `worker.py:123` —
   surgical fallback that hardens the read site without changing the
   sync side
3. Explicit "sync ready" signal that `worker.start()` waits on

(1) seems cleanest as a primary fix; (2) is reasonable defense-in-depth.

## Workaround (current)

After `apply --zip` update, wait ~10s. If the worker container shows
`Exited (1)`, run `docker start <container-name>`. Second boot
succeeds because the merge is complete on disk by then. This is what
we currently document for our deployment (haopaw).

RerankerGuo:
## Summary
- add a bounded stable UTF-8 prompt reader for CoPaw worker SOUL.md / AGENTS.md materialization
- use it during startup and config re-bridge so transient mc mirror/read races do not crash the worker
- add focused unit coverage and changelog entry

## Tests
- git diff --check
- python3 -m py_compile copaw/src/copaw_worker/worker.py copaw/tests/test_worker_health.py
- not run: python3 -m pytest copaw/tests/test_worker_health.py -q (pytest is not installed in this environment)

Fixes #711.
Also addresses #728.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,21 +4,8 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 ---
 
-- feat(qwenpaw): add the QwenPaw worker runtime Python package baseline with runtime config sync, storage sync, heartbeat reporting, Matrix channel overlay, and focused unit tests.
-- fix(controller): surface Kubernetes Pod container failures in Worker backend status and status API responses.
-- feat(controller): expose low-cardinality AgentTeams controller metrics and optional Helm ServiceMonitor.
-- feat(controller): add Matrix AppService Human SSO identity provisioning with hash-derived Matrix IDs, AppService login, deletion deactivation, and Team admin/member identity resolution from Human status.
 - fix(agent): update file-sharing path guidance for CoPaw and Team Leader agents to use `/root/hiclaw-fs/agents/...` instead of the retired `/root/.hiclaw-worker/...` path.
-- feat(helm): add a Helm LLM preflight hook and reusable `hiclaw llm-preflight` command to validate API key/base URL/model before controller startup.
-
-- fix(copaw): harden Matrix channel control-command handling, task-thread routing, NO_REPLY suppression, and cancellation noise handling.
-- feat(controller): add OpenKruise Sandbox backend support for Workers via `spec.backendRuntime=sandbox`, including SandboxClaim lifecycle, status watches, CRD schema, and Helm RBAC/env wiring.
-- fix(controller): materialize sandbox Worker runtime env/auth material into worker-deps storage before creating SandboxClaim deps and block legacy pod-to-sandbox runtime switches until the Worker is stopped.
 - feat(controller): add per-agent `spec.resources` support for Manager, Worker, Team Leader, and Team Worker CRDs.
-- fix(worker): pass `X-HiClaw-Cluster-ID` when remote Workers refresh controller-issued STS credentials for OSS and Nacos AI registry access.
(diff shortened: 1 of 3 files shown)


RerankerGuo:
@shiyiyue1102 Hi! This PR addresses #711 / #728. Just rebased onto current `main` (was conflicting) — CI is now green (18/18). Adds bounded stable UTF-8 reads for `SOUL.md` / `AGENTS.md` during CoPaw startup and re-bridge, with replacement-decoding fallback. Could you take a look when convenient?

shiyiyue1102:
The bounded prompt-read retry addresses a plausible startup race, but the branch conflicts with main and the behavior needs to be reviewed against the current CoPaw startup and config bridge. Please rebase, keep the change limited to stable prompt reads, and add a focused test that simulates a partial or transiently invalid UTF-8 write followed by a valid file. The retry limits and failure behavior should be explicit so a permanently corrupt prompt does not silently delay startup without a useful error.

RerankerGuo:
Closing because the CoPaw/QwenPaw startup and configuration bridge changed substantially after this branch. The startup-race hypothesis needs to be reproduced against current main before carrying forward a retry policy.

Issue #867, opened 2026-05-27
如何提升自定义路由的优先级 || How to improve the priority of custom routing


SKY-ALT-TECH:
<img width="1624" height="508" alt="Image" src="https://github.com/user-attachments/assets/f80c9230-4d22-4e0b-965d-8efd6e9c8000" />

<img width="1910" height="873" alt="Image" src="https://github.com/user-attachments/assets/d43149d1-daef-43ab-a001-c4f1decc2fe6" />
手动在higress控制台配置的路由总是无法匹配上，走的永远是默认路由。是我配置的不对，还是有什么可以定义路由优先级的方式？

---
<img width="1624" height="508" alt="Image" src="https://github.com/user-attachments/assets/f80c9230-4d22-4e0b-965d-8efd6e9c8000" />

<img width="1910" height="873" alt="Image" src="https://github.com/user-attachments/assets/d43149d1-daef-43ab-a001-c4f1decc2fe6" />
The route configured manually in the higress console always fails to match, and the default route is always used. Is my configuration wrong, or is there any way to define routing priority?

shiyiyue1102:
HiClaw 初始化时会创建默认 AI Route（`default-ai-route`），Manager / Worker 的 gateway consumer 会被授权到 AI routes 上。手工在 Higress 控制台新增 route 后，如果请求始终走默认路由，通常需要一起检查 route 的匹配条件、path/domain、provider 绑定，以及 `authConfig.allowedConsumers` 是否包含对应的 consumer。

能否贴一下这个自定义 AI Route 的完整配置 JSON，或至少补充 route path / domain、provider、model match 条件和 allowedConsumers？这样可以判断是匹配优先级问题，还是 route 条件没有命中。

---
HiClaw will create a default AI Route (`default-ai-route`) when it is initialized, and the gateway consumer of Manager/Worker will be authorized to AI routes. After manually adding a route in the Higress console, if the request always takes the default route, you usually need to check the matching conditions of the route, path/domain, provider binding, and whether `authConfig.allowedConsumers` contains the corresponding consumer.

Can you post the complete configuration JSON of this custom AI Route, or at least add the route path/domain, provider, model match conditions and allowedConsumers? This can determine whether it is a matching priority problem or whether the route condition is not hit.

RerankerGuo:
I can add a small FAQ clarification for this: custom AI routes need explicit model matching, and the default route must also be constrained when multiple routes exist; otherwise the default route can keep matching requests first.

[...]

PR #980, opened 2026-07-02
docs: clarify Higress AI route matching

RerankerGuo:
## Summary
- add English and Chinese FAQ entries for custom Higress AI routes that never match
- explain that an unconstrained default-ai-route can match all model requests
- document that multiple AI routes should each use explicit modelPredicates, including the default route

## Tests
- git diff --check

Addresses #867.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -14,6 +14,7 @@
 - [Cannot connect to Matrix server locally](#cannot-connect-to-matrix-server-locally)
 - [How to talk to a Worker directly](#how-to-talk-to-a-worker-directly)
 - [How to connect third-party, local, or multi-provider models](#how-to-connect-third-party-local-or-multi-provider-models)
+- [Why does my custom Higress AI route never match](#why-does-my-custom-higress-ai-route-never-match)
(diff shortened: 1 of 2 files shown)


RerankerGuo:
@shiyiyue1102 Hi! This PR documents the fix for #867 (custom AI routes never matching). It explains that an unconstrained `default-ai-route` can match every model request, and recommends setting explicit `modelPredicates` on every route (including the default) when running multi-route setups. English + Chinese FAQ entries.

`translate` is green and the branch is MERGEABLE. Docs-only. Could you take a look?

shiyiyue1102 (approved):
LGTM. The FAQ guidance matches the current gateway behavior and the built-in Higress management skill: an unconstrained default-ai-route can match every model, while multi-provider routing should use explicit, non-overlapping modelPredicates on both custom and default routes. The English and Chinese entries are aligned and the docs-only diff is clean.

---

LGTM。FAQ 指引与当前网关行为及内置 Higress 管理 skill 一致：未限制的 default-ai-route 可能匹配所有模型；多供应商路由应在自定义路由和默认路由上都配置明确且不重叠的 modelPredicates。中英文内容对齐，文档差异干净。

Issue #626, opened 2026-04-15
Security review: 3 validated findings from ClawGuard scan


ClawSafe:
Hi, I ran a ClawGuard scan on this repo and manually validated a few concrete findings that look worth tightening:

1. `README.md:49`
   - The quick-start docs advertise `curl | bash` as the setup path.
   - Exact text: `One Command Setup: curl | bash and you're done`
   - This normalizes remote-script execution without pinning or integrity verification.

2. `copaw/src/copaw_worker/sync.py:448`
   - The sync cleanup path removes directories with `shutil.rmtree(child)`.
   - In context this is deleting local skill directories based on remote state, which deserves extra path safety checks and guardrails before recursive delete.

3. `copaw/src/copaw_worker/worker.py:466`
   - Another cleanup path removes directories with `shutil.rmtree(child)` when skills are considered stale.
   - Same concern here: recursive delete on dynamically selected paths should be tightly constrained and validated.

I skipped several scanner hits that looked like likely false positives and only kept the ones above because they were easy to confirm directly.

If useful, I can share the full scan output in a follow-up.

Scanner: https://github.com/hesoyam2221/clawguard

shiyiyue1102:
抱歉回复晚了。@maplefeng-a @kkqqqqqq 这个看起来属于 CoPaw / Worker Runtime 安全边界方向。我检查了当前源码，`shutil.rmtree` 清理路径仍需要结合运行时路径约束再判断，暂时不能确认 v1.1.2 已完整覆盖这里的安全建议，麻烦有空帮忙看下。

---
Sorry for the late reply. @mapplefeng-a @kkqqqqqqq This looks to be in the CoPaw/Worker Runtime security boundary direction. I checked the current source code and found that the `shutil.rmtree` cleanup path still needs to be combined with the runtime path constraints to determine. I cannot confirm that v1.1.2 has fully covered the security recommendations here. Please help me take a look when you have time.

shiyiyue1102:
这条有具体安全扫描发现，涉及 CoPaw 同步和运行时文件处理。@maplefeng-a 麻烦看下哪些是真问题、哪些是扫描噪声，先把可落地部分筛出来。

---
This particular security scan found this, involving CoPaw synchronization and runtime file processing. @maplefeng-a Please check which ones are real problems and which ones are scanning noise. First, screen out the parts that can be landed.

RerankerGuo:
[...]

PR #982, opened 2026-07-02
fix(copaw): guard managed skill cleanup

RerankerGuo:
## Summary

- add a managed-directory removal guard for CoPaw skill cleanup paths
- refuse recursive deletion unless the target is a non-symlink direct child of the expected managed directory
- cover the sync cleanup guard with unit tests

Addresses the recursive-delete hardening items in #626.

## Verification

- `python3 -m py_compile copaw/src/copaw_worker/sync.py copaw/src/copaw_worker/worker.py copaw/tests/test_sync.py`
- `PYTHONPATH=/tmp/agentteams-pytest:copaw/src python3 -m pytest copaw/tests/test_sync.py` (15 passed)
- `git diff --check`

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,21 +4,8 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 ---
 
-- feat(qwenpaw): add the QwenPaw worker runtime Python package baseline with runtime config sync, storage sync, heartbeat reporting, Matrix channel overlay, and focused unit tests.
-- fix(controller): surface Kubernetes Pod container failures in Worker backend status and status API responses.
-- feat(controller): expose low-cardinality AgentTeams controller metrics and optional Helm ServiceMonitor.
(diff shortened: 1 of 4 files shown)


RerankerGuo:
@shiyiyue1102 Hi! This PR addresses #626. Just rebased onto current `main` (was conflicting) — CI is now green (18/18). Adds managed child-directory guards around CoPaw recursive skill cleanup; refuses deletion unless the target is a non-symlink direct child of the expected managed directory. Could you take a look when convenient?

shiyiyue1102:
The intent is reasonable defense in depth, but the current value needs stronger evidence. These deletion candidates already come from direct-child iteration, and Python shutil.rmtree refuses a directory symlink rather than following it, so the concrete escape or data-loss path on current main is not yet clear. Please provide a failing reproduction against main, rebase the conflicted branch, and avoid duplicating the removal helper in both sync.py and worker.py by using one shared implementation. Otherwise this is more complexity than the demonstrated risk warrants.

RerankerGuo:
Closing because the requested failing reproduction against current main is still absent, while direct-child iteration and shutil.rmtree behavior undermine the proposed escape scenario. A new focused report can be opened if a concrete data-loss path is reproduced.

Issue #888, opened 2026-05-30
v1.1.2 ERROR: Higress Gateway not ready after 120s


h31105:
[HiClaw] 正在启动 Manager 容器...
0da0f6ae6672e08e5931a6290fec48f7dd916ea009f3b1d3ee0b925241b1f54b
[HiClaw] Embedded controller started: hiclaw-controller
[HiClaw] Waiting for Tuwunel (Matrix)...
[HiClaw] Tuwunel (Matrix) is ready (2s)
[HiClaw] Waiting for MinIO...
[HiClaw] MinIO is ready (0s)
[HiClaw] Waiting for Higress Gateway...
[HiClaw] ERROR: Higress Gateway not ready after 120s

gateway 日志如下：

Mode=full
O11Y components are not installed.
O11Y=off
USE_PLUGIN_SERVER=on
Checking the readiness of Higress Pilot...
2026-05-30T18:46:24.347271Z     info    FLAG: --concurrency="16"
2026-05-30T18:46:24.347298Z     info    FLAG: --domain="higress-system.svc.cluster.local"
2026-05-30T18:46:24.347302Z     info    FLAG: --help="false"
2026-05-30T18:46:24.347304Z     info    FLAG: --log_as_json="false"
2026-05-30T18:46:24.347306Z     info    FLAG: --log_caller=""
2026-05-30T18:46:24.347308Z     info    FLAG: --log_output_level=""
2026-05-30T18:46:24.347309Z     info    FLAG: --log_stacktrace_level="default:none"
2026-05-30T18:46:24.347314Z     info    FLAG: --log_target="[stdout]"
2026-05-30T18:46:24.347316Z     info    FLAG: --meshConfig="./etc/istio/config/mesh"
2026-05-30T18:46:24.347318Z     info    FLAG: --outlierLogPath=""
2026-05-30T18:46:24.347320Z     info    FLAG: --profiling="true"
2026-05-30T18:46:24.347322Z     info    FLAG: --proxyComponentLogLevel="misc:error"
2026-05-30T18:46:24.347324Z     info    FLAG: --proxyLogLevel="warning"
2026-05-30T18:46:24.347325Z     info    FLAG: --serviceCluster="higress-gateway"
2026-05-30T18:46:24.347327Z     info    FLAG: --stsPort="0"
2026-05-30T18:46:24.347329Z     info    FLAG: --templateFile=""
[...]

PR #984, opened 2026-07-03
fix(install): surface Higress gateway startup diagnostics

RerankerGuo:
## Summary
- print architecture, relevant Linux kernel limits, and the Higress Gateway log tail when the embedded installer times out waiting for `/status`
- distinguish a low `fs.inotify.max_user_instances` limit from the ARM64 Envoy/V8 crash that can persist after sysctls are raised
- explain that disabling Wasm is useful diagnostic evidence, but is not a supported workaround because AgentTeams relies on Higress Wasm for gateway authentication and AI routing
- document the same troubleshooting guidance in the English and Chinese FAQ

Refs #888.

## Verification
- `bash -n install/hiclaw-install.sh`
- `git diff --check`
- extracted-function simulation verifies ARM64, all three kernel values, the
  fatal log tail, and the Wasm compatibility warning
- confirmed the branch is based on current `main` (`6f0c7da`)
- PowerShell parser not run locally because `pwsh` is not installed in this environment

## CI note

Both refreshed runs passed every build and 9/10 integration matrix jobs. The
only failures were two different pre-existing `test-06-multi-worker` setup
races in copaw/hermes SHARD_A:

- [run 29233390641](https://github.com/agentscope-ai/AgentTeams/actions/runs/29233390641): Bob was fully provisioned, but the collaborative-task message denied a pending Tool Guard approval instead of starting a new Manager turn
- [run 29236623909](https://github.com/agentscope-ai/AgentTeams/actions/runs/29236623909): the test aborted before Bob setup because the Manager DM did not become quiet within 180 seconds

This PR only changes installer diagnostics and FAQ text. #985 now removes both
Manager-conversation dependencies from Bob's deterministic test setup.

## Temporary dependency

- This branch is temporarily stacked on #985 (`78e9385`), whose full CI passed all 10 integration matrices.
- Relative to #985, this PR retains only its original scoped changes; the shared test fix disappears from the diff after #985 merges.
- Fresh checks on `b0d4f81d` passed every build and all 10 integration matrices.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -34,6 +34,7 @@ Record release-facing changes here before the next release.
 - **Install script model env passthrough**: Pass custom model override env vars to the Controller container so that `AGENTTEAMS_MODEL_VISION` and related settings actually reach the config generator.
 - **Bridge model capability propagation**: Propagate model `input` modalities from openclaw.json through `_write_providers_json()` so QwenPaw's `ModelInfo` receives `supports_image`/`supports_video`/`supports_multimodal` flags instead of relying on fail-open defaults.
 - **Manager diagnostic loops**: Manager prompts and Worker lifecycle guidance stop repeated no-op troubleshooting commands and treat a missing Worker in `agt get workers` as the deletion boundary instead of looping on Matrix room probes. ([#975](https://github.com/agentscope-ai/AgentTeams/pull/975))
+- **Higress startup diagnostics**: Report architecture, kernel limits, and the current gateway log when the embedded Higress Gateway fails to become ready, including ARM64 Wasm/V8 guidance. ([#984](https://github.com/agentscope-ai/AgentTeams/pull/984))
 
 ---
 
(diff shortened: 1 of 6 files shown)


RerankerGuo:
@shiyiyue1102 Hi! Quick note on the SHARD_A copaw/copaw and copaw/hermes failures: I downloaded the test artifacts from run 28634278368. In the copaw/hermes case, the `hiclaw-worker-bob` session directory does not exist at all and Manager never invoked `hiclaw create worker --name bob` — Manager only created `alice`. In the copaw/copaw case, bob did get created and reached 14 LLM calls, but a later assertion still failed.

So the failure mode is the Manager not creating bob (a test-prompt interpretation flake), not anything in the install diagnostics this PR adds. The change itself is shell diagnostics + a sysctl note in the FAQ. Could you either re-run CI or merge if you agree the flake is unrelated? Happy to dig further if you can point me at what assertion actually failed.

RerankerGuo:
@shiyiyue1102 Hi! Re-rebased onto current `main` (HEAD `0b562ff`) and re-ran CI. SHARD_A is now passing on this PR — but a different integration test failed:

`integration-tests (llm-interaction-2, SHARD_B_TESTS, copaw, hermes)` — that's `tests/test-14-git-collab.sh` (the 4-phase non-linear multi-worker git collaboration workflow). Took 57 minutes before failing.

This is a **different failure pattern** from the earlier SHARD_A copaw/hermes flake. Test 14 has its own LLM-flake surface (3 workers, 4 phases, complex branch-name / file-path / report-name invariants), and there's no equivalent deterministic-bypass like #985 yet for it. The change in this PR is shell diagnostics for Higress Gateway startup + an English/Chinese FAQ note on sysctl checks — it doesn't touch Manager agent runtime, worker git ops, or test-14 setup. Could you re-run SHARD_B copaw/hermes, or merge this PR if you agree the test 14 failure is unrelated? Happy to dig into the test-14 artifacts (run 28926397804, job 85816940745) if it would help.

Also: PR #985 (SHARD_A bob-setup deterministic fix) just turned green and was pinged this morning as priority.

RerankerGuo:
Rebuilt on current `main` with the stacked #985 commit removed. The Bash and PowerShell installers now report architecture, kernel limits, and `/var/log/agentteams/higress-gateway.log` when the embedded gateway times out; FAQ guidance is synchronized.

Verification:
- `/bin/bash install/tests/test-higress-timeout-diagnostics.sh` — pass
- `/bin/bash -n install/agentteams-install.sh install/tests/test-higress-timeout-diagnostics.sh`

The historical commit used an obsolete noreply identity; the rebuilt commit was reset to `RerankerGuo <121015044+RerankerGuo@users.noreply.github.com>` before push.

RerankerGuo:
Closing because the installer, embedded topology, and diagnostics changed through the v1.2.x releases. This conflicting branch lacks a current-main reproduction; any remaining startup-diagnostics gap should return as a focused patch against the current installer.

PR #983, opened 2026-07-03
fix(worker): keep live OpenClaw token out of bulk file sync

RerankerGuo:
## Summary

Fixes #809 by preventing the manual OpenClaw Worker file-sync path from ever replacing the live `openclaw.json` with the stale MinIO copy, even briefly.

The previous flow copied the live file aside, bulk-mirrored the remote workspace over it, and merged the saved local copy afterward. OpenClaw watches this file, so the stale Matrix token could be observed during that overwrite window and used on reconnect.

## Changes

- fetch remote `openclaw.json` into a Worker-specific temporary file;
- exclude the live `openclaw.json` from the bulk `mc mirror`;
- merge Manager-owned remote fields into the untouched local config after the mirror;
- preserve the existing local-first merge contract, including the local Matrix `accessToken`;
- add a focused behavior test with a mock `mc` implementation.

The behavior test reads the live token *inside* the mocked bulk-mirror operation. It fails if `openclaw.json` was not excluded or if the stale remote token became visible. It also verifies that Manager-owned model/channel fields are updated, Worker-owned fields remain, other workspace files are pulled, and the temporary remote config is removed.

## Scope

This PR contains only:

- `manager/agent/worker-agent/skills/file-sync/scripts/agentteams-sync.sh`
- `manager/tests/test-agentteams-sync.sh`
- the required bilingual changelog entry

The previously stacked integration-test commits are no longer part of the branch.

## Verification

- [x] `bash -n manager/agent/worker-agent/skills/file-sync/scripts/agentteams-sync.sh manager/tests/test-agentteams-sync.sh`
- [x] `bash manager/tests/test-agentteams-sync.sh`
- [x] `git diff --check`

## Impact

- Breaking change: no
- Affected runtime: OpenClaw Worker manual/on-demand file sync
- Performance: one object `mc cp` replaces the previous local backup copy; bulk mirror scope is reduced by one file

PR #976, opened 2026-07-02
docs(team-leader): simplify worker name guidance

RerankerGuo:
## Summary
- replace defensive `assignedTo` warning blocks with a concise Matrix-localpart rule
- point DAG and Loop guidance at the same rule without duplicating examples
- simplify task-management Worker Name Canonicality while preserving `hiclaw get workers` / `matrixUserID` lookup guidance

Fixes #913.

## Tests
- git diff --check
- rg -n "❌|mandatory before|Common mistake|Do NOT use CLI|strip prefixes|matrixUserID|assignedTo|assigned_to" manager/agent/team-leader-agent/skills/project-management/SKILL.md manager/agent/team-leader-agent/skills/project-management/references/dag-execution.md manager/agent/team-leader-agent/skills/task-management/SKILL.md

## Temporary dependency

- This branch is temporarily stacked on #985 (`78e9385`), whose full CI passed all 10 integration matrices.
- Relative to #985, this PR retains only its original scoped changes; the shared test fix disappears from the diff after #985 merges.
- Fresh checks on `2d04f86d` passed every build and all 10 integration matrices.

PR #975, opened 2026-07-02
fix(manager): stop repeated diagnostic loops

RerankerGuo:
## Summary
- add Manager prompt safeguards against repeating the same no-op or malformed troubleshooting command
- define Worker deletion completion as the target disappearing from `hiclaw get workers`
- update worker lifecycle guidance to avoid looping on Matrix room probes such as `copaw channels list` after deletion

Fixes #974.

## Tests
- git diff --check
- rg -n "Diagnostic loop safeguard|Worker deletion is complete|Delete Completion Boundary|Stop repeated diagnostics" manager/agent/AGENTS.md manager/agent/copaw-manager-agent/AGENTS.md manager/agent/skills/worker-management/references/lifecycle.md manager/agent/skills/worker-management/SKILL.md

## Temporary dependency

- This branch is temporarily stacked on #985 (`78e9385`), whose full CI passed all 10 integration matrices.
- Relative to #985, this PR retains only its original scoped changes; the shared test fix disappears from the diff after #985 merges.
- Fresh checks on `90c9fd4f` passed every build and all 10 integration matrices.

PR #972, opened 2026-07-02
fix(controller): sync Human worker allowlist

RerankerGuo:
## Summary
- sync Human `spec.accessibleWorkers` into target Worker `channelPolicy.groupAllowExtra` so room access also grants runtime message permission
- mark controller-managed allowlist entries with a Worker annotation so revokes/deletes only remove entries created by Human reconciliation
- add HumanReconciler tests for add, revoke, and preserving manual allowlist entries

Fixes #954.

## Tests
- git diff --check
- rg -n "<<<<<<<|=======|>>>>>>>" changelog/current.md hiclaw-controller/internal/controller

Local: Go 1.25.0 `go test ./internal/controller` passes, including the three Human allowlist regression tests.

## Temporary dependency

- This branch is temporarily stacked on #985 (`78e9385`), whose full CI passed all 10 integration matrices.
- Relative to #985, this PR retains only its original scoped changes; the shared test fix disappears from the diff after #985 merges.
- Fresh checks on `7e36aa71` passed every build and all 10 integration matrices.

PR #985, opened 2026-07-03
test: make multi-worker bob setup deterministic

RerankerGuo:
## Summary
- make `test-06-multi-worker` create Bob via `hiclaw apply worker` instead of a Manager DM request
- remove the obsolete Manager-DM quiet gate because deterministic CLI setup does not depend on Manager conversation state
- use the full 180-second asynchronous provisioning window without reading the removed DM `REPLY`
- keep the rest of the test focused on multi-worker collaboration after Alice and Bob exist

## Why
`test-02-create-worker` already covers Manager-mediated worker creation. Repeating that LLM interaction in `test-06` made unrelated PR checks depend on Manager backlog and Tool Guard timing rather than multi-worker collaboration.

Two refreshed #984 runs exposed both races:

- run `29233390641`: Bob was provisioned, but a Tool Guard approval was pending when the collaborative-task message arrived, so that message denied the tool instead of becoming a new task
- run `29236623909`: all tests through `test-05` passed, but `test-06` aborted because the Manager DM did not become quiet within 180 seconds, before Bob setup even started

Direct controller setup removes both dependencies while preserving strict checks for Bob's persisted state, runtime, Higress consumer, and MinIO files. The test still exercises the actual Manager collaboration flow after setup.

## Verification
- `bash -n tests/test-06-multi-worker.sh`
- `git diff --check`
- rebased onto current `main` (`6f0c7da`)
- [refreshed CI run 29238893338](https://github.com/agentscope-ai/AgentTeams/actions/runs/29238893338): all builds and all 10 integration matrix jobs passed, including the previously failing copaw/hermes SHARD_A

I could not run the embedded integration test locally because this environment cannot access `/var/run/docker.sock` (`permission denied`).

### DIFF
--- a/tests/test-06-multi-worker.sh
+++ b/tests/test-06-multi-worker.sh
@@ -35,53 +35,27 @@ wait_for_manager_agent_ready 300 "${DM_ROOM}" "${ADMIN_TOKEN}" || {
     exit 1
 }
 
-# test-05 can leave CoPaw Manager finishing heartbeat / pending-worker cleanup
-# replies in the admin DM. Let that prior turn go quiet before measuring Bob's
-# create-worker ack/provisioning SLA; the post-request waits below stay strict.
-if ! matrix_wait_for_sender_quiet "${ADMIN_TOKEN}" "${DM_ROOM}" "@manager" 20 180; then
-    log_fail "Manager DM did not become quiet before Bob create request"
-    test_teardown "06-multi-worker"
-    test_summary
-    exit 1
-fi
-
 # Alice is running from previous tests; bob will be created below (offset=0 is correct for new workers)
 wait_for_worker_container "alice" 60
 METRICS_BASELINE=$(snapshot_baseline "alice" "bob")
 TEST_WORKER_RUNTIME="${AGENTTEAMS_DEFAULT_WORKER_RUNTIME:-openclaw}"
-# worker-management/SKILL.md tells Manager to ask admin for FOUR inputs
-# (name / runtime / SOUL / skills) before running `agt create worker`
-# and not to invent defaults. A vague prompt that only names the worker is
-# therefore a coin flip — sometimes Manager replies with a confirmation
-# request, never calls the CLI, and the consumer/SOUL.md polls below
-# silently time out. Spell out all four inputs and tell Manager to skip
-# confirmation so this test exercises actual Worker creation.
-#
-# The runtime is explicit because the CI matrix runtime is the source of truth;
-# rendered Manager workspace text may contain fallback defaults.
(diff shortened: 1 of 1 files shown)


RerankerGuo:
@shiyiyue1102 Hi! This PR makes the multi-worker `test-06-multi-worker.sh` bob setup deterministic by replacing the LLM-driven DM instruction with a direct `hiclaw apply worker --name bob ...` call. This bypasses the SHARD_A flake that has been blocking PRs #972, #973, #975, #976, #983, #984 (all hit "Manager never invoked hiclaw create worker --name bob" — confirmed by inspecting #984's downloaded artifacts where the `hiclaw-worker-bob` session directory did not exist at all).

Just rebased onto current `main` and CI is **18/18 green** (verified just now). If you can land this first, the next CI runs on the six Tier-3 PRs above should turn green automatically — they all hit the same root cause this PR fixes. Branch is MERGEABLE. Could you prioritize this one?

RerankerGuo:
CI update: run [#1647](https://github.com/agentscope-ai/AgentTeams/actions/runs/29327402001) failed before any integration test started. All affected jobs exited during embedded install because the base workflow reads the new `AGENTTEAMS_LLM_API_KEY` secret name while the repository still provides the legacy secret. I opened #1026 with a focused workflow fallback. The #985 test change was not reached in these failed jobs.

RerankerGuo:
Fresh independent evidence after the repository secret recovered: draft #1041's base-branch run [29489698196](https://github.com/agentscope-ai/AgentTeams/actions/runs/29489698196/job/87594978711) reached the full copaw/hermes SHARD_A suite and reproduced the setup race this PR removes.

In unpatched `test-06`, the Manager's create-Bob DM reply was empty, so both reply assertions failed. Bob was nevertheless provisioned, his Higress consumer and MinIO files existed, and the later collaborative task completed successfully. #985 replaces only that flaky LLM-driven setup/reply dependency with `hiclaw apply worker`, while retaining those downstream provisioning and collaboration checks.

The current red checks on #985 are still the older pre-test secret failure; this newer run does not include #985, but directly reproduces its target failure on the current base. No patch change is needed.

RerankerGuo:
Rebuilt as one independent commit on current `main`. The PR now changes only `tests/test-06-multi-worker.sh`: Bob is deterministic test setup through `agt create worker --no-wait`, while the remaining assertions still exercise multi-worker collaboration.

Verification:
- `/bin/bash -n tests/test-06-multi-worker.sh`
- `git diff --check`

The full integration behavior requires the project CI environment; `shellcheck` is not installed locally.

RerankerGuo:
Closing the remaining branch from the old stacked test-stabilization series. The current integration suite has changed, and this patch should not stay open without a fresh reproducible flake on current main.

PR #986, opened 2026-07-03
refactor(controller): decouple team worker CRs and rename API group

shiyiyue1102:
## Summary
- rename Kubernetes API resources from hiclaw.io/v1beta1 to agentteams.io/v1beta1
- add Team spec.workerMembers references so Teams coordinate standalone Worker CRs instead of owning inline member runtimes
- keep decoupled Team membership lookup, access resolution, status aggregation, and docs/tests aligned with the new API shape

## Compatibility
- Breaking change: Kubernetes API group and CRD names move from hiclaw.io to agentteams.io.
- Team worker membership now references Worker CRs through spec.workerMembers.

## Validation
- go test ./api/v1beta1 ./internal/auth ./internal/accessresolver ./internal/controller
- make check-crd-sync


shiyiyue1102:

maplefeng-a (changes requested):
Thanks for the update. I agree with the direction of moving Teams to reference standalone Worker CRs, but I found a few gaps in the new `spec.workerMembers` path that should be fixed before merge.

1. `workerMembers` identities are still enriched as standalone Workers.

`CREnricher.EnrichIdentity` first gets the Worker CR and returns immediately after reading `agentteams.io/team` / `agentteams.io/role` annotations. In the new decoupled model, the referenced Worker CR exists but TeamReconciler does not stamp those annotations, so a `workerMembers` team leader/worker is left with empty `Team` and the default worker role. That means STS/access resolution and team-leader authorization can take the standalone Worker path instead of the team member path.

Could you update the enricher to also reverse-lookup `spec.workerMembers.name` when a Worker CR exists without team annotations, and add tests for both a referenced leader and a referenced worker?

2. The workers API still only recognizes the old inline Team members.

`GetWorker` returns the Worker CR response immediately when the CR exists, and `findTeamMember` / `teamMemberToResponse` only look at `spec.leader` and `spec.workers`. For a decoupled Team, `GET /api/v1/workers/<member>` and list/filter responses will show the referenced Worker as standalone, with missing `team` / `role`, instead of the Team membership described by this PR.

Could you make the resource handler resolve `spec.workerMembers` as well, preferably reusing the field indexer rather than scanning only the legacy fields, and cover GET/list/team-filter behavior in tests?

3. A Worker can be referenced by multiple Teams.

`validateWorkerMembers` only rejects duplicates inside the same Team. Since `InjectWorkerCoordination` writes a single `TeamName` / `TeamLeaderName` into the Worker AGENTS.md, two Teams referencing the same Worker will overwrite each other's coordination context. The docs state that each Worker can belong to only one Team, so the controller should reject or degrade this case before writing coordination context.

---

这次方向我认可：Team 引用 standalone Worker CR 是合理的。但当前 `spec.workerMembers` 新路径还有几个合并前需要补齐的点。

1. `workerMembers` 身份仍会被识别成 standalone Worker。

`CREnricher.EnrichIdentity` 会先查 Worker CR，查到后只读 `agentteams.io/team` / `agentteams.io/role` annotation 就直接返回。新 decoupled 模型下，被引用的 Worker CR 本来就存在，但 TeamReconciler 不会给它写这些 annotation，所以 team leader / worker 的 `Team` 为空、role 仍是普通 worker。这样 STS/access resolver 和 team-leader 授权会走 standalone Worker 路径，而不是 team member 路径。

建议让 enricher 在 Worker CR 存在但没有 team annotation 时，也反查 `spec.workerMembers.name`，并补 leader/worker 两种引用场景的测试。

2. workers API 仍只识别旧的 inline Team 成员。

`GetWorker` 查到 Worker CR 后会直接返回 Worker response，`findTeamMember` / `teamMemberToResponse` 也只看 `spec.leader` 和 `spec.workers`。对 decoupled Team 来说，`GET /api/v1/workers/<member>` 以及 list/filter 会把被引用的 Worker 显示成 standalone，缺少 `team` / `role`，和 PR 描述的新 membership 语义不一致。

建议让 resource handler 也解析 `spec.workerMembers`，最好复用 field indexer，不只扫描 legacy 字段，并补 GET/list/team-filter 的测试。

3. 同一个 Worker 可以被多个 Team 引用。

`validateWorkerMembers` 只拒绝同一个 Team 内部重复引用。可是 `InjectWorkerCoordination` 会向 Worker AGENTS.md 写入单值 `TeamName` / `TeamLeaderName`，两个 Team 引用同一个 Worker 会互相覆盖协调上下文。文档也写了每个 Worker 只能属于一个 Team，所以 controller 应该在写协调上下文前拒绝或 Degraded 这个场景。

shiyiyue1102:
Thanks for catching these. Fixed by resolving spec.workerMembers membership in auth identity enrichment and Worker REST GET/list paths, and by rejecting a Worker CR referenced by more than one Team before coordination context is injected. Added regression coverage for decoupled leader/worker identity, Worker API GET/team filtering, and cross-Team duplicate references.

maplefeng-a (approved):
Thanks for the update. I rechecked the latest head and the three previously requested gaps are addressed: `spec.workerMembers` is now resolved during identity enrichment, the Worker REST API now returns the referenced Worker CRs with the correct team/role view, and Teams now reject sharing the same referenced Worker before coordination context is injected.

I also verified the focused controller packages and the full `hiclaw-controller` test suite locally. This looks ready to merge from my side.

PR #989, opened 2026-07-04
fix(copaw): harden Matrix channel routing

shiyiyue1102:
## Summary
- require Matrix group-room control commands to be explicitly mentioned and slash-prefixed
- route reasoning/tool progress into task threads while keeping final replies visible in the main room
- suppress NO_REPLY and cancellation noise in CoPaw Matrix output

## Validation
- `python3 -m py_compile copaw/src/matrix/channel.py copaw/tests/test_channel_mention.py`
- `git diff --check`

`uv run --python 3.11 --project copaw pytest tests/test_channel_mention.py` was attempted locally but could not build `python-olm==3.2.16` on this macOS environment because its native libolm build needs unavailable/incompatible local C++ build tooling (`cmake`/`gmake`, then clang const assignment failure). GitHub CI should run the normal Linux test environment.


shiyiyue1102:

maplefeng-a (approved):
CI is green and the CoPaw-scoped change looks good under the agreed review policy. Approving.

PR #992, opened 2026-07-04
feat(controller): add Matrix Human SSO identity

shiyiyue1102:
## Summary
- add Human identity-source plumbing for external SSO Matrix identities
- route Human provisioning/login/deletion through AppService-aware identity sources
- add AppService event handling tests, room metadata helpers, and CRD schema for identitySource/lastActiveAt

## Validation
- go test ./internal/matrix ./internal/server ./internal/controller ./internal/service ./internal/config ./internal/app
- make check-crd-sync
- git diff --check


shiyiyue1102:

maplefeng-a (changes requested):
Thanks for the update. I think the SSO/AppService direction is reasonable, but before merging we should close two integration gaps:

1. SSO Human as Team admin still goes through the legacy localpart/password path. `resolveTeamAdminActor` derives the Matrix user from `human.Spec.EffectiveUsername(...)` and passes that localpart into `LoginAsHuman`, while `external_sso` derives a hash localpart from issuer/subject and `HumanReconciler` clears `Status.InitialPassword`. That means a Team admin backed by an SSO Human either fails the explicit `matrixUserId` check or resolves/logs in as the wrong Matrix identity. Please drive this path from the Human's resolved/status identity source instead of the legacy username path, and add coverage for an SSO Human used as Team admin.

2. The new AppService transaction handler is not reachable in the running controller. `HandleTransactions` is implemented and tested directly, but `NewHTTPServer` only registers `/api/v1/appservice/rotate-token`, and `Config.MatrixConfig()` never populates `AppServicePushURL`, so the rendered registration omits `url`. The homeserver therefore has nowhere to push `/_matrix/app/v1/transactions/...`, so mention wake / `lastActiveAt` updates will not run outside unit tests. Please either wire the routes and registration URL end to end, or split/defer the event-handler part from this PR.

Once these are covered, the rest of the direction looks reasonable.

---

这次 SSO/AppService 的方向我觉得是合理的，但合并前需要先补齐两个集成缺口：

1. SSO Human 作为 Team admin 时仍然走 legacy localpart/password 路径。`resolveTeamAdminActor` 现在从 `human.Spec.EffectiveUsername(...)` 推导 Matrix 用户并传给 `LoginAsHuman`，但 `external_sso` 实际上是基于 issuer/subject 派生 hash localpart，`HumanReconciler` 也会清空 `Status.InitialPassword`。这样一来，SSO Human 做 Team admin 时，要么显式 `matrixUserId` 校验失败，要么解析/登录成另一个 legacy 风格的 Matrix 身份。建议这里改成基于 Human 已解析/已落 status 的 identity source 来驱动，并补一个 SSO Human 做 Team admin 的覆盖。

2. 新增的 AppService transaction handler 还没有接入真实 controller 运行路径。`HandleTransactions` 已实现且有直接单测，但 `NewHTTPServer` 目前只注册了 `/api/v1/appservice/rotate-token`，`Config.MatrixConfig()` 也没有填充 `AppServicePushURL`，所以渲染出的 registration 会省略 `url`。因此 homeserver 没有地方 push `/_matrix/app/v1/transactions/...`，mention wake / `lastActiveAt` 更新在单测外不会生效。建议把 route 和 registration URL 端到端接上，或者把 event-handler 部分从这个 PR 里拆出/延后。

shiyiyue1102:
Thanks for the detailed review. Fixed in f131add9: Team admin resolution now uses the Human identity source/status identity, so external_sso admins log in through the derived AppService localpart instead of the legacy username/password path; and the AppService transaction/user/room routes are mounted when AppService mode is enabled, with the registration URL derived from the controller URL and passed through initializer registration. Added focused coverage for SSO Human as Team admin, HTTP transaction route wiring, and MatrixConfig push URL propagation. Validation: go test ./internal/matrix ./internal/server ./internal/controller ./internal/service ./internal/config ./internal/app, make check-crd-sync, and git diff --check.

maplefeng-a:
Thanks for the update. The SSO Human-as-Team-admin path looks addressed now.

I compared the AppService wiring with the existing controller-side contract, and I think one integration gap remains: the AppService registration URL should stay as the homeserver-reachable controller base URL, while the controller mounts the Matrix AS endpoints under `/_matrix/app/v1/...`.

In this PR, `appServicePushURL(controllerURL)` returns `<controllerURL>/_matrix/app/v1`. A spec-compliant homeserver will call the Matrix AS transaction path under the registered base URL, so this can become `/_matrix/app/v1/_matrix/app/v1/transactions/...`, while the controller only mounts `/_matrix/app/v1/transactions/{txnId}`.

Could you adjust this to keep the registered URL as the controller base URL, and add a regression test that composes the rendered registration URL with the homeserver transaction path and reaches the mounted handler? It would also be good to preserve a separate Matrix-side controller URL override for deployments where the worker-facing controller URL and the homeserver-facing controller URL differ.

---

SSO Human 作为 Team admin 的路径现在看起来已经修好了。

我对照了一下现有 controller 侧的 AppService contract，这里还剩一个集成缺口：AppService registration 里的 `url` 应该保持为 homeserver 可访问到的 controller base URL，而 controller 自己把 Matrix AS endpoints 挂在 `/_matrix/app/v1/...` 下。

这个 PR 里 `appServicePushURL(controllerURL)` 返回的是 `<controllerURL>/_matrix/app/v1`。按标准 homeserver 的拼法，它会在注册的 base URL 下请求 Matrix AS transaction path，所以实际可能变成 `/_matrix/app/v1/_matrix/app/v1/transactions/...`；但 controller 现在只挂了 `/_matrix/app/v1/transactions/{txnId}`。

建议这里改成注册 controller base URL，并加一个回归测试：用渲染出的 registration URL 拼 homeserver transaction path，确认能打到已挂载 handler。也建议保留一个 Matrix 侧 controller URL override，用于 worker-facing controller URL 和 homeserver-facing controller URL 不同的部署形态。

shiyiyue1102:
Thanks for catching the AppService URL composition issue. Fixed in 39d64a5e: the rendered registration URL now stays at the homeserver-facing controller base URL, with AGENTTEAMS_MATRIX_APPSERVICE_CONTROLLER_URL as the Matrix-side override when it differs from the worker-facing controller URL. I also added regression coverage that composes the rendered registration URL with /_matrix/app/v1/transactions/... and verifies it reaches the mounted controller handler. Validation: go test ./internal/config ./internal/server ./internal/matrix, make check-crd-sync, git diff --check, and added-line naming scan.

shiyiyue1102:
@maplefeng-a The rerun is green now after the AppService URL fix. Could you take another look when convenient?

maplefeng-a (approved):
CI is green. The remaining AppService re-registration concern is limited to an upgrade/recovery path and not a fresh-install blocker, so I am approving this head.

PR #993, opened 2026-07-05
fix(controller): surface pod container failures

shiyiyue1102:
## Summary
- mark Kubernetes backend status as failed when worker/init containers report image pull, config, crash-loop, or non-zero termination failures
- include backend failure messages in worker runtime status API responses
- reconcile matching worker Pods when Ready or container waiting status changes without a Pod phase transition

## Scope
This is a focused controller hardening slice from the confirmed Sandbox follow-up plan. It does not include QwenPaw, TeamHarness, Edge/local remote worker, or credential/JWT changes.

## Validation
- go test ./internal/backend ./internal/controller ./internal/server
- git diff --check
- added-line naming scan for final AgentTeams naming

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,6 +4,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 ---
 
+- fix(controller): surface Kubernetes Pod container failures in Worker backend status and status API responses.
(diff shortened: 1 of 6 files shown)


maplefeng-a (approved):
CI is green and the controller failure-surfacing change looks good. Approving.

PR #998, opened 2026-07-05
feat(controller): harden sandbox worker deps

shiyiyue1102:
## Summary
- harden Sandbox-backed Worker reconciliation with runtime-affecting claim recycle and stale Sandbox cleanup
- move worker-deps env/token/data materialization behind controller-owned Sandbox setup, including bounded ServiceAccount token projection
- preserve OpenHuman and QwenPaw as parallel Worker runtimes in backend image routing and runtime config generation
- update Worker CRD/schema/RBAC plus focused backend/controller/service/agentconfig tests

## Validation
- go test ./internal/backend ./internal/controller ./internal/service ./internal/agentconfig
- make check-crd-sync
- git diff --cached --check

## Scope
This intentionally keeps the change in the Sandbox/WorkerDeps/Worker runtime contract area. It does not include TeamHarness plugin runtime or Edge/local remote worker follow-ups.


shiyiyue1102:

maplefeng-a (approved):
CI is green and the sandbox/WorkerDeps scoped changes look good after source review. Approving.

PR #1002, opened 2026-07-08
feat(controller): add AGENTTEAMS_MODEL_VISION env to override model-l…

LUOSENGWA:
## Problem

Two gaps in `config.go` where `agentconfig.Config` fields were declared but
never wired:

1. **ModelVision**: Custom multimodal models not in the built-in presets
   table (e.g. local `qwen3.6-27b-fp8`) fall through to `vision=false`.
   This is correct for models that genuinely lack vision
   (e.g. `deepseek-v4-pro`), but the preset table cannot know about
   user-defined local multimodal models.

2. **ModelReasoning**: Same pattern — `ModelReasoning *bool` declared in
   `types.go`, handled in `generator.go`, but never connected in `config.go`.

Both fields were always nil, so `resolveModelSpec()` never applied the
overrides.

## Fix

Add `AGENTTEAMS_MODEL_VISION` and `AGENTTEAMS_MODEL_REASONING` env var
support, following the same pattern as `ModelContextWindow` /
`ModelMaxTokens`:

- `Config.ModelVision *bool`, `Config.ModelReasoning *bool` — nil = use
  model default
- `envOptionalBool()` helper — returns nil when env unset (unlike
  `envBoolDefault` which returns a concrete bool)
- Pass through via `AgentConfig()` → `resolveModelSpec()`

## Behavior

PR-A does not change any model's default behavior. It adds override knobs
for deployers to set at Controller startup.

**ModelVision** (`AGENTTEAMS_MODEL_VISION`):

| Scenario | Before | After |
|---|---|---|
| `qwen3.6-27b-fp8` (local multimodal), env not set | `vision=false` ❌ | `vision=false` (unchanged — knob provided but not yet turned) |
| `qwen3.6-27b-fp8`, `AGENTTEAMS_MODEL_VISION=true` | `vision=false` ❌ | `vision=true` ✅ |
| `qwen3.6-plus` (in presets, vision:true), env not set | `vision=true` ✅ | unchanged |
| `deepseek-v4-pro` (no vision support), env not set | `vision=false` ✅ | unchanged (correct; should NOT set this env) |

**ModelReasoning** (`AGENTTEAMS_MODEL_REASONING`):

| Scenario | Before | After |
|---|---|---|
| Any model, env not set | preset default | unchanged |
| Any model, `AGENTTEAMS_MODEL_REASONING=true` | preset default | `reasoning=true` |
| Any model, `AGENTTEAMS_MODEL_REASONING=false` | N/A | `reasoning=false` |

Backward compatible: nil default means "use model default".

## Usage

Deployers set the env vars on the Controller:

    AGENTTEAMS_MODEL_VISION=true
    AGENTTEAMS_MODEL_REASONING=true

Then restart the Controller. All generated model specs will include
`"image"` in their `Input` array (vision) and the specified reasoning
capability.

Note: `AGENTTEAMS_MODEL_VISION` is a global override (same pattern as
`ModelContextWindow` / `ModelMaxTokens`). It affects all models. For
text-only models like DeepSeek this is harmless — their API ignores the
capability declaration. `AGENTTEAMS_MODEL_REASONING` overrides the
reasoning field for all models where the provider supports it.

## Files changed

- `hiclaw-controller/internal/config/config.go`: +17 lines
- `changelog/current.md`: +3 lines

2 commits:
- `71f5ef7` feat(controller): add AGENTTEAMS_MODEL_VISION env
- `28ca8bc` feat(controller): wire ModelReasoning env through config

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -5,6 +5,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 ---
 
 - feat(controller): add `AGENTTEAMS_MODEL_VISION` env to override model-level vision capability for custom models not in the built-in presets table (e.g. local multimodal models like `qwen3.6-27b-fp8`).
+- feat(controller): add `AGENTTEAMS_MODEL_REASONING` env to override model-level reasoning capability for custom models not in the built-in presets table.
 
 - feat(qwenpaw): add the QwenPaw worker runtime Python package baseline with runtime config sync, storage sync, heartbeat reporting, Matrix channel overlay, and focused unit tests.
 - fix(controller): surface Kubernetes Pod container failures in Worker backend status and status API responses.
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,6 +4,8 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 ---
 
+- feat(controller): add `AGENTTEAMS_MODEL_VISION` env to override model-level vision capability for custom models not in the built-in presets table (e.g. local multimodal models like `qwen3.6-27b-fp8`).
+
 - feat(qwenpaw): add the QwenPaw worker runtime Python package baseline with runtime config sync, storage sync, heartbeat reporting, Matrix channel overlay, and focused unit tests.
 - fix(controller): surface Kubernetes Pod container failures in Worker backend status and status API responses.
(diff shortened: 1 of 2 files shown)


LUOSENGWA:
The integration-tests (llm-interaction, SHARD_A, copaw) failure is a pre-existing CI flake — the copaw Manager does not respond to identity-config DMs in _setup_manager_identity() (see tests/run-all-tests.sh:232). This is unrelated to the config.go changes in this PR. The test-controller suite passes cleanly.

LUOSENGWA:
Close with merged #1103

PR #1003, opened 2026-07-08
feat(controller): allow overriding APIG endpoint

shiyiyue1102:
## Summary
- add an optional Helm value for the Alibaba Cloud APIG OpenAPI endpoint
- pass the value to the controller as HICLAW_APIG_ENDPOINT for ai-gateway installs
- let the APIG SDK client use the override, falling back to apig.<region>.aliyuncs.com

## Validation
- git diff --check
- go test ./internal/config ./internal/gateway (from hiclaw-controller)

Split from the refreshed internal/open-source diff based on codex/sync-opensource-main-20260701 @ 0bf155a2.


shiyiyue1102:

maplefeng-a (approved):
CI is green and the APIG endpoint override change looks good after source review. Approving.

PR #1004, opened 2026-07-08
fix(controller): skip unsupported gateway port exposure

shiyiyue1102:
## Summary
- treat gateway.ErrUnsupportedOp from port exposure reconciliation as a supported no-op
- preserve the current exposed port status when the selected gateway provider does not manage ports
- update the reconcile comment from Higress-specific wording to gateway-provider wording

## Validation
- git diff --check
- go test ./internal/controller (from hiclaw-controller)

Split from the refreshed internal/open-source diff based on codex/sync-opensource-main-20260701 @ 0bf155a2.

### DIFF
--- a/hiclaw-controller/internal/controller/member_reconcile.go
+++ b/hiclaw-controller/internal/controller/member_reconcile.go
@@ -1861,7 +1861,7 @@ func extraHostsForBackend(wb backend.WorkerBackend) []string {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


shiyiyue1102:
@maplefeng-a The checks are passing now; could you help review this when you have time?

maplefeng-a (approved):
CI is green and the unsupported gateway port exposure handling looks good after source review. Approving.

Issue #899, opened 2026-06-02
大模型超时无法回复 || Large model times out and cannot reply


berserker3912:
### Bug Description

在沟通界面向大模型提问，大模型长时间无法回复，最后报错超时

### Steps to Reproduce

在沟通界面向大模型提问，大模型长时间无法回复，最后报错超时

### Relevant Logs

```shell
ERROR /opt/copaw-venv/lib/python3.11/site-packages/agentscope_runtime/engine/runner.py:342 | 2026-06-02 05:27:53 | {'code': 'MODEL_TIMEOUT', 'message': 'Model inference timed out: qwen, timeout limit: 60 seconds\n(Details:  /tmp/copaw_query_error_0edg9ra8.json)'}: Traceback (most recent call last):
  File "/opt/copaw-venv/lib/python3.11/site-packages/copaw/app/runner/runner.py", line 535, in query_handler
    async for msg, last in stream_printing_messages(
  File "/opt/copaw-venv/lib/python3.11/site-packages/agentscope/pipeline/_functional.py", line 192, in stream_printing_messages
    raise exception from None
  File "/opt/copaw-venv/lib/python3.11/site-packages/agentscope/agent/_agent_base.py", line 455, in __call__
    reply_msg = await self.reply(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/copaw-venv/lib/python3.11/site-packages/agentscope/agent/_agent_meta.py", line 120, in async_wrapper
    current_output = await original_func(
                     ^^^^^^^^^^^^^^^^^^^^
  File "/opt/copaw-venv/lib/python3.11/site-packages/copaw/agents/react_agent.py", line 1026, in reply
    return await super().reply(
           ^^^^^^^^^^^^^^^^^^^^
  File "/opt/copaw-venv/lib/python3.11/site-packages/agentscope/agent/_agent_meta.py", line 120, in async_wrapper
    current_output = await original_func(
                     ^^^^^^^^^^^^^^^^^^^^
  File "/opt/copaw-venv/lib/python3.11/site-packages/agentscope/tracing/_trace.py", line 390, in wrapper
    return await func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/copaw-venv/lib/python3.11/site-packages/agentscope/agent/_react_agent.py", line 437, in reply
[...]

PR #1011, opened 2026-07-09
docs: add local model timeout diagnostics

RerankerGuo:
## Summary

- Add FAQ diagnostics for OpenAI-compatible local models that pass install preflight but time out during real chat.
- Include container-side `curl` verification for Xinference-style services.
- Call out `MODEL_TIMEOUT` with a 60 second limit as a signal to verify current official QwenPaw/CoPaw images.
- Add the same guidance to the Chinese FAQ.

## Why

Helps users debug #899, where install-time connectivity passed but runtime model calls from the chat interface timed out against a local Xinference OpenAI-compatible endpoint.

## Validation

- `git diff --check`

Docs-only change; no runtime tests were needed.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -439,6 +439,30 @@ container itself, not your Mac or host machine. Use a reachable host address,
 for example `http://host.docker.internal:<port>/v1` on Docker Desktop, or the
 host LAN IP on Linux/Podman when `host.docker.internal` is not available.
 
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing after the documentation tree was restructured in #1164. This conflicting patch targets the previous FAQ layout; any remaining timeout guidance gap should be added against the new documentation structure.

PR #1017, opened 2026-07-10
chore(controller): keep OSS worker deployment local and edge

shiyiyue1102:
## Summary
- keep the open-source Worker deployment surface to Local managed pods and the Edge worker flow
- reject Remote Worker deployment and remove the Worker `targetCluster` API/CRD/status surface entirely
- stop registering the OpenKruise sandbox backend from the OSS controller, remove Sandbox/OpenKruise Helm defaults/env/RBAC, and remove sandbox-only Worker CRD fields such as `volumes` and `mounts`
- update user docs so OSS Worker deployment no longer advertises Remote or Sandbox paths

## Scope
This is an intentional breaking CRD/API cleanup for the current main branch before a release. The OSS controller now exposes Local and Edge Worker deployment only. Historical low-level sandbox implementation code is left internal and unregistered in this PR, but it is no longer exposed through Helm, CRDs, or the OSS controller backend registry.

## Validation
- `git diff --check`
- Ruby YAML parse for `helm/hiclaw/values.yaml`, `helm/hiclaw/crds/workers.agentteams.io.yaml`, `hiclaw-controller/config/crd/workers.agentteams.io.yaml`
- `go test ./internal/app ./internal/controller ./api/v1beta1` from `hiclaw-controller/`
- `go test ./internal/auth ./internal/remoteclient` from `hiclaw-controller/`
- `go test ./internal/backend` from `hiclaw-controller/`

Note: `helm template` was not run locally because `helm` is not installed.


shiyiyue1102:

maplefeng-a (approved):
CI is green and the change looks good after source review. Approving.

PR #1018, opened 2026-07-12
fix(controller): use AgentTeams Matrix AppService env

shiyiyue1102:
## Summary
- switch Matrix AppService runtime secret keys to AGENTTEAMS_MATRIX_APPSERVICE_* in Helm and the local installer
- make controller config read the AgentTeams-prefixed AppService env vars
- update the rotate-token guidance to point at the new env names

## Scope
This is a narrow runtime-env slice of the HiClaw to AgentTeams rename. It does not change CRD/API groups, CLI command names, image names, container names, storage prefixes, Matrix room aliases, Worker targetCluster, Remote/Sandbox worker behavior, TeamHarness, QwenPaw runtime internals, SchedulerX, or Alibaba credential/provider behavior.

## Validation
- git diff --cached --check
- bash -n install/hiclaw-install.sh
- from hiclaw-controller: go test ./internal/config ./cmd/hiclaw


shiyiyue1102:

maplefeng-a (approved):
CI is green and the AppService env rename looks aligned after source review. Approving.

PR #1019, opened 2026-07-12
fix(controller): remove OSS target cluster routing

shiyiyue1102:
## Summary
- remove the OSS-facing cluster auth header from the hiclaw CLI API client and multipart uploads
- stop reading cluster headers in controller auth middleware and keep TokenReview on the local Kubernetes API path
- remove controller target-cluster routing from K8s/Sandbox backend paths, ServiceBackend lookup, and auth identity enrichment
- stop injecting or forwarding AGENTTEAMS_CLUSTER_ID from worker pods, TeamHarness, and find-skill STS requests; tests now assert the header/env is absent

## Scope
This is a narrow follow-up to the OSS Local/Edge Worker deployment surface: Remote K8s/targetCluster is no longer exposed in OSS, so the remote cluster auth/header/env and controller routing contract is removed instead of renamed. It does not change CRDs, API groups, CLI command names, image names, Matrix aliases, storage prefixes, SchedulerX, or Alibaba credential/provider internals.

## Validation
- git diff --check
- bash -n manager/tests/test-hiclaw-find-skill.sh
- bash manager/tests/test-hiclaw-find-skill.sh
- from hiclaw-controller/: go test ./cmd/hiclaw ./internal/auth ./internal/backend ./internal/controller ./internal/app ./api/v1beta1
- from hiclaw-controller/: go test ./...

Not run: python3 -m pytest plugins/tests/teamharness/adapters/qwenpaw/test_adapter.py (current local Python does not have pytest installed).


shiyiyue1102:

maplefeng-a (commented):
Thanks for the update. I found one remaining caller that still uses the legacy cluster header: `plugins/teamharness/skills/agent/find-skills/scripts/hiclaw-find-skill.sh`.

The k8s backend now injects `AGENTTEAMS_CLUSTER_ID`, and after this PR the controller auth middleware reads `X-AgentTeams-Cluster-ID`. If this TeamHarness script keeps checking only `HICLAW_CLUSTER_ID` and sending `X-HiClaw-Cluster-ID`, remote worker STS requests can lose the cluster ID and fall back to local TokenReview.

Could you update this TeamHarness find-skill script the same way as the manager/copaw/hermes copies, and add/extend coverage for that path?

---

这边还有一个调用方没有同步：`plugins/teamharness/skills/agent/find-skills/scripts/hiclaw-find-skill.sh` 仍然只读取 `HICLAW_CLUSTER_ID` 并发送旧的 `X-HiClaw-Cluster-ID`。

现在 k8s backend 注入的是 `AGENTTEAMS_CLUSTER_ID`，而这个 PR 后 controller auth middleware 读取的是 `X-AgentTeams-Cluster-ID`。如果 TeamHarness 这份脚本不一起更新，远端 worker 的 STS 请求会丢 cluster ID，进而回落到本地 TokenReview。

麻烦把这份 TeamHarness find-skill 脚本也按 manager/copaw/hermes 那几份同步一下，并补上对应覆盖。

shiyiyue1102:
Thanks for catching this. Fixed by updating the TeamHarness find-skill script to resolve AGENTTEAMS_CLUSTER_ID and send X-AgentTeams-Cluster-ID, and by adding that script to the existing find-skill regression test matrix. Validation: git diff --check; bash -n manager/tests/test-hiclaw-find-skill.sh; bash manager/tests/test-hiclaw-find-skill.sh.

shiyiyue1102:
Correction to my previous reply: after re-checking the OSS Worker scope, Remote K8s/targetCluster is no longer exposed here, so this PR now removes the OSS-facing remote cluster header/env contract instead of renaming it.

Updated in 47c6e8ce:
- the hiclaw CLI and find-skill scripts no longer send cluster headers;
- controller auth middleware no longer reads the cluster header;
- the k8s backend and TeamHarness no longer inject or forward AGENTTEAMS_CLUSTER_ID;
- tests now assert that both the AgentTeams and legacy cluster headers are absent.

Validation: git diff --check; bash -n manager/tests/test-hiclaw-find-skill.sh; bash manager/tests/test-hiclaw-find-skill.sh; go test ./cmd/hiclaw ./internal/auth ./internal/backend.

Not run locally: python3 -m pytest plugins/tests/teamharness/adapters/qwenpaw/test_adapter.py because this Python environment does not have pytest installed.

shiyiyue1102:
Updated again in 6b5426fa to remove the remaining controller-side target-cluster routing surface:

- TokenReview authentication is local-only now; the authenticator no longer accepts or caches by cluster ID.
- K8s/Sandbox backend paths no longer keep WithRemoteTarget, TargetClusterID, remote namespace creation, or remote ServiceClient routing.
- Service management now resolves the same-cluster client directly.
- The remote identity enrichment checks and remote-routing tests were removed or replaced with local-only coverage.

Validation added after this change: git diff --check; bash -n manager/tests/test-hiclaw-find-skill.sh; bash manager/tests/test-hiclaw-find-skill.sh; from hiclaw-controller/: go test ./cmd/hiclaw ./internal/auth ./internal/backend ./internal/controller ./internal/app ./api/v1beta1; go test ./....

maplefeng-a (approved):

PR #1020, opened 2026-07-12
fix(controller): skip manager reconciler when disabled

shiyiyue1102:
## Summary
- skip ManagerReconciler registration when Manager provisioning is disabled
- keep the existing CR count collector manager-skip behavior unchanged
- avoid requiring Manager CR dependencies in Worker-only OSS deployments

## Scope
This is a narrow controller startup fix for AGENTTEAMS_MANAGER_ENABLED/HICLAW_MANAGER_ENABLED=false deployments. It does not change CRDs, API groups, Helm defaults, worker deployment modes, Matrix AppService envs, image names, TeamHarness, QwenPaw, Remote/Sandbox workers, SchedulerX, or Alibaba credential/provider internals.

## Validation
- git diff --check
- go test ./internal/app ./internal/metrics

### DIFF
--- a/hiclaw-controller/internal/app/app.go
+++ b/hiclaw-controller/internal/app/app.go
@@ -647,30 +647,34 @@ func (a *App) initReconcilers(_ context.Context) error {
 		return fmt.Errorf("setup HumanReconciler: %w", err)
 	}
 
-	mgrReconciler := &controller.ManagerReconciler{
-		Client:           a.mgr.GetClient(),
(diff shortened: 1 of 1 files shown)


shiyiyue1102:
@maplefeng-a The checks are passing now; could you help review this when you have time?

maplefeng-a (approved):
CI is green and the ManagerReconciler gating change looks good after source review. Approving.

PR #1026, opened 2026-07-14
fix(ci): fall back to legacy LLM API key secret

RerankerGuo:
## Summary

- Keep the runtime-facing environment variable as `AGENTTEAMS_LLM_API_KEY`.
- Resolve its value from the new Actions secret first, then fall back to `HICLAW_LLM_API_KEY` while repository secrets are migrated.
- Apply the same fallback to PR integration tests and release-baseline install/test jobs.

## Root cause

The AgentTeams contract rename changed the workflow to read only `secrets.AGENTTEAMS_LLM_API_KEY`, but the configured repository secret still uses the legacy name. Run [#1647](https://github.com/agentscope-ai/AgentTeams/actions/runs/29327402001) therefore failed eight integration jobs before any test started with `LLM API Key is required`.

## Verification

- Parsed `.github/workflows/test-integration.yml` with PyYAML.
- Confirmed all three workflow consumers use the fallback expression.
- `git diff --check`

This unblocks current PR CI, including #985, without exposing or renaming any secret.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -357,7 +357,7 @@ jobs:
       - name: Run embedded integration tests
         id: tests
         env:
-          AGENTTEAMS_LLM_API_KEY: ${{ secrets.AGENTTEAMS_LLM_API_KEY }}
+          AGENTTEAMS_LLM_API_KEY: ${{ secrets.AGENTTEAMS_LLM_API_KEY || secrets.HICLAW_LLM_API_KEY }}
           AGENTTEAMS_LLM_PROVIDER: qwen
           AGENTTEAMS_DEFAULT_MODEL: ${{ inputs.model || 'qwen3.6-plus' }}
           AGENTTEAMS_MANAGER_RUNTIME: ${{ matrix.manager_runtime }}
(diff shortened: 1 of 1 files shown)


RerankerGuo:
Self-CI bootstrap note: run [#1655](https://github.com/agentscope-ai/AgentTeams/actions/runs/29329006577) uses the `pull_request_target` workflow definition from current `main`, so the fallback expression in this branch is not active for its own run. All 9 pre-integration checks passed. I inspected the raw logs for all 10 failed matrix jobs; every job stopped before tests with the identical `LLM API Key is required` installer error. Once this workflow patch is merged, later PR runs will use the fallback.

RerankerGuo:
Rebuilt the branch on current `main`. The diff remains limited to the three integration workflow secret bindings, preferring `AGENTTEAMS_LLM_API_KEY` and falling back to the legacy `HICLAW_LLM_API_KEY` secret.

Verification:
- reviewed the resulting workflow diff; `actionlint` is not installed locally, so GitHub Actions will provide the workflow parser check.

RerankerGuo:
Closing because this legacy fork-secret fallback no longer matches the current integration workflow and required-check setup. Keeping an obsolete CI compatibility branch open would add maintenance cost without current evidence.

PR #1029, opened 2026-07-15
fix(make): target AgentTeams containers in dev utils

RerankerGuo:
## Summary

- make `make status` list the renamed `agentteams-*` containers
- make `make clean` remove `agentteams-test-worker-*` containers created by the current test harness

## Root cause

The AgentTeams contract rename updated the container names and most Makefile targets, but these two Docker name filters still used the old `hiclaw-*` prefixes. As a result, `make status` omitted current containers and `make clean` left current test workers behind.

## Validation

- ran `make status` against a recording Docker stub and verified the `name=agentteams-` filter plus controller/manager output
- ran `make clean TEST_CONTAINER=agentteams-manager-test` against the stub and verified both current test workers were removed
- ran `make -n status` and `make -n clean TEST_CONTAINER=agentteams-manager-test`
- ran `git diff --check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -799,7 +799,7 @@ verify: ## Run post-install verification against the running Manager container
 
 status: ## Show status of Manager and all Worker containers
 	@echo "==> AgentTeams container status:"
-	@docker ps -a --filter "name=hiclaw-" --format "table {{.Names}}\t{{.Status}}\t{{.Image}}" 2>/dev/null \
+	@docker ps -a --filter "name=agentteams-" --format "table {{.Names}}\t{{.Status}}\t{{.Image}}" 2>/dev/null \
 		|| echo "  (no containers found or Docker not available)"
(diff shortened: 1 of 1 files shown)


RerankerGuo:
CI note: all 9 build/translation checks passed on run [#1657](https://github.com/agentscope-ai/AgentTeams/actions/runs/29384404631). I inspected the raw logs for all 10 failed integration matrix jobs; every job stopped before tests during embedded install with the identical `LLM API Key is required` error. This is the base `pull_request_target` secret-name regression addressed by #1026, so none of the failures exercised this PR's `make status` / `make clean` changes. The focused local Docker-stub validation remains documented in the PR body.

RerankerGuo:
Closing as superseded by current main. Makefile status and cleanup targets already select agentteams-prefixed containers.

PR #1009, opened 2026-07-09
feat(teamharness): add QwenPaw workerflow integration

shiyiyue1102:
## Summary
- Add TeamHarness QwenPaw adapter support for runtime team context, MCP client wiring, task trace packaging, and role skills.
- Expand TeamHarness MCP with message, roomflow, artifact, projectflow, and taskflow behavior needed for workerflow handoff.
- Add focused adapter/package/MCP contract tests and a QwenPaw TeamHarness e2e script.

## Scope
This PR excludes SchedulerX skills, credential/JWT bootstrap helpers, edge/local worker runtime, and broad naming/Helm/controller changes.

## Validation
- git diff --check HEAD
- bash -n tests/test-21-team-project-dag.sh
- bash -n tests/test-26-qwenpaw-teamharness-plugin-mode.sh
- ruby plugins/tests/teamharness/test-contracts.rb
- ruby plugins/tests/teamharness/mcp/test-server.rb
- uv run --with pytest --with pyyaml pytest plugins/tests/teamharness/adapters/qwenpaw/test_adapter.py plugins/tests/teamharness/adapters/qwenpaw/test_package.py plugins/tests/teamharness/test_trace.py plugins/tests/teamharness/test_trace_integration.py
- ruby plugins/tests/teamharness/mcp/tools/test-message.rb
- ruby plugins/tests/teamharness/mcp/tools/test-projectflow.rb
- ruby plugins/tests/teamharness/mcp/tools/test-taskflow.rb
- ruby plugins/tests/teamharness/mcp/tools/test-filesync.rb
- ruby plugins/tests/teamharness/mcp/tools/test-artifact.rb
- ruby plugins/tests/teamharness/mcp/tools/test-protocol.rb
- ruby plugins/tests/teamharness/mcp/tools/test-roomflow.rb

PR #1035, opened 2026-07-15
fix(teamharness): align QwenPaw and AgentSpec contracts

RerankerGuo:
## Summary

Two contract surfaces drifted after recent changes:

- #1009 intentionally made TeamHarness QwenPaw-only, but the plugin CLI test still required a Claude command. Assert the exact QwenPaw path and command instead.
- the AgentTeams hard rename updated the Ruby boundary assertion, but the boundary document still described the HiClaw CLI and AgentSpec package. Align those terms with the current product contract.

## Verification

Ran the canonical `plugins/tests/run-integration-tests.sh` entrypoint with Ruby 2.7 and loopback proxy bypassed:

- plugin CLI: 5/5 passed
- TeamHarness manifest and boundary contracts: passed
- QwenPaw adapter: 10/10 passed
- QwenPaw package build: passed
- TeamHarness server, message, filesync, projectflow, and taskflow contract tests: passed
- `git diff --check`

### DIFF
--- a/docs/teamharness-boundary-and-contracts.md
+++ b/docs/teamharness-boundary-and-contracts.md
@@ -18,7 +18,7 @@ TeamHarness owns:
   projects, task delegation, and task execution.
 - Explicit MCP tools for team messages, shared files, project flow, task flow,
   and plugin health.
-- A single plugin tarball installed through the HiClaw `agentteams` CLI by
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing as superseded by current main. The TeamHarness docs and QwenPaw plugin probe assertions already use the AgentTeams contract and the qwenpaw-only detection definition.

PR #1034, opened 2026-07-15
fix(test): update import mock Manager name

RerankerGuo:
## Summary

The hard rename updated `hiclaw-import.sh` to discover and execute `agentteams-manager`, but its unit-style Docker mock still returned and stripped `hiclaw-manager`. As a result, the test no longer reached the mocked `hiclaw` command and failed 5 of 6 assertions. Align the mock with the production container name.

## Verification

- `bash tests/test-16-import-worker-package-default.sh` (6/6 passed; previously 1/6)
- `bash -n tests/test-16-import-worker-package-default.sh install/hiclaw-import.sh`
- `git diff --check`

PR #1033, opened 2026-07-15
fix(test): align QwenPaw test with AgentTeams contracts

RerankerGuo:
## Summary

Test 26 still targeted several retired HiClaw contracts after the hard rename: the QwenPaw provider ID, embedded controller container and data directory, Kubernetes API path, and Worker/Team `apiVersion` values. Align the integration test with the canonical AgentTeams runtime and API contracts so it can create resources and validate the generated workspace once CI reaches the shard.

This changes only the integration test; production behavior already uses these AgentTeams values.

## Verification

- extracted workspace projection check against an `agentteams-gateway` fixture: `ok`
- exact static contract check for controller, token path, API group, resource declarations, and provider ID
- `bash -n tests/test-26-qwenpaw-teamharness-plugin-mode.sh`
- `go test ./api/v1beta1 ./internal/config ./internal/service` with Go 1.25.0
- `git diff --check`

The previous integration run stopped during `install-embedded` with the base-workflow LLM secret issue tracked in #1026, before any test shard ran.

PR #1030, opened 2026-07-15
fix(test): update MinIO mock for AgentTeams alias

RerankerGuo:
## Summary

- update the MinIO unit-test mock to recognize the `agentteams/` alias
- align the mock implementation with the renamed paths already used by the test cases

## Root cause

The AgentTeams hard rename updated the fake MinIO paths and assertions, but the mock resolver still classified only `hiclaw/*` as remote paths. As a result, `mc cp` treated every `agentteams/*` object path as a local filesystem path and all MinIO merge scenarios failed before exercising the intended behavior.

## Impact

This restores the unit test without changing production behavior. The test now exercises create, update, preservation, repair, idempotency, and multi-worker cases against the current storage alias.

## Validation

- `bash manager/tests/test-update-builtin-section-minio.sh` (20 passed)
- `bash manager/tests/test-update-builtin-section.sh` (40 passed)
- `bash -n manager/tests/test-update-builtin-section-minio.sh manager/scripts/lib/builtin-merge.sh`
- `git diff --check`

PR #1028, opened 2026-07-14
test: honor AgentTeams worker runtime env

RerankerGuo:
## Summary

- Make tests 02, 06, and 15 prefer `AGENTTEAMS_DEFAULT_WORKER_RUNTIME`.
- Preserve the legacy variable as a fallback for direct local test invocations.
- Keep `openclaw` as the final default.

## Root cause

The integration matrix now exports `AGENTTEAMS_DEFAULT_WORKER_RUNTIME`, but these three worker-creating tests still read only `HICLAW_DEFAULT_WORKER_RUNTIME`. Copaw and Hermes matrix jobs therefore explicitly created OpenClaw workers, so their reported runtime coverage did not match the configured matrix dimension.

## Verification

- `bash -n` passes for all three scripts.
- Shell assertions cover new-variable precedence, legacy fallback, and the default value in every file.
- `git diff --check`

This is test-only and does not change runtime behavior.

PR #1036, opened 2026-07-15
fix(test): target Manager agent for MCP checks

RerankerGuo:
## Summary

In the embedded architecture, `TEST_CONTROLLER_CONTAINER` names the infrastructure/controller container while `TEST_AGENT_CONTAINER` names the Manager Agent container. Test 12 still used the controller variable for all `mcporter` commands, so the GitHub MCP checks ran in a container without the Manager workspace or MCP configuration.

Prefer `TEST_AGENT_CONTAINER`, retain the controller fallback for legacy all-in-one deployments, and use the current Manager name as the final fallback.

## Verification

- recording Docker stub before the fix: all 18 `mcporter` calls targeted `agentteams-controller`
- recording Docker stub after the fix: all 18 calls target `agentteams-manager`
- `bash -n tests/test-12-github-mcp-tools.sh`
- `git diff --check`

### DIFF
--- a/tests/test-12-github-mcp-tools.sh
+++ b/tests/test-12-github-mcp-tools.sh
@@ -22,8 +22,8 @@ fi
 TEST_OWNER="${TEST_GITHUB_OWNER:-higress-group}"
 TEST_REPO="${TEST_GITHUB_REPO:-hiclaw}"
 
-# Manager container name
-MANAGER_CONTAINER="${TEST_CONTROLLER_CONTAINER:-hiclaw-manager}"
+# Manager Agent container name
+MANAGER_CONTAINER="${TEST_AGENT_CONTAINER:-${TEST_CONTROLLER_CONTAINER:-agentteams-manager}}"
 
 # Helper function to call mcporter inside Manager container
 mcporter_call() {


RerankerGuo:
CI attribution: all 9 translation/base/image checks passed for `ba16f9f`. I inspected the raw logs for all 10 failed matrix jobs in run [#1663](https://github.com/agentscope-ai/AgentTeams/actions/runs/29392334055); each log has an empty `AGENTTEAMS_LLM_API_KEY`, exits from `install-embedded` with code 2 and `LLM API Key is required`, and reports no test output. The workflow therefore never reached test 12. This is the same `pull_request_target` base-workflow secret blocker addressed by #1026, so the locally verified Manager-container fix is unchanged.

shiyiyue1102:
This two-line test correction is reasonable, but it has little standalone value. There is a larger group of tiny rename/test-contract PRs here (#1033, #1034, #1035, #1030, #1028, and nearby dev-helper fixes). Please consolidate the related corrections into one focused test-contract cleanup and list the exact failing tests each change fixes. That will let us validate and merge the rename fallout as one coherent unit instead of reviewing many two-to-eight-line PRs separately.

RerankerGuo:
Closing as superseded by current main. tests/test-12-github-mcp-tools.sh already selects TEST_AGENT_CONTAINER with the agentteams-manager fallback, which is the complete functional change from this PR.

PR #1037, opened 2026-07-15
docs(quickstart): align AgentTeams runtime names

RerankerGuo:
## Summary

- Update the English and Chinese quickstarts to use the current `AGENTTEAMS_*` install and Worker environment contract.
- Replace retired local container, image, volume, network, env-file, and `*.hiclaw.io` domain names with the values created by the current installer.
- Keep intentional compatibility names such as the `hiclaw` CLI, install script paths, and `hiclaw-docker-proxy`.

## Why

The AgentTeams hard cut updated the installer and Worker entrypoint, but both quickstarts still contained copy-paste commands for retired `HICLAW_*` variables and `hiclaw-*` containers. Those commands now target missing containers or fail Worker startup because `worker-entrypoint.sh` requires `AGENTTEAMS_*`.

## Verification

- Compared every changed value against `Makefile`, `install/hiclaw-install.sh`, and `worker/scripts/worker-entrypoint.sh`.
- Ran a focused stale-contract scan across both quickstarts.
- `git diff --check`

Refs #958

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -61,7 +61,7 @@ For YAML-driven workflows, use `install/hiclaw-apply.sh` (copies files into `age
 
 ### 1.2 Login to Element Web
 
-Open http://127.0.0.1:18088 in your browser (direct access port). Alternatively, access via the gateway at http://matrix-client-local.hiclaw.io:18080 if you've added the domain to your `/etc/hosts`.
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing as superseded by the current documentation. The quickstart on main already reflects the AgentTeams runtime and container contracts covered by this stale branch.

PR #1038, opened 2026-07-15
docs(worker): align AgentTeams runtime contract

RerankerGuo:
## Summary

- Update the English and Chinese Worker guides to use current AgentTeams container, image, environment, domain, and provider names.
- Correct the internal Matrix gateway port and the default storage bucket shown in copy-paste examples.
- Describe the current change-triggered and fallback file-sync behavior and remove an obsolete `create-worker.sh` reference.

## Why

The runtime hard cut moved Worker startup to `AGENTTEAMS_*`, `agentteams-*`, and `*.agentteams.io`, but both guides still documented retired values. Following the existing commands can target missing containers, fail required-variable validation, or read a nonexistent model provider key.

## Verification

- Compared the documented contract with `Makefile`, `shared/lib/hiclaw-env.sh`, `worker/scripts/worker-entrypoint.sh`, and the Worker OpenClaw template.
- Confirmed no open PR modifies either Worker guide.
- Ran a stale-contract scan across both guides.
- Checked bilingual command parity and relative documentation links.
- `git diff --check`

Refs #958


RerankerGuo:
Closing as superseded by the current documentation. The worker guide on main already reflects AgentTeams container names, environment variables, endpoints, and sync behavior.

PR #1042, opened 2026-07-16
fix(install): restore mode-specific step routing

RerankerGuo:
## What changed

- let mode-specific installer steps reach their dedicated `should_skip_step` cases
- preserve Keep-All upgrade behavior in each dedicated case
- add a focused shell regression test for Quick Start, non-interactive, Embedded, Keep-All, and manual routing

## Why

Bash `case` statements stop at the first matching pattern. The Keep-All pattern also listed `step_manager_runtime`, `step_e2ee`, `step_docker_proxy`, `step_idle`, and `step_hostshare`, so their later mode-specific cases were unreachable.

As a result, Quick Start did not skip four advanced prompts, the non-interactive routing contract failed for five steps, and Embedded mode did not skip Docker proxy configuration at the router level. Some step functions had defensive non-interactive guards, but Quick Start still reached interactive prompts and the routing helper contradicted its own rules.

## Validation

- the new test against `origin/main` fails 10 assertions (19/29 pass)
- `bash install/tests/test-step-routing.sh` passes 29/29 on this branch
- `bash -n` passes for every tracked shell script
- ShellCheck error-level scan passes for both changed scripts
- `git diff --check`

### DIFF
--- a/install/agentteams-install.sh
+++ b/install/agentteams-install.sh
@@ -1677,8 +1677,9 @@ should_skip_step() {
             local _env="${AGENTTEAMS_ENV_FILE:-${HOME}/agentteams-manager.env}"
             [ ! -f "${_env}" ] && return 0
             ;;
-        # Keep-All upgrade mode: skip all config steps (step_volume/step_workspace handled separately)
-        step_llm|step_admin|step_network|step_ports|step_domains|step_github|step_skills|step_runtime|step_manager_runtime|step_dashboard|step_e2ee|step_docker_proxy|step_idle|step_hostshare)
+        # Keep-All upgrade mode: skip base config steps. Steps with additional
+        # mode-specific rules are handled by their dedicated cases below.
+        step_llm|step_admin|step_network|step_ports|step_domains|step_github|step_skills|step_runtime|step_dashboard)
             [ "${AGENTTEAMS_UPGRADE}" = "1" ] && [ "${AGENTTEAMS_UPGRADE_KEEP_ALL}" = "1" ] && return 0
             ;;
         step_volume|step_workspace)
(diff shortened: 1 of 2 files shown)


RerankerGuo:
CI attribution for head `34a3b2a`, workflow run [#1668](https://github.com/agentscope-ai/AgentTeams/actions/runs/29477087577):

- Translation, change detection, the controller image, and all six downstream image builds passed.
- I inspected the raw logs for all 10 failed integration jobs. Every job has exactly one empty `AGENTTEAMS_LLM_API_KEY`, one `LLM API Key is required` installer error, `make test-embedded` exiting with error 2, and zero integration-test start markers.
- The installer stops in `step_llm`, before it reaches any of the mode-specific steps changed by this PR.

The remaining red checks are therefore the fork-secret bootstrap issue addressed by #1026, not a branch-specific failure. Locally, the focused routing regression fails 10 assertions on `origin/main` and passes all 29 assertions on this head; repository-wide shell syntax and changed-file error-level ShellCheck also pass.

shiyiyue1102:
The installer routing bug is real, but the PR is not ready yet. The new test loads should_skip_step through source process substitution; on the supported macOS Bash 3 environment that producer races with source, so the function is missing and 23 assertions fail. Please change the test to load the function synchronously in a Bash-3-compatible way, rerun it on macOS and Linux, and update the branch so the currently failed required integration checks are green before final review.

RerankerGuo:
Rebuilt on current `main` and fixed the macOS Bash 3 race called out in review. The test now materializes `should_skip_step` into a temporary file before sourcing it, instead of sourcing process substitution.

Verification on macOS system Bash 3.2.57:
- `/bin/bash install/tests/test-step-routing.sh` — 30 passed, 0 failed
- `/bin/bash -n install/agentteams-install.sh install/tests/test-step-routing.sh`
- `git diff --check`

The assertions cover quick-start, non-interactive, embedded, keep-all upgrade, and manual routing.

PR #1043, opened 2026-07-16
fix(cli): reject unsafe plugin archive links

RerankerGuo:
## Summary

- reject symbolic links, hard links, and special-file members before extracting plugin tarballs
- add a CLI regression that attempts a real symlink-based write outside the extraction directory

## Root cause

`_safe_extract_tar` validated every member path before calling `extractall`, but a link member can change how a later, otherwise valid member path resolves during extraction. A crafted plugin package could therefore write outside the temporary extraction directory.

The plugin package contract only needs regular files and directories, so rejecting all other member types closes this path without changing normal TeamHarness packages.

## Validation

- New regression on `origin/main`: fails after creating `outside/pwned.txt`
- New regression on this branch: passes and leaves the outside path untouched
- Existing unaffected CLI package/install/update/uninstall paths: 5/5 pass
- Full CLI suite with the non-overlapping #1035 assertion correction applied: 6/6 pass
- Production module compiles with Python 3.8 and Python 3.11
- `git apply --check` confirms #1035 remains independently applicable
- `git diff --check`

### DIFF
--- a/plugins/cli/src/agentteams_cli/plugin_manager.py
+++ b/plugins/cli/src/agentteams_cli/plugin_manager.py
@@ -67,6 +67,8 @@ def _safe_extract_tar(package: Path, target: Path) -> None:
                     raise ValueError(f"unsafe tar member: {member.name}")
                 if member.name.startswith("/") or ".." in Path(member.name).parts:
                     raise ValueError(f"unsafe tar member: {member.name}")
+                if not (member.isfile() or member.isdir()):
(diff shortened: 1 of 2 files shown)


shiyiyue1102 (approved):
LGTM. The extraction guard now rejects symlinks, hard links, and special-file members before tar extraction, and the regression test covers the symlink escape path. The scoped required check is green.

PR #569, opened 2026-04-06
feat: add Codex runtime support and improve manager coordination

zhoubintao-pku:
## Summary
- add Codex runtime support for manager and workers, and wire the related install/build/runtime paths
- improve manager coordination by triggering Codex heartbeat turns and auto-joining workers into project rooms with joined-members validation
- add regression coverage for heartbeat and project-room coordination, and update related agent prompts/changelog

## Test Plan
- [x] `python3 -m unittest discover -s manager/tests -p 'test_*.py'`
- [x] `bash manager/tests/test_create_project.sh`
- [x] `bash tests/test-05-heartbeat.sh`

PR #1045, opened 2026-07-16
fix(replay): capture immediate manager replies

RerankerGuo:
## Summary

- snapshot the latest Manager event before sending a replay task
- pass that event through the existing `after_event` boundary so an immediate reply is recognized as new
- avoid the history request in `REPLAY_WAIT=0` mode and add a deterministic command-stub regression

## Root cause

`wait_for_manager_reply` read its baseline after `send_message`. If the Manager replied before that read completed, the new reply became the baseline and the script waited until timeout for another event. The function already accepted an `after_event` argument, but never used it.

## Validation

- regression on `origin/main`: exits 1 with `Timeout: no reply` even though the fake Matrix API already contains `fast reply`
- `scripts/tests/test-replay-fast-reply.sh` (2/2 assertions pass after the fix)
- `bash -n scripts/replay-task.sh scripts/tests/test-replay-fast-reply.sh`
- `shellcheck -S warning scripts/replay-task.sh scripts/tests/test-replay-fast-reply.sh`
- `git diff --check`

## Overlap check

PR #569 changes the runtime-readiness block in the same script but does not touch the send/reply event boundary. This focused patch is behaviorally independent.

### DIFF
--- a/scripts/replay-task.sh
+++ b/scripts/replay-task.sh
@@ -205,11 +205,7 @@ wait_for_manager_reply() {
     local timeout="${4:-${REPLY_TIMEOUT}}"
     local elapsed=0
 
-    # Snapshot the latest event_id from manager before we start waiting
-    local baseline_event
-    baseline_event=$(read_messages "${token}" "${room_id}" 5 2>/dev/null | \
-        jq -r --arg user "@${MANAGER_USER}:" \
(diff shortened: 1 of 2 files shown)


shiyiyue1102 (approved):
LGTM. Capturing the latest Manager event before sending closes the fast-reply race without adding history reads to REPLAY_WAIT=0. The focused regression covers both paths, the diff is scoped, and the required check is green.

PR #999, opened 2026-07-05
feat(install): integrate TaDashboard as optional component

nillikechatchat:
## Summary

Integrate TaDashboard as an optional component into the AgentTeams installation.

### Changes
- `install/hiclaw-install.sh`: add Dashboard wizard step, container startup, i18n messages, env persistence
- `install/hiclaw-verify.sh`: add Dashboard accessibility check
- `Makefile`: add `install-dashboard`, `update-dashboard`, `uninstall-dashboard`, `build-dashboard` targets

### Environment variables
| Variable | Default | Description |
|----------|---------|-------------|
| `HICLAW_DASHBOARD` | `1` | Enable/disable Dashboard installation |
| `HICLAW_PORT_DASHBOARD` | `13000` | Dashboard host port |
| `HICLAW_DASHBOARD_IMAGE` | `hiclaw-dashboard:latest` | Dashboard image |
| `HICLAW_AI_GATEWAY_ADMIN_URL` | auto-detected | Higress Console URL for shared auth |

PR #1046, opened 2026-07-16
fix(verify): route embedded checks to controller

RerankerGuo:
## Summary

- detect the embedded controller separately from the Manager container
- run MinIO and Matrix localhost checks in the controller while keeping Agent health checks in the Manager
- discover custom gateway/console host ports from Docker or Podman mappings when container env does not expose them
- preserve the legacy single-container fallback and add a recording regression for both topologies

## Root cause

The post-install verifier still assumed all services shared the Manager network namespace. Current embedded installs run Tuwunel and MinIO inside `agentteams-controller` and create `agentteams-manager` as a separate container on `agentteams-net`, so probing the Manager at `127.0.0.1:9000` and `:6167` reports false failures. The controller also does not receive custom host-port values as environment variables, making the hard-coded port defaults inaccurate for customized installs.

## Validation

- regression on `origin/main`: embedded topology reports `4/6` while legacy reports `6/6`
- `install/tests/test-verify-container-routing.sh`: embedded and legacy assertions both pass after the fix, including custom `19080`/`19001` mappings
- `bash -n install/hiclaw-verify.sh install/tests/test-verify-container-routing.sh`
- `shellcheck -S warning install/hiclaw-verify.sh install/tests/test-verify-container-routing.sh`
- `git diff --check`

## Overlap check

PR #999 only adds a Dashboard probe, and #569 only adds a Codex Agent-health branch. Neither changes the embedded infrastructure-container routing or mapped-port discovery addressed here.

### DIFF
--- a/install/agentteams-verify.sh
+++ b/install/agentteams-verify.sh
@@ -3,6 +3,7 @@
 #
 # Usage:
 #   bash install/agentteams-verify.sh [container_name]   # default: agentteams-manager
+#   AGENTTEAMS_VERIFY_INFRA_CONTAINER=<name>         # default: agentteams-controller
 #
 # Runs 7 read-only reachability checks and prints PASS/FAIL per check.
 # Exit code: 0 if all pass, 1 if any fail.
@@ -22,8 +23,9 @@
 #        kubectl get pod -l app=agentteams-manager -o jsonpath='{.items[0].metadata.name}'
 #
 #   2. Internal service checks (checks #2, #3, #6)
-#      These use `docker exec ... curl 127.0.0.1:PORT` which works because all
-#      services share a single container network namespace.
+#      Local embedded installs route infrastructure checks to the controller
+#      container and the Agent check to the Manager container. Legacy installs
+#      fall back to the Manager container for all three checks.
 #      In K8s each service is a separate Pod/Service; replace with:
 #        `kubectl exec <manager-pod> -- curl http://<service-name>.<ns>.svc:PORT`
 #      or use `kubectl port-forward svc/<name> LOCAL:REMOTE` for a one-shot probe.
@@ -39,7 +41,8 @@
 
 # No set -e: each check is independent; failures do not abort subsequent checks.
 
-CONTAINER="${1:-agentteams-manager}"
+MANAGER_CONTAINER="${1:-agentteams-manager}"
+INFRA_CONTAINER="${AGENTTEAMS_VERIFY_INFRA_CONTAINER:-agentteams-controller}"
 
 # ---------- Docker/Podman detection ----------
 # TODO(k8s): extend to three-way detection (docker / podman / kubectl)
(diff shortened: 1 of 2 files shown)


RerankerGuo:
CI attribution for run [29483183894](https://github.com/agentscope-ai/AgentTeams/actions/runs/29483183894): translation, change detection, the controller/base build, and all six downstream image builds passed.

I audited the raw archive for all 10 failed integration jobs. Every job has exactly one empty `AGENTTEAMS_LLM_API_KEY`, exits from `install-embedded` with `LLM API Key is required`, and records no test-start marker. There are zero `hiclaw-verify` invocations, so the changed verifier never ran. This is the same base-workflow bootstrap issue addressed by #1026; no branch-specific code change is warranted.

The focused local routing regression remains green for both embedded and legacy topologies (`2 passed, 0 failed`).

shiyiyue1102:
Thanks for the detailed CI attribution. I reviewed the verifier routing and reran the focused regression locally: both embedded and legacy routing cases pass. The implementation matches the current embedded topology by keeping infrastructure checks in the controller container and the Agent health check in the Manager container.

The remaining blocker is that the required integration checks on the current head are still recorded as failed, even though the archived jobs appear to have exited before the tests because the LLM key was empty. Please update or rebase the branch onto the latest `main` and rerun the required checks so the PR has a clean, current CI result. Once that is green, this should be ready for final review.

EricZHANG1688:
> Thanks for the fix — exactly the symptom I was hitting.
>
> I'd like to add a host-side access perspective as a downstream application developer:
>
> After deploying the `latest` image locally, I noticed that **no host-side tool can reach Matrix 6167 directly**. I built a two-account round-trip verification (Alice sends via Tuwunel → Bob reads from the same room via a different `access_token`), and it works inside the container via `docker exec`, but there's no way for an external client to connect from the host.
>
> Repro on host (after a fresh `install/hiclaw-install.sh` deploy):
>
> ```bash
> # Inside the container — works
> $ docker exec hiclaw-controller curl -s -o /dev/null -w "%{http_code}\n" \
>     http://127.0.0.1:6167/_matrix/client/versions
> 200
>
> # From the host — fails
> $ curl -s -o /dev/null -w "%{http_code}\n" --max-time 3 \
>     http://localhost:6167/_matrix/client/versions
> 000
>
> # docker ps confirms 6167 is not published
> $ docker ps --filter name=hiclaw-controller --format '{{.Ports}}'
> 127.0.0.1:18001->8001/tcp, 127.0.0.1:8099->8080/tcp, 127.0.0.1:18088->8088/tcp
> # 6167 missing
> ```
>
> I traced `install/hiclaw-install.sh` and found that only four host ports are published: 18080/18001/18088/18888. 6167 is intentionally internal. The Helm `tuwunel-service.yaml` is `type: ClusterIP`, so it is also unreachable from outside the cluster.
>
> This is **complementary** to #1046's fix direction — #1046 addresses the intra-cluster "where to run the localhost check" problem; my concern is the **cross-cluster** "host users cannot reach Matrix at all" problem. They don't conflict, but the documentation gap is real either way.
>
> Suggested options, in order of preference:
>
> 1. **Default-publish 6167** (best, with `AGENTTEAMS_MATRIX_EXPOSE_PORT` env var for opt-in/opt-out, matching least-privilege)
> 2. **Document the workaround** in README: "host users can verify Matrix via `docker exec <container> curl http://127.0.0.1:6167`"
> 3. **Provide a bridge subcommand**: `agentteams matrix-port-forward [--port 6167]` for host-side tools
>
> I have a 7-check reproduction script covering HTTP probes, TCP socket probes, `docker inspect` IP routing, port mapping comparison, and `docker-compose.yml` static scan. Happy to share the full script or a minimal gist if useful.
>
> Thanks again for the responsive maintenance on the embedded topology!

RerankerGuo:
Rebuilt on current `main` and migrated the patch to the canonical `install/agentteams-verify.sh` entrypoint. The verifier now routes MinIO/Matrix checks to the embedded controller while keeping the runtime health check in the Manager, with legacy single-container fallback preserved.

Verification:
- `bash install/tests/test-verify-container-routing.sh` — 2 passed, 0 failed
- `bash -n install/agentteams-verify.sh install/tests/test-verify-container-routing.sh`

PR #1047, opened 2026-07-16
fix(scripts): redact complete Matrix events

RerankerGuo:
## Summary

- apply the existing recursive PII redaction to the complete formatted Matrix event
- cover sender IDs, state-event content, media URLs, and relation metadata
- redact values whose JSON field names identify passwords, tokens, or API credentials even when their values are short or provider-specific
- preserve the explicit `--no-redact` behavior and add focused regression tests

## Problem

The debug exporter advertises PII-safe output by default, but `format_event` only redacted the message body. Matrix sender IDs and non-message content were exported verbatim, while message URLs and relation metadata also bypassed redaction. Recursive string-only handling additionally missed generic secrets represented as JSON fields such as `password` or `apiKey`. These gaps could leak phone numbers, email addresses, IPs, or credentials when debug bundles are shared.

## Testing

- `python3 -m unittest discover -s scripts/tests -p "test_*.py" -v` (4 tests)
- `python3 -m py_compile scripts/export-debug-log.py scripts/tests/test_export_debug_log.py`
- `git diff --check`

Baseline assertions against current `main` reproduce both the Matrix metadata leak and secret-named JSON field leak; the same assertions pass with this change.

### DIFF
--- a/scripts/export-debug-log.py
+++ b/scripts/export-debug-log.py
@@ -72,6 +72,14 @@ _PII_PATTERNS: list[tuple[str, re.Pattern]] = [
     ("SSN",           re.compile(r'\b\d{3}-\d{2}-\d{4}\b')),
 ]
 
+_SECRET_FIELD_PATTERN = re.compile(
+    r'(?i)^(?:password|passwd|pwd|secret|token|api[_-]?key'
+    r'|access[_-]?key(?:[_-]?secret)?|secret[_-]?access[_-]?key'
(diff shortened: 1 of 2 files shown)


shiyiyue1102 (approved):
LGTM. Default debug export now redacts the complete formatted Matrix event, including sender IDs, relation metadata, media URLs, and secret-named JSON fields, while preserving the explicit no-redact path. The focused tests pass and the scoped required check is green.

PR #1044, opened 2026-07-16
fix(import): align wrappers with current CLI flags

RerankerGuo:
## Summary

- stop Bash and PowerShell import wrappers from advertising or forwarding flags that the current `hiclaw` CLI does not implement
- document structured `mcpServers` configuration through YAML and align English/Chinese examples
- add a recording regression test for supported forwarding, unsupported rejection, and wrapper help

## Root cause

The wrappers lagged behind the current Go CLI contract after its apply/import behavior changed. They still accepted `--mcp-servers`, `--dry-run`, and `--prune`, so users only saw a failure after the arguments reached the container CLI.

## Validation

- `PWSH_BIN=/tmp/powershell-7.6.3/pwsh install/tests/test-import-wrapper-contract.sh` (`14 passed, 0 failed`; baseline was `2 passed, 8 failed`)
- existing import package tests with PR #1034 applied (`6 passed, 0 failed`)
- `bash -n install/hiclaw-import.sh install/hiclaw-apply.sh install/tests/test-import-wrapper-contract.sh`
- `shellcheck -S warning install/hiclaw-import.sh install/hiclaw-apply.sh install/tests/test-import-wrapper-contract.sh`
- PowerShell AST parse
- `go run ./cmd/hiclaw apply worker --help` confirms the removed flags are absent
- `git diff --check`

PR #1048, opened 2026-07-16
fix(migrate): print runnable ZIP import command

RerankerGuo:
## Summary

- include the required `worker` resource type and Worker name in generated ZIP import instructions
- reject normalized Worker names that the `hiclaw` CLI cannot accept
- add an end-to-end regression test for the import command and invalid names

## Problem

`generate-zip.sh` currently tells users to run `bash hiclaw-import.sh --zip ...`. The current import wrapper treats its first argument as a resource type, so that command fails with `Unknown resource type: --zip`.

The generator can also normalize inputs such as `!!!` to an empty name or preserve a leading hyphen, then build a package and print an import command that `hiclaw apply worker` rejects. The generated instructions should be directly runnable and follow the CLI's Worker-name contract.

## Testing

- baseline generator completes the ZIP but fails the new command assertion
- `bash migrate/skill/tests/test-generate-zip.sh`
- `shellcheck migrate/skill/tests/test-generate-zip.sh`
- `bash -n migrate/skill/scripts/generate-zip.sh migrate/skill/tests/test-generate-zip.sh`
- `git diff --check`

Full ShellCheck on the existing generator reports only its pre-existing SC2018/SC2019 `tr` warnings. This does not modify the import wrappers changed by #1044; it fixes their standalone migration-package caller.

### DIFF
--- a/migrate/skill/scripts/generate-zip.sh
+++ b/migrate/skill/scripts/generate-zip.sh
@@ -38,6 +38,11 @@ fi
 # Normalize name
 WORKER_NAME=$(echo "${WORKER_NAME}" | tr 'A-Z' 'a-z' | tr -cd 'a-z0-9-')
 
+if [[ ! "${WORKER_NAME}" =~ ^[a-z0-9][a-z0-9-]*$ ]]; then
+    echo "ERROR: Invalid worker name: must start with a lowercase letter or digit and contain only lowercase letters, digits, and hyphens"
+    exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


shiyiyue1102 (approved):
LGTM. The generated migration instruction now matches the import wrapper contract by including the worker resource type and name, and the normalized name validation matches the CLI worker-name rules. The end-to-end regression passes and the required check is green.

PR #1049, opened 2026-07-16
fix(migrate): analyze current cron job format

RerankerGuo:
## Summary

- read current OpenClaw cron files from their `jobs` array
- retain support for legacy top-level job arrays
- emit genuinely empty dependency categories as `[]`, not `[""]`
- add end-to-end analysis regressions for both cron formats and an empty environment

## Problem

`generate-zip.sh` already documents the current cron schema as `{"version":1,"jobs":[...]}`, but `analyze.sh` still evaluates `.[].payload...` as if the root were a job array. jq errors are intentionally suppressed by the heuristic scanner, so current-format cron commands silently disappear from `tool-analysis.json` and their required binaries are omitted from migration planning.

The JSON serialization also pipes empty Bash arrays through `printf`, which emits one blank line and turns every empty dependency category into `[""]`. That pollutes the analysis report and downstream migration manifest with a fake package/binary entry.

## Testing

- the regression reports `cron_payload_commands: 0` and misses the cron-only binary on current `main`
- an empty baseline analysis produces `[""]` for apt, pip, npm, and unknown binaries; the fixed output uses `[]`
- `bash migrate/skill/tests/test-analyze-cron.sh`
- `shellcheck -S warning migrate/skill/tests/test-analyze-cron.sh`
- `bash -n migrate/skill/scripts/analyze.sh migrate/skill/tests/test-analyze-cron.sh`
- `git diff --check`

The test verifies current object-wrapped jobs, legacy top-level arrays, and an environment with no detected dependencies. This is separate from #1048, which only corrects the final generated import command.

### DIFF
--- a/migrate/skill/scripts/analyze.sh
+++ b/migrate/skill/scripts/analyze.sh
@@ -249,10 +249,10 @@ fi
 log "Step 7: Generating tool-analysis.json..."
 
 # Build JSON using jq
-APT_JSON=$(printf '%s\n' "${APT_PACKAGES[@]}" 2>/dev/null | jq -R . | jq -s . 2>/dev/null || echo '[]')
-PIP_JSON=$(printf '%s\n' "${PIP_PACKAGES[@]}" 2>/dev/null | jq -R . | jq -s . 2>/dev/null || echo '[]')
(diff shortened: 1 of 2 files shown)


shiyiyue1102 (approved):
LGTM. The analyzer now supports both the current object-wrapped cron format and the legacy top-level array, and empty dependency categories serialize as real empty arrays. The change is scoped, includes focused regression coverage, and the required check is green.

PR #1050, opened 2026-07-16
fix(hack): persist containerized skopeo auth

RerankerGuo:
## Summary

- mount a persistent host auth directory for containerized skopeo
- explicitly point `REGISTRY_AUTH_FILE` at the mounted auth file
- keep inspect/copy mounts read-only and login mounts read-write
- check stored credentials with `skopeo login --get-login` instead of probing a fixed image tag
- allow overriding the host file with `SKOPEO_AUTH_FILE`

## Problem

The current `quay.io/skopeo/stable:latest` image sets `REGISTRY_AUTH_FILE=/tmp/auth.json`, while `mirror-images.sh` mounts the host credentials at `/.config/containers`. The image therefore ignores that mount: interactive login writes into its temporary filesystem, the credentials disappear when the container exits, and the following copy cannot reuse them.

The old login check also inspects `${TARGET_NS}/all-in-one:latest`. Valid credentials are misreported as missing whenever that repository or tag does not exist. The fix mounts the auth file parent at `/auth`, overrides `REGISTRY_AUTH_FILE` for every container invocation, and asks skopeo directly whether the registry has a stored login. Mounting the directory permits skopeo to atomically create or replace `auth.json`.

## Testing

- baseline command-stub run records all inspect/login/copy calls using the ignored `/.config/containers` mount and fails the persistence assertion
- the fixed test performs an initial login, then verifies a second run reuses credentials without another interactive prompt even though image inspection would fail
- `bash hack/tests/test-mirror-container-auth.sh`
- `shellcheck -S warning hack/mirror-images.sh hack/tests/test-mirror-container-auth.sh`
- `bash -n hack/mirror-images.sh hack/tests/test-mirror-container-auth.sh`
- `git diff --check`

The regression verifies that all calls share `/auth/auth.json`, login is read-write, later use is read-only, and the auth file remains on the host.

### DIFF
--- a/hack/mirror-images.sh
+++ b/hack/mirror-images.sh
@@ -27,6 +27,9 @@
 #
 #   # Override target registry / namespace
 #   TARGET_REGISTRY=my-registry.example.com TARGET_NS=myns ./hack/mirror-images.sh
+#
+#   # Override the persistent auth file used in container mode
+#   SKOPEO_AUTH_FILE=/path/to/auth.json USE_CONTAINER=1 ./hack/mirror-images.sh
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


shiyiyue1102 (approved):
LGTM. Containerized Skopeo operations now share an explicit persistent auth file, with read-only mounts for non-interactive operations and a writable mount only for login. The regression verifies first login, credential reuse, and mount modes, and the required check is green.

PR #1051, opened 2026-07-16
test: honor AgentTeams GitHub token env

RerankerGuo:
## Summary

- normalize GitHub test credentials into `TEST_GITHUB_TOKEN`, preferring the current `AGENTTEAMS_GITHUB_TOKEN` contract
- load both current and legacy GitHub token names from an existing installation's env file
- document the current variable and add a focused precedence regression test

## Problem

The installers persist and pass `AGENTTEAMS_GITHUB_TOKEN`, but integration tests 08-13 only recognize `TEST_GITHUB_TOKEN` or the retired `HICLAW_GITHUB_TOKEN`. Running the suite with the generated AgentTeams env therefore reports that no GitHub token is configured and skips those tests.

The normalization preserves explicit test overrides and the legacy variable while making the current installer output work directly.

## Testing

- the regression pointed at `origin/main` fails with `expected token canonical, got <empty>`
- `bash tests/unit/test-github-token-env.sh`
- direct test 08 with only `AGENTTEAMS_GITHUB_TOKEN` reaches the later LLM prerequisite instead of the GitHub-token skip
- `bash -n tests/lib/test-helpers.sh tests/run-all-tests.sh tests/unit/test-github-token-env.sh`
- `shellcheck -S error tests/lib/test-helpers.sh tests/run-all-tests.sh tests/unit/test-github-token-env.sh`
- `git diff --check`

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -76,7 +76,7 @@ make test SKIP_INSTALL=1
 | Variable | Required | Description |
 |----------|----------|-------------|
 | `HICLAW_LLM_API_KEY` | Yes | LLM API key for Agent behavior |
-| `HICLAW_GITHUB_TOKEN` | No | GitHub PAT for tests 08-11 |
+| `AGENTTEAMS_GITHUB_TOKEN` | No | GitHub PAT for tests 08-13 (`HICLAW_GITHUB_TOKEN` remains supported) |
 
 ## Helper Libraries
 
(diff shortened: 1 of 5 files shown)


shiyiyue1102:
This is a useful test-harness compatibility cleanup, but its standalone value is limited because it does not change runtime behavior. Before further review, please either link it to a concrete currently failing workflow caused by the GitHub token variable mismatch, or consolidate it with the related test-environment contract cleanup such as #1028. Keeping several small environment-variable test PRs separate adds more review and merge overhead than the changes justify.

RerankerGuo:
Closing to reduce review overhead as requested. This test-only compatibility change has no linked current failing workflow and overlaps the broader test-environment contract cleanup. It can be reconsidered if a reproducible failure remains on current main.

PR #1052, opened 2026-07-16
fix(manager): emit runnable template install commands

RerankerGuo:
## Summary

- stop accepting and forwarding the unsupported `--mcp-servers` override from the Nacos Worker template installer
- keep the wrapper's own `--dry-run` preview without adding an unsupported flag to the reported `hiclaw_args`
- align the agent-facing runtime list with the current CLI and add a focused command-stub regression

## Root cause

`hiclaw apply worker` supports package, model, skills, and runtime overrides, but it has no `--mcp-servers` or `--dry-run` flags. The installer advertised both: a normal MCP override failed in Cobra, while dry-run JSON presented a command that could not be executed. Structured MCP configuration remains part of the confirmed template package.

## Verification

- `bash manager/tests/test-install-worker-template.sh` (3/3)
- baseline dry-run output contains both unsupported flags; patched output contains neither
- `go run ./cmd/hiclaw apply worker --help` confirms the supported flag set
- Bash syntax and warning-level ShellCheck pass for the script and regression test
- `git diff --check`

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -27,6 +27,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `openclaw-bas
 
 **Bug Fixes**
 
+- **Worker template install CLI contract**: Nacos template imports now emit only flags supported by `hiclaw apply worker`; dry-run previews remain runnable, and unsupported command-line MCP overrides are no longer advertised.
 - **Legacy Team channel policy**: Legacy Team reconciliation now writes final Matrix channel allow-lists to member runtime config and re-adds the Team Leader to the Manager allow-list so controller integration tests observe durable policy state.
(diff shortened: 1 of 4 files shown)


shiyiyue1102:
The wrapper-contract cleanup is useful, but the current standalone value is reduced by overlap with #1044 and #1053, and this branch now conflicts with main. Please wait for the import and runtime-switch contracts to settle, then rebase and keep only the Nacos template installer behavior that is still missing. If the remaining diff is only another expression of the same unsupported MCP argument problem, consolidate it with the relevant contract PR instead of maintaining a separate branch.

RerankerGuo:
Closing to follow the maintainer consolidation request. The branch overlaps #1044 and #1053, and no distinct current-main failure has been demonstrated for the remaining Nacos installer behavior. It can be revisited after those contracts settle if a focused gap remains.

PR #1060, opened 2026-07-18
fix(copaw): route Team Leader assignments to Team Room

shiyiyue1102:
## Summary
- route concrete Team Leader worker assignments sent through the CoPaw `message` tool from Leader DM to Team Room
- record the outbound event in the effective Team Room session
- add a regression test reproducing the beta1 Integration Tests failure

## Root cause
Integration Tests run 29590753647 showed the Leader using the `message` tool to assign a worker, but the tool sent directly to the requested Leader DM and bypassed the Matrix channel reroute.

## Validation
- `PYTHONPATH=src .venv/bin/pytest tests/test_message_tool.py tests/test_worker_matrix_channel.py -q` (27 passed)
- `git diff --check`

## Dependency
This branch is stacked on #1058 so the image-affecting fix is recorded in the fresh post-beta1 `changelog/current.md`. Merge #1058 first; its archive diff will then disappear from this PR.

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -4,3 +4,6 @@ Record image-affecting changes to `manager/`, `worker/`, `openclaw-base/` here b
 
 ---
 
+**Bug Fixes**
+
+- **CoPaw Team coordination routing**: Route Team Leader worker assignments sent through the `message` tool from Leader DM to Team Room, matching the Matrix channel send path.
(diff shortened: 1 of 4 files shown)


maplefeng-a (approved):
CI is green and the CoPaw Team Room routing fix looks good to me.

PR #1072, opened 2026-07-23
refactor: hard cut Team and Worker CR contracts

shiyiyue1102:
## What changed

- Make `Worker` CRs the sole owners of runtime configuration, identity, resources, skills, MCP/package settings, channel policy, and lifecycle.
- Make `Team` CRs reference existing Workers through `spec.workerMembers`, with exactly one `team_leader`.
- Remove inline Team member specs, registry migration/write-back, legacy annotations, Team-owned Worker runtime reconciliation, and compatibility fallbacks.
- Update the REST API, `agt` CLI, authorization, Manager skills/scripts, installers, documentation, Helm CRDs, and integration tests to use the terminal CR model.
- Preserve Worker CRs when a Team is deleted and block Worker deletion while it is referenced by a Team.

## Why

The Team and Worker CRDs are new contracts and do not need migration compatibility. Keeping inline members, registry files, annotations, and CR-backed state in parallel created multiple sources of truth and made ownership ambiguous.

This change hard-cuts the repository to one end-to-end model: Worker owns runtime state; Team owns membership and coordination.

## Impact

- Existing callers must create or update Worker CRs before referencing them from a Team.
- Old `spec.leader`, `spec.workers`, registry files, Team/role annotations, and CLI compatibility flags are no longer accepted.
- Deleting a Team no longer cascades into deleting its Worker resources.

## Validation

- `GOCACHE=/tmp/agentteams-hard-cut-go-cache go test ./...`
- `go build ./...`
- Full controller integration suite and focused `TestTeam` integration tests
- Shell syntax checks for all changed `.sh` files
- PowerShell installer parser validation
- Controller/Helm Team CRD byte-for-byte comparison
- `git diff --check`
- Targeted scan for removed compatibility identifiers


shiyiyue1102:
@johnlanni @Jing-ze @maplefeng-a The latest checks are all passing and this PR is ready for review. Could you take a look when convenient?

maplefeng-a (commented):
Thanks for the hard-cut cleanup. The direction looks aligned with moving Team/Worker to explicit CR ownership, but I think two contract edges should be fixed before this lands:

1. The "block Worker deletion while referenced by a Team" protection currently only covers the REST API path. `DeleteWorker` rejects a referenced worker through `findTeamForMember`, but direct Worker CR deletion still goes through `WorkerReconciler`'s `DeletionTimestamp -> reconcileDelete` path and removes the finalizer after cleanup. I also do not see a validating webhook/admission path in this PR. That means `kubectl delete worker <team-member>` can still delete a Worker that is referenced by a Team, which violates the new declarative CR contract. Could we add the same protection on the CR path, preferably via validation/admission, or otherwise clearly narrow the documented guarantee to the REST/CLI path?

2. `validateTeamWorkerMembers` still accepts an empty member role (`case "", "worker"`), but the generated Team CRD now requires `workerMembers[].role` and restricts it to `team_leader|worker`. A REST caller that omits the non-leader role can pass server-side validation and then fail at the Kubernetes write layer, likely as a generic 500 because `writeK8sError` does not map `Invalid`. Could we either reject empty roles in the REST validator or normalize them to `"worker"` before writing the CR?

CI is green on the current head; these look like contract/API consistency issues rather than test failures.

shiyiyue1102:
Thanks for catching these contract gaps. Fixed in d96f1ed:

- WorkerReconciler now keeps the Worker finalizer and requeues deletion while any Team still references the Worker, covering direct CR deletion.
- The REST Team validator now rejects empty member roles before attempting the Kubernetes write.

Validated with the full agentteams-controller test suite (go test ./...).

maplefeng-a (approved):
The contract gaps I raised have been addressed, and the current head is green. Approving.

PR #1073, opened 2026-07-24
ci: make integration tests safe for fork PRs

shiyiyue1102:
## Summary

- run pull request validation in the unprivileged `pull_request` context
- keep fork and Dependabot runs read-only and select only the four secret-free controller/CR test shards
- preserve the full ten-shard matrix, LLM secrets, metrics, and PR comments for trusted same-repository runs
- use the event's default checkout ref so fork checkouts no longer depend on unsafe `pull_request_target` behavior

## Root cause

`actions/checkout@v4` now rejects checking out fork code from a privileged `pull_request_target` workflow unless unsafe checkout is explicitly enabled. The existing integration workflow also carried write permissions and an LLM API key, so enabling that bypass would expose secrets and repository permissions to untrusted code.

## Impact

Fork pull requests can build images and run the non-secret integration coverage without checkout failures. Trusted branches, pushes, tags, and manual runs retain the existing complete integration coverage. The generated matrix removes its internal secret marker before expansion so existing job names remain stable.

## Validation

- `actionlint .github/workflows/test-integration.yml`
- `git diff --check`
- matrix simulation: 10 trusted shards, 4 fork-safe shards
- verified the workflow contains no `pull_request_target`, `allow-unsafe-pr-checkout`, or explicit fork repository checkout

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -1,7 +1,7 @@
 name: Integration Tests
 
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - main
     paths:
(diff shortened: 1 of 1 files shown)


maplefeng-a (approved):
CI is green. The workflow now uses the unprivileged pull_request event for PR validation, keeps fork and Dependabot runs on the secret-free matrix, and limits PR-commenting/metrics steps to trusted same-repository PRs. Approving under that security boundary.

PR #1079, opened 2026-07-26
fix(install): support pre-v1.2 image env contract

shiyiyue1102:
## What changed

- select exactly one controller environment contract from the requested image version
  - images older than v1.2.0 receive only the legacy environment prefix
  - v1.2.0 and newer images receive only `AGENTTEAMS_*`
- treat `latest` according to the known stable release, so the current v1.1.2 `latest` tag uses the legacy contract
- keep legacy resource names aligned with the renamed installer by using the `agentteams-` resource prefix, network, controller URL, and storage settings
- keep Matrix AppService variables exclusive to the new contract and legacy storage/resource variables exclusive to the old contract
- normalize manual `1.2.0.beta.1` input to the published `v1.2.0-beta.1` image tag
- add a version-boundary regression test and run it in CI

## Why

The v1.1.2 images were mirrored into the new AgentTeams image repository without changing their runtime contract. The current installer injected only `AGENTTEAMS_*`, while those images still read the legacy environment prefix. As a result, Tuwunel received an empty registration token and MinIO received invalid credentials, causing installation to stop while waiting for Matrix.

Selecting one contract by image version fixes old images without mixing old and new runtime configuration in the same container.

## User impact

Users can install the mirrored v1.1.2 images, including the current `latest` tag, with the renamed AgentTeams installer. Images from v1.2.0 onward continue to receive only the new environment contract.

Users may enter either the published `v1.2.0-beta.1` tag or the shorthand `1.2.0.beta.1` in the custom-version prompt.

## Validation

- `bash -n install/agentteams-install.sh`
- `git diff --check`
- isolated end-to-end installer smoke using real Matrix and LLM conversations
  - `latest` pulled the v1.1.2 embedded digest
  - v1.1.2 controller environment: legacy prefix present, `AGENTTEAMS_*` absent
  - Tuwunel server-version endpoint responded successfully
  - MinIO live-health endpoint responded successfully
  - the v1.1.2 CoPaw Manager container was created and reached `running`
  - v1.1.2 Manager replied `LATEST_112_SMOKE_OK`
  - manual `1.2.0.beta.1` input pulled `v1.2.0-beta.1` images
  - beta1 controller environment: `AGENTTEAMS_*` present, legacy prefix absent
  - beta1 Manager completed onboarding and replied `BETA_DOT_DIALOGUE_OK`

### DIFF
--- a/.github/workflows/helm-lint.yml
+++ b/.github/workflows/helm-lint.yml
@@ -4,8 +4,10 @@ on:
   pull_request:
     paths:
       - 'helm/**'
+      - 'install/agentteams-install.sh'
       - 'openhuman/scripts/openhuman-worker-entrypoint.sh'
       - 'tests/check-agentteams-rename-defaults.sh'
+      - 'tests/check-installer-pre-1.2-env-compat.sh'
       - 'tests/check-helm-agentteams.sh'
   workflow_dispatch: ~
 
@@ -18,6 +20,9 @@ jobs:
 
       - name: Check AgentTeams rename contracts
         run: bash tests/check-agentteams-rename-defaults.sh
+
+      - name: Check pre-v1.2 installer compatibility
+        run: bash tests/check-installer-pre-1.2-env-compat.sh
 
       - name: Set up Helm
         uses: azure/setup-helm@v4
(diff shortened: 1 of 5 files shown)


shiyiyue1102:
@johnlanni @Jing-ze CI is green now. Could you take a look when convenient?

maplefeng-a (approved):
Looks good. CI is green and the installer compatibility change is scoped to the pre-v1.2 environment contract.

PR #1084, opened 2026-07-27
fix(install): keep wasm plugins on primary registry

### LINKED ISSUE #429: HiClaw v1.0.8 local install: Higress gateway stays in Envoy warming because basic-auth/key-auth WasmPlugin OCI URLs in us-west-1 require auth
## Bug Description

On a fresh local HiClaw v1.0.8 install in WSL2 (Ubuntu 24.04 + Docker Desktop), the Higress gateway inside `hiclaw-manager` never became healthy.

Element Web, Matrix, and the manager container all came up, but the AI gateway listener on `:8080` stayed stuck in Envoy **warming** state, which blocked normal manager chat behavior.

Root cause appears to be that two WasmPlugin CRDs point to OCI URLs in the `us-west-1` Alibaba registry that do not allow anonymous fetches in this environment:

- `basic-auth-1.0.0`
- `key-auth.internal`

Because the controller cannot fetch those WASM OCI images, it cannot push the ECDS configs, so the listener never fully activates.

When I changed only those two plugin URLs from:

- `oci://higress-registry.us-west-1.cr.aliyuncs.com/plugins/basic-auth:2.0.0`
- `oci://higress-registry.us-west-1.cr.aliyuncs.com/plugins/key-auth:2.0.0`

to:

- `oci://higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/basic-auth:2.0.0`
- `oci://higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/key-auth:2.0.0`

the gateway immediately recovered:
- all ECDS filters reported `update_success`
- Envoy listener `0.0.0.0:8080` became active
- health checks returned HTTP 200
- AI gateway requests started working
- manager Matrix chat started responding

This looks like a registry/bootstrap issue in Higress plugin delivery, not a frontend issue.

## Environment

- Host OS: Windows 11
- Runtime: WSL2 Ubuntu 24.04.3 LTS
- Docker: Docker Desktop with WSL integration
- HiClaw version: v1.0.8
- Install mode: local/manual
- Model provider behind gateway: Ollama via local API
- Main container: `hiclaw-manager`

## Steps to Reproduce

1. Fresh install HiClaw v1.0.8 locally in WSL2.
2. Let `hiclaw-manager` start normally.
3. Check gateway state and manager behavior.
4. Observe that:
   - manager container is up
   - Matrix and Element are reachable
   - but gateway `:8080` is not healthy / appears stuck in warming
   - manager chat is non-functional or inconsistent

## Actual Result

- Gateway listener on `:8080` does not become fully active.
- Requests to the mapped host port connect but reset or fail until plugin URLs are changed.
- Manager logs show gateway startup timeout.
- Higress gateway/controller logs show repeated wasm/ECDS errors.

## Expected Result

Gateway listener `:8080` should become healthy on a fresh local install without requiring manual edits to WasmPlugin CRDs.

## Relevant Evidence

Observed symptoms/logs included:

- `Higress Gateway did not become available within 180s` :contentReference[oaicite:0]{index=0}
- repeated `wasmconvert the wasm config` errors in `higress-gateway.log` :contentReference[oaicite:1]{index=1}
- repeated ECDS/ACK failures while the listener remained unavailable :contentReference[oaicite:2]{index=2}
- anonymous access to the `us-west-1` registry failed, while `cn-hangzhou` worked in this environment 

Claude-assisted diagnosis also isolated that only `basic-auth` and `key-auth` were failing fetch/load, while `ai-proxy` and `ai-statistics` were succeeding. 

## Workaround

Back up the current WasmPlugin CRDs, then change only these two CRD `spec.url` values:

- `basic-auth-1.0.0` → `oci://higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/basic-auth:2.0.0`
- `key-auth.internal` → `oci://higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/key-auth:2.0.0`

After this change:

- all 4 ECDS filters showed `update_success`
- Envoy `0.0.0.0:8080` became active
- health check returned HTTP 200
- AI gateway requests succeeded
- manager responded in Matrix again 

## Suggested Fix

One of these would likely solve it:

1. ship these WasmPlugin URLs using a registry that works anonymously for local/public installs
2. mirror/fallback automatically if the default registry fails
3. avoid blocking listener activation on these plugin fetch failures for the local install path
4. document the registry dependency clearly in install docs if intentional

## Notes

This report is specifically about the gateway/bootstrap problem.

There is a separate runtime issue in this same environment involving `qwen3:30b-a3b` reasoning responses and OpenClaw reading only `message.content`, but that is a different bug and should be filed separately.

zcxGGmu:
## Summary
- Keep the installer-selected AgentTeams image registry regional as before.
- Add a separate `AGENTTEAMS_HIGRESS_WASM_PLUGIN_REGISTRY` defaulting to the primary `cn-hangzhou` registry for Higress WASM plugin pulls.
- Persist that WASM registry into both shell and PowerShell installer env files, and add a CI check to prevent regressing back to the regional image registry.

Fixes #429

## Testing
- `bash tests/check-installer-wasm-registry.sh`
- `bash -n install/agentteams-install.sh tests/check-installer-wasm-registry.sh`
- `python - <<'PY' ... yaml.safe_load(.github/workflows/helm-lint.yml) ... PY`
- `git diff --check`

## Notes
- PowerShell syntax parsing was not run locally because `pwsh` is not installed in this environment; the new CI check verifies the expected PowerShell env-file contract textually.


oss-maintainer (commented):
### Summary
Well-structured fix: decouples the Higress WASM plugin registry from the regional `AGENTTEAMS_REGISTRY` and pins it to the primary registry (`higress-registry.cn-hangzhou.cr.aliyuncs.com`) via the new `AGENTTEAMS_HIGRESS_WASM_PLUGIN_REGISTRY` override.

Verified across the change:
- Shell and PowerShell installers both resolve the new variable with the same default and persist it into the generated env file and `$config` respectively — consistent.
- The generated env file now writes both `AGENTTEAMS_HIGRESS_WASM_PLUGIN_REGISTRY` and `HIGRESS_ADMIN_WASM_PLUGIN_IMAGE_REGISTRY` from the same source, so upgrades converge to the primary registry.
- New guard test `tests/check-installer-wasm-registry.sh` asserts the old timezone-coupled assignment is gone in both installers, and is wired into the helm-lint workflow path filter and job steps.
- Changelog and README env-var table updated.

No issues found. One question for maintainers (non-blocking): for installs upgrading from a version whose generated env file already contains the old region-based `HIGRESS_ADMIN_WASM_PLUGIN_IMAGE_REGISTRY`, please confirm the env file is re-generated (not merely appended to) so the stale regional value cannot survive.

Note: I could not verify CLA status for this repository (no `license/cla` commit-status context is configured), so this is a COMMENT rather than an APPROVE per policy.

---

PR #1088, opened 2026-07-28
feat: add Atlas Cloud LLM preset

binyangzhu000-sudo:


binyangzhu000-sudo:
Rebased the PR onto current `main` (`124f06d1`) and resolved the `changelog/current.md` conflict. The PR is now mergeable.

Local verification on the resulting tree:
- `go test ./cmd/agt -run 'TestResolveLLMPreflight' -count=1`
- `go test ./cmd/agt -count=1`
- `bash tests/check-helm-agentteams.sh`
- `helm lint helm/agentteams/` (1 chart linted, 0 failed)
- `gofmt`, `bash -n`, conflict-marker, and diff-scope checks

The exact-head GitHub Actions runs currently show `action_required`, so they need maintainer approval to execute for this fork PR.

agentteams-controller/cmd/agt/llm_preflight.go
```diff
@@ -171,6 +172,8 @@ func resolveLLMPreflightBaseURL(provider, baseURL string) (string, error) {
 			baseURL = defaultOpenAICompatibleBaseURL
 		case "qwen":
 			baseURL = defaultQwenCompatibleBaseURL
+		case "atlascloud":
```
oss-maintainer, line 175: The atlascloud preset is wired into the preflight CLI and the Helm helper, but not into the controller initializer that actually creates the Higress LLM provider (agentteams-controller/internal/initializer/initializer.go:301 — switch over qwen | openai-compat | default). With AGENTTEAMS_LLM_PROVIDER=atlascloud and no AGENTTEAMS_OPENAI_BASE_URL (any controller deployment not fed by this chart's runtime-env Secret), the default branch registers a type=openai provider with no custom URL, so Higress routes requests to api.openai.com with the Atlas Cloud key while `agt llm-preflight` passes against api.atlascloud.ai. Consider adding a `case "atlascloud"` that applies defaultAtlasCloudBaseURL when OpenAIBaseURL is empty (or normalizing it at config load) so the preset does not depend solely on the Helm helper injecting the env var.

agentteams-controller/cmd/agt/llm_preflight.go
```diff
@@ -20,6 +20,7 @@ import (
 const (
 	defaultOpenAICompatibleBaseURL  = "https://api.openai.com/v1"
 	defaultQwenCompatibleBaseURL    = "https://dashscope.aliyuncs.com/compatible-mode/v1"
+	defaultAtlasCloudBaseURL        = "https://api.atlascloud.ai/v1"
```
oss-maintainer, line 23: The preset is currently Helm/k8s-only. Local and embedded installs have no atlascloud wiring: manager/scripts/init/setup-higress.sh:29-33 maps a default URL only for qwen, and its provider case at :188 handles only qwen/openai-compat/fallback — the fallback reads AGENTTEAMS_LLM_API_URL (not AGENTTEAMS_OPENAI_BASE_URL) and would register an openai-type provider with no custom URL, i.e. pointing at api.openai.com. install/agentteams-install.sh's provider menu likewise only offers openai-compat/qwen. If local support is intended, add an atlascloud case mirroring openai-compat (DNS service source + openaiCustomUrl); otherwise it is worth stating the scope. Also note the default URL is now hardcoded in two places (this Go const and the agentteams.llmBaseURL Helm helper).

helm/agentteams/templates/_helpers.tpl
```diff
@@ -80,6 +80,18 @@ Shared runtime Secret name.
 {{- printf "%s-runtime-env" (include "agentteams.fullname" .) }}
 {{- end }}
 
+{{/*
+Effective OpenAI-compatible LLM base URL. Explicit values always win; known
+providers can supply a usable endpoint when credentials.llmBaseUrl is omitted.
+*/}}
+{{- define "agentteams.llmBaseURL" -}}
```
oss-maintainer, line 87: The new preset is not documented anywhere users discover configuration: README.md:177, README.zh-CN.md:203, README.ja-JP.md:170 and docs/usage/deployment/kubernetes.md:124 (+ zh-cn counterpart) still enumerate credentials.llmProvider as only `openai-compat`/`qwen`, and README.md:143 says non-OpenAI providers must also set llmBaseUrl — no longer true for atlascloud, where the chart supplies the default. Adding the preset (and its implicit base URL) to those tables would make it discoverable.

tests/check-helm-agentteams.sh
```diff
@@ -19,4 +19,22 @@ helm template agentteams "${CHART}" "${COMMON_ARGS[@]}" > "${render}"
 grep -q 'name: agentteams-controller' "${render}"
 grep -q 'app.kubernetes.io/name: agentteams' "${render}"
 
+helm template agentteams "${CHART}" "${COMMON_ARGS[@]}" \
+    --set credentials.llmProvider=atlascloud \
+    --set credentials.defaultModel=deepseek-ai/deepseek-v4-pro > "${render}"
+
+if [ "$(grep -c 'AGENTTEAMS_OPENAI_BASE_URL: "https://api.atlascloud.ai/v1"' "${render}")" -ne 2 ]; then
```
oss-maintainer, line 26: The new assertions cover the atlascloud default and the explicit-override case, but not the backward-compat invariant: for other providers with llmBaseUrl empty, AGENTTEAMS_OPENAI_BASE_URL must remain absent from both Secrets. A regression in agentteams.llmBaseURL that emitted a default for every provider would still pass both new checks. Consider adding a zero-occurrence assertion for AGENTTEAMS_OPENAI_BASE_URL to the initial default render (COMMON_ARGS only).

oss-maintainer (commented):
﻿### Summary
Adds an `atlascloud` LLM provider preset to `agt llm-preflight` and the Helm chart (new `agentteams.llmBaseURL` helper supplies the default base URL when `credentials.llmBaseUrl` is omitted). The core change is correct: the Go case follows the existing provider pattern, helper precedence is right (explicit `llmBaseUrl` always wins), and the Helm render assertions match the two Secrets that emit `AGENTTEAMS_OPENAI_BASE_URL`. Main gap: the controller initializer's provider switch does not know about `atlascloud` yet, so the preset depends solely on the chart injecting the env var.

### Findings
- **[Warning]** `agentteams-controller/cmd/agt/llm_preflight.go:175` — controller initializer (`internal/initializer/initializer.go`) has no `atlascloud` case; a controller deployed with `AGENTTEAMS_LLM_PROVIDER=atlascloud` and no injected `AGENTTEAMS_OPENAI_BASE_URL` would route to api.openai.com while preflight passes.
- **[Info]** `agentteams-controller/cmd/agt/llm_preflight.go:23` — preset is Helm/k8s-only; local/embedded install paths are not wired, and the default URL now lives in two places (Go const + Helm helper).
- **[Info]** `helm/agentteams/templates/_helpers.tpl:87` — preset is not documented in READMEs / k8s deployment docs, which still enumerate only `openai-compat`/`qwen`.
- **[Info]** `tests/check-helm-agentteams.sh:26` — consider a negative assertion that `AGENTTEAMS_OPENAI_BASE_URL` stays absent for other providers when `llmBaseUrl` is empty.

---

PR #1100, opened 2026-07-28
fix(install): use legacy storage prefix for pre-v1.2 images

shiyiyue1102:
## Summary

- select the embedded controller storage prefix from the target image version
- keep `agentteams/agentteams-storage` for v1.2.0 and newer images
- use the legacy storage alias for pre-v1.2 images
- extend the existing installer compatibility test to cover storage-prefix selection

## Root cause

The installer already selected the legacy environment-variable prefix for pre-v1.2 images, but it always injected the current storage prefix. v1.1.2 OpenClaw and CoPaw workers configure the legacy MinIO alias, so they could not restore `agents/<worker>/` and exited after creation.

## Validation

- `bash -n install/agentteams-install.sh`
- `git diff --check`
- isolated v1.1.2 OpenClaw smoke: worker remained running and replied `OPENCLAW_OK` over Matrix
- isolated v1.1.2 CoPaw smoke: startup mirror completed, model preflight returned HTTP 200, worker reported ready, and replied `COPAW_OK` over Matrix

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -6,6 +6,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `hermes/`, `o
 
 **Bug Fixes**
 
+- **CoPaw Worker workspace projection**: Write Worker prompts, skills, tool configuration, and Matrix agent settings into CoPaw's default workspace so Team Leaders load their assigned role and join the Team Room. ([9074def](https://github.com/agentscope-ai/AgentTeams/commit/9074def3))
 - **Team Worker room boundary convergence**: Remove Manager again after standalone Worker infrastructure reconciliation restores regular Team Worker personal-room membership. ([b5b0add](https://github.com/agentscope-ai/AgentTeams/commit/b5b0add))
 - **Team Worker reference enforcement**: Keep referenced Worker CRs protected during direct deletion and reject Team API members whose required role is empty. ([d96f1ed](https://github.com/agentscope-ai/AgentTeams/commit/d96f1ed))
 - **Team Worker room membership**: Force Manager out of regular Team Worker personal rooms when equal Matrix power levels prevent a normal kick. ([43545c2](https://github.com/agentscope-ai/AgentTeams/commit/43545c2))
@@ -15,6 +16,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `hermes/`, `o
 
 **Branding and Compatibility**
 
+- **Pre-v1.2 installer storage prefix**: Select the legacy `hiclaw/agentteams-storage` prefix for v1.1.2 images while keeping the AgentTeams prefix for v1.2 and newer images. ([ed81b7b](https://github.com/agentscope-ai/AgentTeams/commit/ed81b7b9))
 - **Complete AgentTeams runtime rename**: Rename installer and Helm entrypoints, the controller Go module and CLI, and container filesystem paths to AgentTeams while preserving thin compatibility aliases and upgrade migration for existing HiClaw installations. ([3121f5f](https://github.com/agentscope-ai/AgentTeams/commit/3121f5f))
 - **Hard-cut AgentTeams naming**: Remove retired-brand installer wrappers, environment fallbacks, CLI aliases, Helm naming branches, runtime path migrations, and active source paths so fresh AgentTeams deployments use one canonical contract end to end. ([d20e606](https://github.com/agentscope-ai/AgentTeams/commit/d20e606617edefbbc42c28c1201c5629fa73fd88))
 - **Hard-cut Team and Worker resources**: Make Worker CRs the sole owners of runtime configuration and lifecycle, make Team CRs reference existing Workers through `spec.workerMembers`, and remove inline-member, registry, migration, and dependent-script compatibility paths. ([b3cf360](https://github.com/agentscope-ai/AgentTeams/commit/b3cf360))
--- a/copaw/src/copaw_worker/bridge.py
+++ b/copaw/src/copaw_worker/bridge.py
@@ -48,7 +48,9 @@ def _patch_copaw_paths(working_dir: Path) -> None:
         import copaw.constant as _const
         _const.WORKING_DIR = working_dir
         _const.SECRET_DIR = secret_dir
-        _const.ACTIVE_SKILLS_DIR = working_dir / "active_skills"
+        _const.ACTIVE_SKILLS_DIR = (
+            working_dir / "workspaces" / "default" / "skills"
+        )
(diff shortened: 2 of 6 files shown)


kailiangshang:
## Additional evidence: deterministic recursive workspace copies on the default v1.1.2 compatibility path

I independently reproduced this again on **2026-07-29** with the official
`agentteams-manager-copaw:v1.1.2` image. This is a high operational-severity
bug: the invalid storage alias makes the Manager mirror its workspace into a
local child of that same workspace. Repeated background syncs keep nesting the
copy and can consume the disk backing the Manager workspace.

This is especially urgent because the GitHub `latest` release is currently
`v1.1.2`, and current `main` also uses `v1.1.2` as its stable fallback. In other
words, the default compatibility path is the affected path.

### Root-cause chain

1. Current `main` (`8de237da736a542766e132836b29c0a2a9c48740`) correctly
   selects the legacy `HICLAW_` environment contract for pre-v1.2 images, but
   still injects
   `HICLAW_STORAGE_PREFIX=agentteams/agentteams-storage`:
   [installer line 3896](https://github.com/agentscope-ai/AgentTeams/blob/8de237da736a542766e132836b29c0a2a9c48740/install/agentteams-install.sh#L3896).
2. The `v1.1.2` Manager configures the remote `mc` alias as `hiclaw`, not
   `agentteams`:
   [v1.1.2 lines 171-180](https://github.com/agentscope-ai/AgentTeams/blob/a99457830fafb99c991bdb666aa8a1eef2f83b12/manager/scripts/init/start-manager-agent.sh#L171-L180).
3. It later runs `mc mirror /root/manager-workspace/
   "${HICLAW_STORAGE_PREFIX}/manager/"` initially and from a background sync
   loop:
   [v1.1.2 lines 1186-1204](https://github.com/agentscope-ai/AgentTeams/blob/a99457830fafb99c991bdb666aa8a1eef2f83b12/manager/scripts/init/start-manager-agent.sh#L1186-L1204).
4. Because `agentteams` is not a configured remote alias, `mc` treats
   `agentteams/agentteams-storage/manager/` as a local relative path. The
   destination is therefore inside its own source.
5. The sync commands suppress errors with `2>/dev/null || true`, so the Manager
   may appear healthy while recursive disk growth continues.

### Isolated reproduction

Environment:

```text
Linux 6.8.0-124-generic x86_64
Docker Client=29.2.1 Server=29.2.1
Image=agentteams-manager-copaw:v1.1.2
Digest=sha256:e8c0506500b26182140d7617c0a323c9bf63c8872246a2a30c0ac0a597ab9079
```

Isolation boundaries: `--network none`, `--cpus 0.5`, `--memory 256m`,
`--pids-limit 128`, a 32 MiB tmpfs Manager workspace, no ports, no persistent
AgentTeams volume, and no Matrix, MinIO, Dashboard, LLM, or API key.

With the value injected by current `main`:

```text
HICLAW_STORAGE_PREFIX=agentteams/agentteams-storage
AGENTTEAMS_ALIAS=MISSING
COMMAND=mc mirror /root/manager-workspace/ agentteams/agentteams-storage/manager/ --overwrite
PASS_1_NESTED_DIRS=1
PASS_2_NESTED_DIRS=4
PASS_3_NESTED_DIRS=7
PASS_4_NESTED_DIRS=10
NESTED_STORAGE_DIRS=10
RESULT=destination was treated as a local child of its own source
```

The generated path is visibly recursive:

```text
/root/manager-workspace/
└── agentteams/agentteams-storage/manager/
    └── agentteams/agentteams-storage/manager/
        └── agentteams/agentteams-storage/manager/...
```

### Fixed-prefix control

Using the same image and limits, but changing only the legacy storage prefix
and providing the alias expected by `v1.1.2`:

```text
HICLAW_STORAGE_PREFIX=hiclaw/agentteams-storage
MC_HOST_hiclaw=configured
PASS_1_RC=0 LOCAL_STORAGE_DIRS=0
PASS_2_RC=0 LOCAL_STORAGE_DIRS=0
PASS_3_RC=0 LOCAL_STORAGE_DIRS=0
PASS_4_RC=0 LOCAL_STORAGE_DIRS=0
RESULT=configured remote alias did not create a local destination
```

The filesystem assertion is the important control result: the correct alias
never creates a local destination under the Manager workspace.

### PR #1100 verification

I also checked out PR #1100 head
`ed81b7b940ee6b49e636f73694b3ba5b9bf035c4` and ran:

```text
$ bash tests/check-installer-pre-1.2-env-compat.sh
PASS: installer selects exactly one controller env contract by image version

$ bash tests/check-agentteams-rename-defaults.sh
PASS: active source tree uses AgentTeams-only names and contracts

$ bash -n install/agentteams-install.sh
# exit 0

$ git diff --check
# exit 0
```

The PR selects `hiclaw/agentteams-storage` for the pre-v1.2 contract and keeps
`agentteams/agentteams-storage` for v1.2+.

Given that `v1.1.2` is still the default latest release, I recommend merging
[#1100](https://github.com/agentscope-ai/AgentTeams/pull/1100) as soon as its
required CI/review gates permit. A follow-up defense-in-depth check should also
reject an unconfigured `mc` alias before any mirror command, rather than
allowing silent fallback to a local path.

shiyiyue1102:
Thanks for the detailed independent reproduction. This confirms both the legacy alias mismatch and the recursive local-mirror risk on the default v1.1.2 path. This PR keeps the compatibility fix version-scoped; I agree that rejecting an unconfigured mc alias would be a useful defense-in-depth follow-up.

kailiangshang:
> Thanks for the detailed independent reproduction. This confirms both the legacy alias mismatch and the recursive local-mirror risk on the default v1.1.2 path. This PR keeps the compatibility fix version-scoped; I agree that rejecting an unconfigured mc alias would be a useful defense-in-depth follow-up.

Thanks. While reproducing this bug, I noticed a related installation and
operability issue that I think would be valuable as a separate follow-up rather
than expanding the scope of #1100.

Would you be open to a feature proposal for a unified host runtime root, for
example:

```bash
AGENTTEAMS_ROOT_DIR=/srv/agentteams
```

The installer would derive component-specific directories such as:

- `/srv/agentteams/data`
- `/srv/agentteams/manager`
- `/srv/agentteams/dashboard`
- `/srv/agentteams/config`

Each component would still receive only its own bind mount; the entire root
would not be mounted into every container. The existing behavior would remain
unchanged when `AGENTTEAMS_ROOT_DIR` is not set.

The README would document the resulting layout, backup behavior, upgrade rules,
and uninstall behavior.

If this direction makes sense, I can open a dedicated feature issue first and
prepare a separate PR from `main` after #1100 is merged.

maplefeng-a (approved):
Looks good. CI is green, and the version-scoped legacy storage prefix plus CoPaw workspace projection changes look safe to merge.

PR #1104, opened 2026-07-29
fix(setup-higress): surface HTTP errors instead of silent WARNINGs

### LINKED ISSUE #908: K8s 模式下 Token Plan 配置无法热更新，Worker 报错 Model not exist || K8s mode, the Token Plan configuration cannot be hot updated, and the Worker reports an error Model not exist.
### Bug Description

在 K8s 部署模式下，通过一键脚本添加/更新了 Token Plan（例如新增 qwen3.7-max 等模型），但现有 Worker 无法调用新模型。Worker 日志报错 {"message":"Model not exist."} 以及 Model service check failed: HTTP 400。
What I expected:
更新 Token Plan 后，Higress 网关应自动同步 modelMapping 配置，且 Worker 应能识别新模型并成功路由到 DashScope 上游服务。

### Steps to Reproduce

在 K8s 环境下完成 HiClaw 部署。
运行配置脚本或通过 UI 更新 LLM Provider 配置（添加新模型 Token Plan）。
在 Worker (smart-editor/deepresearcher 等) 中尝试调用新模型。
观察 Worker 报错 Model not exist 且容器状态异常。

### AI Analysis

I Analysis:
Tool used: HiClaw Manager Agent (Built-in)
Analysis:

脚本空跑问题：/opt/hiclaw/scripts/init/setup-higress.sh 中检测到 HICLAW_RUNTIME=k8s 时直接跳过了 Higress Admin API 的路由与 Provider 更新逻辑。
网关配置缺失：由于上述跳过，Higress 的 qwen Provider 缺少针对新模型的 modelMapping，导致入口拦截请求。
容器未重载：Worker 容器未重启，openclaw.json 中的模型列表未更新，Preflight 检查失败。
网络解析问题：Worker 容器内缺失 aigw-local.hiclaw.io 的 hosts 解析，导致临时网络中断（Connection refused）。

### Relevant Logs

```shell
# Worker 报错日志
"Error: Error occurred during execution of model: unknown"
"Model service check failed: model chat completion preflight returned HTTP 400"

# 溯源发现 (setup-higress.sh)
if [ "${HICLAW_RUNTIME}" = "k8s" ]; then
    # 跳过所有 Higress 路由和 Provider 的 API 更新
    return 0
fi
```

### Component

CoPaw Worker

### Version / Commit

v1.1.2

---
### Bug Description

In K8s deployment mode, Token Plan is added/updated through one-click script (for example, new models such as qwen3.7-max are added), but the existing Worker cannot call the new model. The Worker log reports errors {"message":"Model not exist."} and Model service check failed: HTTP 400.
What I expected:
After updating the Token Plan, the Higress gateway should automatically synchronize the modelMapping configuration, and the Worker should be able to recognize the new model and successfully route to the DashScope upstream service.

### Steps to Reproduce

Complete HiClaw deployment in K8s environment.
Run the configuration script or update the LLM Provider configuration via the UI (add new model Token Plan).
Try calling the new model in a Worker (smart-editor/deepresearcher, etc.).
Observe that the Worker reports an error Model not exist and the container status is abnormal.

### AI Analysis

I Analysis:
Tool used: HiClaw Manager Agent (Built-in)
Analysis:

Script empty running problem: When HICLAW_RUNTIME=k8s is detected in /opt/hiclaw/scripts/init/setup-higress.sh, the routing and Provider update logic of the Higress Admin API are directly skipped.
Missing gateway configuration: Due to the above skip, Higress's qwen Provider lacks modelMapping for the new model, causing the entry to intercept the request.
Container not reloaded: Worker container not restarted, model list in openclaw.json not updated, Preflight check failed.
Network resolution problem: The hosts resolution of aigw-local.hiclaw.io is missing in the Worker container, resulting in temporary network interruption (Connection refused).

### Relevant Logs

```shell
# Worker error log
"Error: Error occurred during execution of model: unknown"
"Model service check failed: model chat completion preflight returned HTTP 400"

# Traceability discovery (setup-higress.sh)
if [ "${HICLAW_RUNTIME}" = "k8s" ]; then
    # Skip all Higress routing and Provider API updates
    return 0
fi
```

### Component

CoPaw Worker

### Version/Commit

In v1.1.2

RerankerGuo:
## Summary

Closes #908

The `higress_api` helper in `setup-higress.sh` used to swallow nearly every Console API failure as a log-level WARNING with exit 0:
- `"success":false` responses → only logged WARNING, caller never noticed
- HTTP 405 Method Not Allowed → fell through to generic WARNING
- HTML responses only matched `<!DOCTYPE html>` (not `<html>`)
- curl transport errors (`|| true`) → completely ignored
- Callers never checked the return value anyway

This produced the exact symptom reported in #908: after a K8s token-plan hot-reload, the every-boot IDEMPOTENT section would log "OK" but workers would later fail with "Model not exist / HTTP 400" because Higress had never accepted the new routes (HTTP 405 or a session-expired HTML page being silently dropped).

### manager/scripts/init/setup-higress.sh
1. **higress_api now returns real exit codes** — 0 only on success, non-zero on error (curl err, HTML response, 401/403, 405, `success:false`, or any unexpected HTTP code)
2. **Extended HTML detection** — case-insensitive match for both `<!DOCTYPE html>` and `<html...>` tags
3. **Truncated error bodies** — error messages keep the first 200-500 chars of response to avoid flooding logs, but keep enough context to diagnose
4. **Added `higress_api_soft` wrapper** — for NON-IDEMPOTENT first-boot steps where a prior boot may have already created the resource and an error is non-fatal
5. **NON-IDEMPOTENT section uses `higress_api_soft`** — preserves the old tolerant behaviour for one-time setup, since those resources are marker-protected and may already exist on a retry
6. **IDEMPOTENT section hard-fails on `higress_api` error** — every call appends `|| exit 1` so the AI Gateway / LLM provider hot-reload cannot silently produce a broken config

### changelog/current.md
Added bug-fix entry referencing #908.

## Test Plan
- [x] Shell syntax validation: `bash -n setup-higress.sh` passes
- [ ] CI run on GitHub
- [ ] Manually simulate bad Console endpoint (HTTP 405 proxy) — confirm script exits with ERROR, not a silent WARNING
- [ ] Manually re-run on healthy system — IDEMPOTENT provider update still succeeds after re-reading env vars

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -13,6 +13,7 @@ Record image-affecting changes to `manager/`, `worker/`, `copaw/`, `hermes/`, `o
 - **CoPaw Team assignment localparts**: Route Team Leader assignments that mention a Team Worker by Matrix localpart from Leader DM to Team Room. ([973e291](https://github.com/agentscope-ai/AgentTeams/commit/973e291))
 - **CoPaw Team coordination routing**: Route Team Leader worker assignments sent through the `message` tool from Leader DM to Team Room, matching the Matrix channel send path. ([92c8145](https://github.com/agentscope-ai/AgentTeams/commit/92c8145))
 - **Pinned OpenClaw source fetch**: Fetch the pinned OpenClaw commit directly so the base image build does not depend on a retired-brand external branch name. ([b0081c2](https://github.com/agentscope-ai/AgentTeams/commit/b0081c2))
+- **Token plan hot-reload error propagation**: Make `higress_api()` return non-zero exit codes on `"success":false`, HTTP 405, `<html>` responses (not just `<!DOCTYPE html>`), and curl transport errors; callers in the IDEMPOTENT AI Gateway/Provider section hard-fail with `|| exit 1` so misconfigured routes are never silently accepted. ([#908](https://github.com/agentscope-ai/AgentTeams/issues/908))
 
 **Branding and Compatibility**
 
--- a/manager/scripts/init/setup-higress.sh
+++ b/manager/scripts/init/setup-higress.sh
@@ -35,7 +35,10 @@ fi
 CONSOLE_URL="http://127.0.0.1:8001"
 
 # ============================================================
-# Helper: call Higress Console API, log result, never fail.
+# Helper: call Higress Console API, log result.
+# Returns 0 on success, non-zero on error.
+# Callers SHOULD check the exit code; on IDEMPOTENT (re-run on every boot)
+# steps, callers MUST treat non-zero exit as a blocking failure.
 # ============================================================
 higress_api() {
     local method="$1"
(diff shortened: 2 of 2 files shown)


RerankerGuo:
Closing after re-auditing current main. This branch targets the legacy Manager-side setup-higress path, while the supported embedded setup is now controller-owned. There is no current-main reproduction showing that this patch is still needed.

PR #1109, opened 2026-07-30
Update AgentLoop link in README.md

max-wc:
@shiyiyue1102 help to review


shiyiyue1102 (approved):
LGTM. This adds the intended README tracking parameter to the existing AgentLoop product link. The tracked URL is reachable and the translation check passes.

PR #1116, opened 2026-07-31
fix(controller): extend embedded worker token lifetime

NingHua-jzp:
## Summary

- use a ten-year service-account token lifetime for non-Kubernetes Workers
- configure the embedded kube-apiserver to accept the same maximum token lifetime
- add unit coverage for Worker TokenRequest expiration and the embedded API server flag

## Testing

- `go test ./internal/service ./internal/apiserver ./internal/backend`
- verified kube-apiserver v1.31.3 parses `--service-account-max-token-expiration=87600h0m0s`

### DIFF
--- a/agentteams-controller/internal/apiserver/embedded.go
+++ b/agentteams-controller/internal/apiserver/embedded.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/agentscope-ai/AgentTeams/agentteams-controller/internal/backend"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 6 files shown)


shiyiyue1102:
Thanks for the contribution and for exploring this direction. We are addressing the Docker Worker ServiceAccount token lifecycle in #1120 through short-lived token file rotation, rather than extending the token lifetime to ten years. To avoid maintaining two overlapping approaches, we will close this PR. Thanks again for the implementation and tests.

---

感谢你的贡献和对这一方案的探索。Docker Worker ServiceAccount token 的生命周期问题正在 #1120 中通过短期 token 文件轮转来处理，而不是将 token 有效期延长至十年。为避免同时维护两套重叠方案，我们将关闭这个 PR。再次感谢你提供的实现和测试。

PR #1120, opened 2026-07-31
fix(controller): rotate Docker Worker ServiceAccount token files

### LINKED ISSUE #1119: Docker Workers do not refresh short-lived ServiceAccount tokens
## Problem

Docker Workers receive their ServiceAccount token through the `AGENTTEAMS_AUTH_TOKEN` environment variable when the container is created. The token is static for the lifetime of the container and is not refreshed while the Worker is running or restarted.

With the default one-hour TokenRequest TTL, long-running Docker Workers eventually lose access to Controller APIs.

## Expected behavior

Docker Workers should:

- read the ServiceAccount token from `AGENTTEAMS_AUTH_TOKEN_FILE`;
- store it outside the MinIO-synchronized workspace;
- refresh it before expiration without recreating the Worker;
- atomically replace the token file;
- retain the existing token and retry after transient refresh failures;
- remove the credential storage when the Worker is deleted.

shiyiyue1102:
Closes #1119

## Summary

- Project Docker Worker ServiceAccount tokens into per-Worker named volumes.
- Configure Workers to read `/var/run/secrets/agentteams/token` instead of receiving the bearer token through container environment metadata.
- Refresh short-lived tokens before expiration without recreating running Workers.
- Atomically replace the token file through `token.next`.
- Keep the existing projection and retry after transient failures.
- Refresh stopped Workers immediately after they are started.
- Remove the credential volume when the Worker is deleted.

## Root cause

The Docker backend injected a token only when creating the container. Reconciliation returned immediately for running Workers and restarted stopped containers with their original environment, so no path could replace an expired token.

## Validation

- `go test ./... -count=1`
- Real Docker smoke test verified online token replacement, mode `0400`, and no remaining `token.next` file.


leroiu:
Thanks for this fix — token-file projection + refresh-before-expiry is exactly the right direction. One boundary case we hit on `v1.2.0-beta.1` that I'd like to confirm is covered (or planned): **controller restart**.

In our environment, restarting Docker Desktop (or the controller container) rotates the signing key used to issue the ServiceAccount tokens, which invalidates **all** worker tokens immediately — even well before the 1-hour TTL. Every worker then crash-loops with:

```
[AUTH] authentication failed: token not authenticated: [invalid bearer token, service account token has expired]
```

With the token projected to a file (`/var/run/secrets/agentteams/token`), a restart would still leave a stale signed token in that file unless the controller also re-projects/refreshes it after restart. Question: does the fix in #1120 (or the current main) handle the "controller restart rotates signing key" case, or is the refresh path only TTL-driven? If it's only TTL-driven, the restart case still needs an external recover step (we currently delete worker containers to force a rebuild with a fresh token).

Happy to file a follow-up issue if this isn't covered yet.

PR #1123, opened 2026-08-01
fix(cli): encode exposed Worker ports as integers

### LINKED ISSUE #1122: bug(cli): --expose serializes Worker ports as strings
## Bug Description

**What happened**

`agt create worker --expose 8080` serializes the exposed port as a JSON string:

```json
{"expose":[{"port":"8080"}]}
```

The Controller API contract defines `api/v1beta1.ExposePort.Port` as an `int`, so the request cannot be decoded. The same helper is used by the CLI's create, update, and apply Worker paths.

**What I expected**

The CLI should serialize the port as a JSON number (`{"port":8080}`) and reject invalid port values before sending the request.

## Steps to Reproduce

1. Start an AgentTeams Controller and configure `agt` to use it.
2. Run:
   ```shell
   agt create worker --name expose-test --model qwen3.6-plus --expose 8080 --no-wait
   ```
3. Observe that the request body contains `"port":"8080"` instead of `"port":8080`, causing JSON decoding to fail.

A minimal decoder reproducer reports:

```shell
json: cannot unmarshal string into Go struct field .port of type int
```

## AI Analysis

Tool used: OpenAI Codex

This is a deterministic CLI request-construction bug and does not involve Matrix messages, so Matrix debug logs are not applicable.

`parseExposePorts` splits the CSV input and stores each token unchanged in `map[string]interface{}`. Because the token is a Go string, `encoding/json` emits a JSON string. The receiving API types in both `api/v1beta1` and `internal/server` require an integer.

Affected commands:

- `agt create worker --expose ...`
- `agt update worker --expose ...`
- `agt apply worker --expose ...`

Suggested fix: parse each value as an integer, validate the TCP/UDP port range (`1-65535`), and add a regression test that decodes the generated JSON into a struct with an integer `port` field.

## Relevant Logs

```shell
json: cannot unmarshal string into Go struct field .port of type int
```

## Component

Other (Controller CLI)

## Version / Commit

`47c8f28`

jesseedcp:
Fixes #1122

## Summary

- Parse `--expose` CSV values into integer JSON ports.
- Reject empty, non-numeric, and out-of-range ports before an API request is sent.
- Reuse the same validated encoding in `create worker`, `update worker`, and `apply worker`.
- Add regression coverage and update the current changelog.

## Root cause

`parseExposePorts` previously stored each CSV token directly in `map[string]interface{}`. Since those tokens were Go strings, `encoding/json` produced payloads such as:

```json
{"expose":[{"port":"8080"}]}
```

The Controller's `ExposePort.Port` field is an integer, so otherwise valid `--expose` requests failed during JSON decoding.

## Validation

- `go test ./cmd/agt -count=1`
- `go vet ./cmd/agt`
- `git diff --check`
- Independent code review and verification: no blocking findings

I also ran `go test ./... -count=1`. All affected packages passed; the full suite still reports two pre-existing Windows baseline failures, reproduced on an untouched checkout of `47c8f28`:

- `internal/controller/TestCreateMemberContainerQwenPawWaitsForRuntimeConfig` times out.
- Two `internal/oss` tests cannot execute their temporary extensionless `mc` fixture on Windows.

### DIFF
--- a/agentteams-controller/cmd/agt/apply.go
+++ b/agentteams-controller/cmd/agt/apply.go
@@ -289,6 +289,14 @@ func applyWorkerParams(name, model, runtime, image, identity, soul, soulFile,
 			return err
 		}
 	}
+	var exposePorts []map[string]interface{}
+	if expose != "" {
+		var err error
+		exposePorts, err = parseExposePorts(expose)
+		if err != nil {
+			return err
+		}
+	}
 
 	client := NewAPIClient()
 
@@ -309,7 +317,7 @@ func applyWorkerParams(name, model, runtime, image, identity, soul, soulFile,
 		req["skills"] = splitCSV(skills)
 	}
 	if expose != "" {
-		req["expose"] = parseExposePorts(expose)
+		req["expose"] = exposePorts
 	}
 
 	var resp map[string]interface{}
--- a/agentteams-controller/cmd/agt/create.go
+++ b/agentteams-controller/cmd/agt/create.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -92,7 +93,11 @@ func createWorkerCmd() *cobra.Command {
 				req["skills"] = splitCSV(skills)
 			}
 			if expose != "" {
-				req["expose"] = parseExposePorts(expose)
+				ports, err := parseExposePorts(expose)
+				if err != nil {
+					return err
+				}
+				req["expose"] = ports
 			}
 
 			client := NewAPIClient()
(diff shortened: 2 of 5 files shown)


shiyiyue1102:
The numeric port encoding and shared create/update/apply handling look correct. One validation edge case remains: `splitCSV` silently drops empty entries, so values such as `8080,`, `,8080`, or `8080,,3000` are accepted even though the PR states that empty values are rejected. Could you reject empty CSV segments explicitly and add regression cases for these inputs?

---

数值端口编码以及 create/update/apply 的统一处理是正确的。目前还剩一个校验边界：`splitCSV` 会静默丢弃空项，因此 `8080,`、`,8080` 或 `8080,,3000` 仍会被接受，与 PR 所述“拒绝空值”不一致。建议显式拒绝空的 CSV 项，并补充这些回归用例。

jesseedcp:
Thanks for catching this. Fixed in `97b9433`.

I updated `parseExposePorts` to split the raw `--expose` input before trimming, so empty CSV segments such as `8080,`, `,8080`, and `8080,,3000` are now rejected instead of being silently dropped.

I also added duplicate-port validation after integer parsing, so inputs such as `8080,8080` and `08080,8080` are rejected before they can produce duplicate downstream Service or gateway entries.

Regression coverage now includes empty segments, duplicate ports, whitespace, invalid values, and the valid boundary values `1` and `65535`.

Validation:
- `go test ./cmd/agt -run TestParseExposePorts -count=1`
- `go test ./cmd/agt -count=1`
- `go vet ./cmd/agt`

shiyiyue1102 (approved):
LGTM. The numeric port encoding is correct across create, update, and apply; malformed, duplicate, and boundary port cases are covered by focused tests. `go test ./cmd/agt -count=1` and `go vet ./cmd/agt` pass, and the current head merges cleanly with the latest main.

PR #1127, opened 2026-08-02
fix(copaw): honor model-switch --no-reasoning via enable_thinking=false

XbaoWu:
Fixes #1126

## Summary

- CoPaw Manager `model-switch` ignored `--no-reasoning` (only logged `reasoning=false`) while OpenClaw already persisted `reasoning: false`.
- Persist **model-level** `generate_kwargs.extra_body.enable_thinking=false` on the modern CoPaw provider store (`providers/custom/agentteams-gateway.json` + `active_model.json`).
- Sync `openclaw.json` `reasoning` / primary as bridge SoT so CoPaw restart / re-bridge does not wipe the setting from an empty local write.
- Bridge maps `openclaw.json` `reasoning: false` → model-level CoPaw `generate_kwargs` only (no shared provider-level flag).

## Test plan

- [x] jq / script path: `--no-reasoning` writes model-level `enable_thinking=false`; omit flag clears it
- [x] Local Docker CoPaw Manager (`docker cp` this script): `update-manager-model.sh <model> --no-reasoning` → provider has `enable_thinking=false`, `openclaw.json` has `reasoning=false` (checked immediately after write)
- [x] Clear path (no `--no-reasoning`) restores both sides
- [x] Bridge unit tests added for `reasoning: false` → model-level `generate_kwargs` (and sibling does not force provider-level)
- [x] Restart CoPaw / Manager and confirm chat TTFT drops (no long thinking)

## Notes

- Modern CoPaw provider layout only.
- MinIO Remote→Local overwriting a local `openclaw.json` write is a **separate** persistence issue and is **not** fixed in this PR.

PR #1129, opened 2026-08-02
fix(controller): preserve model-switch openclaw reasoning across Manager reconcile

### LINKED ISSUE #1126: CoPaw: model-switch `--no-reasoning` is a no-op (unlike OpenClaw)
## Bug Description

Environment (where reproduced):
- Install: AgentTeams embedded / Docker (local)
- Version: v1.2.0-class (`2ea0274`)
- `AGENTTEAMS_MANAGER_RUNTIME=copaw`
- `AGENTTEAMS_LLM_PROVIDER=qwen` (DashScope / Bailian via AI Gateway)
- `AGENTTEAMS_DEFAULT_MODEL=qwen3.7-plus`
- OS: macOS (Docker Desktop)

What happened:
When Manager runtime is CoPaw, running `update-manager-model.sh <model> --no-reasoning` logs `reasoning=false` but does **not** persist any CoPaw-side setting that disables thinking. OpenClaw correctly sets `reasoning: false` on the model entry; the CoPaw branch ignores `$REASONING` and only updates model id / context window.

For DashScope Qwen hybrid models (`qwen3.x-plus`), thinking stays enabled on the real chat path, adding multi-second latency per turn. Health preflight already sends `enable_thinking=false` (`copaw/src/copaw_worker/health.py`), but normal chat does not.

What I expected:
`--no-reasoning` should persist something CoPaw chat honors, equivalent to OpenClaw's `reasoning: false` — e.g. `generate_kwargs.extra_body.enable_thinking=false` on the provider / active model entry (modern: `providers/custom/*.json` + `active_model.json`; legacy: `providers.json`).

## Steps to Reproduce

1. Install AgentTeams with `AGENTTEAMS_MANAGER_RUNTIME=copaw` and a Qwen hybrid model (e.g. `qwen3.7-plus`).
2. Inside the Manager container:

```bash
bash /opt/agentteams/agent/skills/model-switch/scripts/update-manager-model.sh qwen3.7-plus --no-reasoning
```

3. Inspect the CoPaw provider store:

```bash
jq '.generate_kwargs, [.extra_models[]? | select(.id|test("qwen")) | {id, generate_kwargs}]' \
  ~/.copaw.secret/providers/custom/agentteams-gateway.json
```

4. Observe: script logs `reasoning=false`, but provider JSON has no `enable_thinking=false` (and chat still emits thinking / high TTFT).

## AI Analysis

Tool used: Cursor

Analysis:
Root cause is in `manager/agent/skills/model-switch/scripts/update-manager-model.sh`:
- CLI parses `--no-reasoning` → `REASONING=false` and logs it.
- OpenClaw path writes `.models.providers["agentteams-gateway"].models[].reasoning = $reasoning`.
- CoPaw path only updates model id / `active_llm` / context window and never uses `$REASONING`.
- Modern CoPaw layout is `~/.copaw.secret/providers/custom/agentteams-gateway.json` (+ `active_model.json`), while the script originally targeted legacy `~/.copaw.secret/providers.json`.

Additionally, Worker bridge (`copaw_worker/bridge.py`) did not map openclaw.json `reasoning: false` into CoPaw `generate_kwargs`, so Workers stay inconsistent even when Controller marks the model non-reasoning.

## Relevant Logs

```text
[agentteams] Updating Manager model: qwen3.7-plus (ctx=200000, max=64000, reasoning=false, input=["text", "image"])
[agentteams] Testing model reachability: .../v1/chat/completions (model=qwen3.7-plus)...
[agentteams] Model test passed (HTTP 200)
[agentteams] Done. CoPaw model is now: qwen3.7-plus (ctx=200000)

# provider store after --no-reasoning (pre-fix): no enable_thinking
generate_kwargs: {}
qwen3.7-plus.generate_kwargs: {}
```

Gateway latency check (same model, request-level flag):

```text
enable_thinking=false  ~1.0s  reasoning_content=no
enable_thinking=true   ~4.5s  reasoning_content=yes
```

## Component

Manager Agent

(also affects CoPaw Worker via missing bridge mapping of `reasoning: false`)

## Version / Commit

`2ea0274` (AgentTeams v1.2.0-class install; Manager runtime=copaw; model=`qwen3.7-plus`)

### LINKED ISSUE #1128: model-switch openclaw.json `reasoning` undone by Manager reconcile / agents/manager sync
### Bug Description

Environment (where reproduced):
- Install: AgentTeams embedded / Docker (local), `AGENTTEAMS_RUNTIME=k8s`
- Manager runtime: CoPaw
- Host workspace bind-mounted into Manager **and** Controller (`agents/manager`)
- OS: macOS (Docker Desktop)

What happened:
After `model-switch` (or a direct edit) sets `openclaw.json` model `reasoning: false`, within ~1–4 seconds the value flips back to `true`.

What I expected:
Hot-updates to manager `openclaw.json` (especially `--no-reasoning` / bridge SoT) should survive Controller Manager reconcile and MinIO Remote→Local sync.

## Root cause

Two writers conflict:

1. **Controller `DeployManagerConfig` → `PutManagerConfig`** regenerates `openclaw.json` from the Manager CR (default `reasoning: true` for custom models), merges only `groupAllowFrom` + plugins, then:
   - `PutObject` to MinIO `agents/manager/openclaw.json`
   - `writeManagerLocalConfig` overwrites the **shared live workspace** bind-mount
2. Controller also mirrors MinIO `agents/` onto `/root/agentteams-fs/agents/`, which is the same host dir as Manager’s workspace — a stale `agents/manager/openclaw.json` in MinIO will pull over the live file.

Related: CoPaw `--no-reasoning` no-op is #1126. This issue is persistence after a correct local write.

### Steps to Reproduce

1. Embedded install with Manager workspace on host (default `~/agentteams-manager`).
2. Set reasoning false on the active model in `openclaw.json` (via model-switch or jq).
3. Wait a few seconds (Manager reconcile).
4. Observe `reasoning` flip back to `true` on disk / in MinIO `agents/manager/openclaw.json`.

### AI Analysis

Tool used: Cursor

Analysis:
- `manager_config.go` `PutManagerConfig` did not preserve per-model `reasoning` and preferred OSS over the live workspace file.
- Manager `mc cp` to `agents/manager/` often 403s (no ACL); pushing only `manager/openclaw.json` is insufficient because Controller regenerates `agents/manager/`.
- Fix: load existing config from the shared local mount first; `mergeModelReasoning`; model-switch best-effort MinIO push.

### Relevant Logs

```shell
Shared mount (Manager workspace == Controller `agents/manager`):


# manager
/Users/.../agentteams-manager -> /root/manager-workspace

# controller
/Users/.../agentteams-manager -> /root/agentteams-fs/agents/manager


After writing `reasoning=false` locally and pushing `manager/openclaw.json`, 100ms poll of local vs two MinIO keys:


t0     local=false  manager/=false  agents/manager/=false
t=3.23 local=false  manager/=false  agents/manager/=true   ← Controller PutObject
t=3.49 local=true   manager/=false  agents/manager/=true   ← live workspace overwritten


Only field that changes in the flip (diff of openclaw.json):


-            "reasoning": false
+            "reasoning": true


Manager cannot update the Controller SoT path (explains why skill-only MinIO push is not enough):


Pushed openclaw.json to agentteams/agentteams-storage/manager/openclaw.json
WARN: failed to push openclaw.json to agentteams/agentteams-storage/agents/manager/openclaw.json
# mc: Insufficient permissions to access this path .../agents/manager/openclaw.json


Flip still happens after killing Manager’s own `mc` sync loops — so the writer is Controller reconcile / shared-mount write, not Manager `start-manager-agent` MinIO→Local alone.
```

### Component

Controller (Manager config) / Manager Agent (model-switch)

### Version / Commit

v1.2.0-class embedded; found while validating #1126 / #1127

XbaoWu:
Fixes #1128

Stacked on #1127 (base: `fix/copaw-model-switch-no-reasoning`). Related write-path fix: #1126 / #1127.

## Summary

- After `model-switch` (or a direct edit) set `openclaw.json` model `reasoning: false`, Controller Manager reconcile regenerated config with default `reasoning: true` and overwrote the shared live workspace / MinIO `agents/manager/openclaw.json` within ~1–4s.
- `PutManagerConfig` / `UpdateManagerGroupAllowFrom` now load existing openclaw **local-first** (live Manager workspace mount), then OSS.
- New `mergeModelReasoning` preserves per-model `reasoning` from the existing config across regenerate (live/hot-update wins over CR / `AGENTTEAMS_MODEL_REASONING` defaults until the workspace is updated again).
- `model-switch` best-effort pushes `openclaw.json` to MinIO `manager/` + `agents/manager/` (the latter often 403 for Manager credentials; WARN-only — persistence relies on Controller merge from live workspace).

## Test plan

- [x] Controller unit tests (Docker `golang:1.25-bookworm`): `TestPutManagerConfig_PreservesModelReasoning`, `TestPutManagerConfig_PrefersLocalReasoningOverStaleOSS`, `TestUpdateManagerGroupAllowFrom_PrefersLocalConfig` (+ existing PutManagerConfig cases)
- [ ] Rebuild / redeploy Controller with this change (script-only Manager update is not enough)

## Notes

- Precedence: live workspace `reasoning` wins over the next Controller regenerate. To flip via CR/env alone, re-run `model-switch` (or edit live openclaw) so the workspace matches.
- Does **not** grant Manager write ACL on `agents/manager/`; that remains Controller SoT.

### DIFF
--- a/agentteams-controller/internal/service/manager_config.go
+++ b/agentteams-controller/internal/service/manager_config.go
@@ -90,6 +90,18 @@ func (l *ManagerConfigStore) writeManagerLocalConfig(data []byte) {
 // PutManagerConfig writes the Manager's openclaw.json to OSS, merging the
 // new config with any existing groupAllowFrom entries to avoid overwriting
 // additions made by UpdateManagerGroupAllowFrom (e.g. team leader IDs).
+//
+// Existing config is loaded from the shared local mount first (live Manager
+// workspace), then OSS. Hot-updates such as model-switch `--no-reasoning`
+// therefore win over the next Manager reconcile regenerate — but only when
+// AgentFSDir is set (embedded / shared-mount). In k8s without that mount
+// the load falls through to OSS; if Manager could not write
+// agents/manager/openclaw.json (often 403), OSS may still hold stale
+// reasoning and the overwrite from issue #1128 can persist.
+//
+// To apply a new AGENTTEAMS_MODEL_REASONING / CR default, update the live
+// openclaw.json (e.g. re-run model-switch) so the workspace matches the
+// desired flag.
 func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
 	if !l.Enabled() {
 		return nil
@@ -102,14 +114,17 @@ func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
 	key := l.managerAgentPrefix() + "/openclaw.json"
 
 	// Read existing config to preserve user customizations on top of the
-	// generated defaults: groupAllowFrom additions plus user-modified plugin
-	// entries (e.g. memory-core dreaming schedule, extra load paths).
-	existingData, err := l.OSS.GetObject(ctx, key)
-	if err == nil && len(existingData) > 0 {
+	// generated defaults: groupAllowFrom additions, per-model reasoning
+	// (model-switch), plus user-modified plugin entries.
+	// Load errors are already logged. Continue with generated config when
+	// existing data is missing so reconcile still writes a usable file.
+	existingData, _ := l.loadExistingManagerConfig(ctx, key)
+	if len(existingData) > 0 {
 		var existingCfg map[string]interface{}
 		var newCfg map[string]interface{}
 		if json.Unmarshal(existingData, &existingCfg) == nil && json.Unmarshal(configJSON, &newCfg) == nil {
 			mergeGroupAllowFrom(existingCfg, newCfg)
+			mergeModelReasoning(existingCfg, newCfg)
 			if merged, mErr := json.MarshalIndent(newCfg, "", "  "); mErr == nil {
 				configJSON = merged
 			}
@@ -126,6 +141,128 @@ func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
 	}
 	l.writeManagerLocalConfig(configJSON)
 	return nil
+}
+
+// loadExistingManagerConfig prefers the shared local Manager workspace (embedded
+// bind-mount), then falls back to OSS. Local-first matters because model-switch
+// updates the workspace immediately while Manager may lack ACL to push
+// agents/manager/ in MinIO.
+//
+// Local bytes are used only when they unmarshal as a JSON object. A
+// corrupt/partial live file falls through to OSS. OSS not-found is not an
+// error; any other OSS read failure is logged and returned so callers such as
+// UpdateManagerGroupAllowFrom can surface it instead of silently dropping
+// the mutation.
+func (l *ManagerConfigStore) loadExistingManagerConfig(ctx context.Context, key string) ([]byte, error) {
+	logger := log.FromContext(ctx).WithName("manager-config")
+
+	if path := l.managerLocalConfigPath(); path != "" {
+		data, err := os.ReadFile(path)
+		switch {
+		case err == nil && len(data) > 0:
+			var probe map[string]interface{}
+			if json.Unmarshal(data, &probe) == nil {
+				return data, nil
+			}
+			logger.Info("local manager openclaw.json is not valid JSON; falling back to OSS", "path", path)
+		case err == nil:
+			// empty file — treat as missing
+		case os.IsNotExist(err):
+			// first write or no live workspace yet
+		default:
+			logger.Error(err, "failed to read local manager openclaw.json; falling back to OSS", "path", path)
+		}
+	}
+
+	data, err := l.OSS.GetObject(ctx, key)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		logger.Error(err, "failed to read manager openclaw.json from OSS", "key", key)
+		return nil, fmt.Errorf("read manager config from OSS: %w", err)
+	}
+	if len(data) == 0 {
+		return nil, nil
+	}
+	return data, nil
+}
+
+// mergeModelReasoning copies per-model reasoning flags from old into new so a
+// Controller regenerate does not undo model-switch --no-reasoning (or similar
+// hot-updates) that already landed on the live openclaw.json.
+//
+// Policy: existing/live reasoning always wins over generated defaults when the
+// model id is present in both. Intentional CR/env flips must update the live
+// workspace (or OSS if local is absent), not rely on regenerate alone.
+func mergeModelReasoning(oldCfg, newCfg map[string]interface{}) {
+	oldByID := modelEntriesByID(oldCfg)
+	newEntries := modelEntries(newCfg)
+	if len(oldByID) == 0 || len(newEntries) == 0 {
+		return
+	}
+	for _, m := range newEntries {
+		id, _ := m["id"].(string)
+		if id == "" {
+			continue
+		}
+		old, ok := oldByID[id]
+		if !ok {
+			continue
+		}
+		if reasoning, has := old["reasoning"]; has {
+			m["reasoning"] = reasoning
+		}
+	}
+}
+
+func modelEntries(cfg map[string]interface{}) []map[string]interface{} {
+	models, _ := cfg["models"].(map[string]interface{})
+	if models == nil {
+		return nil
+	}
+	providers, _ := models["providers"].(map[string]interface{})
+	if providers == nil {
+		return nil
+	}
+	// Prefer the AgentTeams gateway provider; otherwise scan all providers.
+	order := make([]string, 0, len(providers)+1)
+	if _, ok := providers["agentteams-gateway"]; ok {
+		order = append(order, "agentteams-gateway")
+	}
+	for name := range providers {
+		if name != "agentteams-gateway" {
+			order = append(order, name)
+		}
+	}
+	out := make([]map[string]interface{}, 0)
+	for _, name := range order {
+		provider, _ := providers[name].(map[string]interface{})
+		if provider == nil {
+			continue
+		}
+		raw, _ := provider["models"].([]interface{})
+		for _, item := range raw {
+			if m, ok := item.(map[string]interface{}); ok {
+				out = append(out, m)
+			}
+		}
+	}
+	return out
+}
+
+func modelEntriesByID(cfg map[string]interface{}) map[string]map[string]interface{} {
+	entries := modelEntries(cfg)
+	if len(entries) == 0 {
+		return nil
+	}
+	byID := make(map[string]map[string]interface{}, len(entries))
+	for _, m := range entries {
+		if id, ok := m["id"].(string); ok && id != "" {
+			byID[id] = m
+		}
+	}
+	return byID
 }
 
 // mergeGroupAllowFrom copies any extra groupAllowFrom entries from old config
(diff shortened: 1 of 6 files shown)


agentteams-controller/internal/service/manager_config.go
```diff
@@ -128,6 +135,100 @@ func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
...
+// bind-mount), then falls back to OSS. Local-first matters because model-switch
+// updates the workspace immediately while Manager may lack ACL to push
+// agents/manager/ in MinIO.
+func (l *ManagerConfigStore) loadExistingManagerConfig(ctx context.Context, key string) []byte {
+	if path := l.managerLocalConfigPath(); path != "" {
+		if data, err := os.ReadFile(path); err == nil && len(data) > 0 {
+			return data
+		}
+	}
+	data, err := l.OSS.GetObject(ctx, key)
```
oss-maintainer, line 177: `loadExistingManagerConfig` silently swallows all read errors (local and OSS) without logging. For `PutManagerConfig` this matches the old OSS-only behavior, but for `UpdateManagerGroupAllowFrom` it is a behavior change: the old code returned `fmt.Errorf("read manager config from OSS: %w", err)` on non-NotExist failures, which callers logged as a non-fatal warning. Now a transient OSS error silently drops the update with no log line, reducing observability. Suggestion: log a warning when the OSS fallback read fails (distinguishable from not-found), so operators can spot when groupAllowFrom mutations are silently skipped.

agentteams-controller/internal/service/manager_config.go
```diff
@@ -128,6 +135,100 @@ func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
 	return nil
 }
 
+// loadExistingManagerConfig prefers the shared local Manager workspace (embedded
+// bind-mount), then falls back to OSS. Local-first matters because model-switch
+// updates the workspace immediately while Manager may lack ACL to push
+// agents/manager/ in MinIO.
+func (l *ManagerConfigStore) loadExistingManagerConfig(ctx context.Context, key string) []byte {
+	if path := l.managerLocalConfigPath(); path != "" {
```
oss-maintainer, line 159: Local file bytes are returned without JSON validation. If the live workspace has a partial/corrupt write (e.g., model-switch interrupted mid-`mv`), `loadExistingManagerConfig` returns malformed bytes. In `PutManagerConfig`, the subsequent `json.Unmarshal(existingData, &existingCfg)` then fails, which causes the entire merge block to be skipped — losing both groupAllowFrom and reasoning preservation, even if OSS has a valid config. For `UpdateManagerGroupAllowFrom`, a corrupt local file produces a parse error that silently drops the update (callers log and move on). Suggestion: attempt `json.Unmarshal` on local data; on parse failure, log a warning and fall through to the OSS read.

agentteams-controller/internal/service/manager_config.go
```diff
@@ -102,14 +108,15 @@ func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
 	key := l.managerAgentPrefix() + "/openclaw.json"
 
 	// Read existing config to preserve user customizations on top of the
-	// generated defaults: groupAllowFrom additions plus user-modified plugin
-	// entries (e.g. memory-core dreaming schedule, extra load paths).
-	existingData, err := l.OSS.GetObject(ctx, key)
-	if err == nil && len(existingData) > 0 {
+	// generated defaults: groupAllowFrom additions, per-model reasoning
+	// (model-switch), plus user-modified plugin entries.
```
oss-maintainer, line 118: In k8s deployments where the controller does not share a local mount with the Manager (`AgentFSDir` empty), `loadExistingManagerConfig` falls through to OSS. The model-switch script acknowledges its push to `${AGENTTEAMS_STORAGE_PREFIX}/agents/manager/openclaw.json` often fails with 403 (Manager credentials lack write ACL). In that scenario OSS still holds the stale `reasoning: true` value, and the overwrite loop from #1128 persists unchanged. The doc comment on `PutManagerConfig` (lines 94-98) says "live workspace wins" but does not document that this protection is only effective when `AgentFSDir` is set. Suggestion: document the k8s/no-mount limitation in the function comment, and consider whether the model-switch OSS push needs to be elevated from best-effort for the fix to work across all deployment modes.

agentteams-controller/internal/service/manager_config_test.go
```diff
@@ -3,11 +3,185 @@ package service
...
+		AgentFSDir:   dir,
+	})
+	if err := lc.UpdateManagerGroupAllowFrom("@worker1:agentteams.local", true); err != nil {
+		t.Fatalf("UpdateManagerGroupAllowFrom: %v", err)
+	}
+
+	out, err := fake.GetObject(ctx, "agents/manager/openclaw.json")
+	if err != nil {
+		t.Fatalf("GetObject: %v", err)
+	}
```
oss-maintainer, line 185: Test coverage gaps for the new local-first load path: (a) no test for malformed local file — does the function fall through to OSS or return bad bytes? (b) no test for `UpdateManagerGroupAllowFrom` with OSS-only (no local file), which is the original production path that was just changed; (c) no test where model IDs differ between old and new (e.g., model-switch added a new model that the CR doesn't list). Adding at least (a) and (b) would catch regressions in the fallback behavior.

agentteams-controller/internal/service/manager_config.go
```diff
@@ -128,6 +135,100 @@ func (l *ManagerConfigStore) PutManagerConfig(configJSON []byte) error {
...
+	}
+	for _, m := range newEntries {
+		id, _ := m["id"].(string)
+		if id == "" {
+			continue
+		}
+		old, ok := oldByID[id]
+		if !ok {
+			continue
+		}
```
oss-maintainer, line 212: `mergeModelReasoning` copies the `reasoning` value from the old config regardless of its type. If the live workspace has `"reasoning": "false"` (a string, e.g. from a buggy jq expression), the string is propagated verbatim into the generated config. The CoPaw bridge checks `model_cfg.get("reasoning") is False` (identity against bool), so a string value would silently be treated as truthy (reasoning enabled). Consider adding a type guard: `if reasoning, has := old["reasoning"]; has { if _, isBool := reasoning.(bool); isBool { m["reasoning"] = reasoning } }`.

oss-maintainer (commented):
### Summary
Fixes the #1128 reasoning-overwrite loop by loading the existing Manager openclaw config local-first (live workspace mount, then OSS) and preserving per-model `reasoning` across regenerate via `mergeModelReasoning`. The core logic is sound and the fix works for embedded/local deployments where the controller shares the Manager workspace mount; the three new tests cover the happy paths well. Note: the copaw/model-switch portions overlap with the already-approved #1127 and were not re-reviewed here. Before merging, the k8s deployment-mode gap and the silent error swallowing below should be addressed or explicitly documented.

### Findings
- **[Warning]** `manager_config.go:148` — `loadExistingManagerConfig` swallows read errors without logging; for `UpdateManagerGroupAllowFrom` this is an observability regression vs the previous code, which surfaced non-NotExist OSS errors as logged warnings.
- **[Warning]** `manager_config.go:143` — local file bytes are returned without JSON validation; a corrupt live workspace file (e.g. interrupted write) can skip the whole merge block in `PutManagerConfig` or silently drop `UpdateManagerGroupAllowFrom` updates. Suggest validating local JSON and falling through to OSS on parse failure.
- **[Warning]** `manager_config.go:112` — in k8s deployments without a shared mount (`AgentFSDir` empty), the load falls through to OSS, which still holds stale `reasoning: true` (the model-switch OSS push is best-effort and often 403), so the #1128 overwrite loop persists there. Please document this limitation in the function comment and consider whether the OSS push needs to be more than best-effort.
- **[Info]** `manager_config_test.go:171` — coverage gaps: malformed local file, OSS-only path for `UpdateManagerGroupAllowFrom`, and model-ID set differences between old/new configs.
- **[Info]** `manager_config.go:176` — `mergeModelReasoning` copies `reasoning` regardless of type; a string `"false"` would be treated as truthy downstream (CoPaw bridge checks `is False`). Consider a bool type guard.

### Suggestions
1. Log a warning on OSS fallback read failures (distinct from not-found) so operators can see when groupAllowFrom mutations are skipped.
2. `json.Unmarshal` the local bytes before trusting them; on parse failure fall back to OSS with a logged warning.
3. Document the k8s/no-mount limitation on `PutManagerConfig` (the "live workspace wins" guarantee only holds when `AgentFSDir` is set).
4. Add the fallback-path tests listed above.

---

PR #1134, opened 2026-08-03
fix: enforce LF line endings for shell scripts

### LINKED ISSUE #1133: Shell scripts are checked out with CRLF on Windows and fail under WSL
### What happened

A Windows clone made with Git for Windows' common default `core.autocrlf=true` converts all tracked `*.sh` files to CRLF because the repository has no `.gitattributes`. Invoking those scripts through WSL from a mounted Windows path fails before any AgentTeams logic runs.

```text
install/agentteams-install.sh: line 64: syntax error near unexpected token `$'{\r''`
install/agentteams-apply.sh: line 18: syntax error near unexpected token `$'{\r''`
```

### Expected behavior

Shell scripts should remain LF in Windows checkouts and run under WSL or Git Bash without requiring every user to change their global Git configuration.

## Steps to Reproduce

1. On Windows, confirm Git for Windows uses `core.autocrlf=true`.
2. Clone the repository.
3. Run:
   ```powershell
   git ls-files --eol -- install/agentteams-install.sh install/agentteams-apply.sh
   ```
4. Observe `i/lf w/crlf` for both scripts.
5. Run:
   ```powershell
   wsl bash -n /mnt/<drive>/<path>/AgentTeams/install/agentteams-install.sh
   ```
6. Observe the parser error at the first function declaration.

## AI Analysis

Tool used: OpenAI Codex

The repository stores the scripts as LF in the Git index, but it does not define an EOL policy. With `core.autocrlf=true`, Git for Windows therefore writes `*.sh` files as CRLF. WSL executes the files directly from the mounted Windows working tree and Bash treats the remaining carriage return as syntax.

A minimal repository-level fix is:

```gitattributes
*.sh text eol=lf
```

Validation against commit `124f06d`:

- All 141 tracked Shell scripts reported `w/lf`.
- `bash -n` passed for all 141 scripts.
- `install/agentteams-install.sh` and `install/agentteams-apply.sh` both returned syntax-check exit code 0.
- The patch changes only a root `.gitattributes` file; it does not create content changes for the 141 scripts.

## Environment

- Windows 11
- Git for Windows 2.40.0 with `core.autocrlf=true`
- WSL 2.7.11
- Ubuntu 26.04 LTS
- AgentTeams commit: `124f06d`

YangMingSJTU:
## Summary

- add a root `.gitattributes` rule for tracked Shell scripts
- keep `*.sh` files on LF when cloned with Git for Windows and `core.autocrlf=true`
- prevent WSL and Git Bash parser failures caused by CRLF working-tree files

Closes #1133

## Validation

- confirmed all 141 tracked `*.sh` files report `w/lf`
- ran `bash -n` successfully against all 141 tracked Shell scripts
- verified `install/agentteams-install.sh` and `install/agentteams-apply.sh` individually return syntax-check exit code 0
- confirmed the staged change contains only `.gitattributes`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Shell scripts must remain executable by Unix shells after Windows checkouts.
+*.sh text eol=lf


shiyiyue1102 (approved):
LGTM. The repository-level `*.sh text eol=lf` rule is a minimal and effective fix. I verified that it overrides `core.autocrlf=true`, keeps the affected scripts on LF, and applies cleanly to all 145 tracked shell scripts after merging with the latest `main`.

---

LGTM。仓库级的 `*.sh text eol=lf` 是一个最小且有效的修复。我验证了它可以覆盖 `core.autocrlf=true`、确保相关脚本保持 LF，并且与最新 `main` 合并后能正确覆盖全部 145 个受跟踪的 shell 脚本。

PR #1136, opened 2026-08-04
fix: reject unsupported declarative apply flags

### LINKED ISSUE #1135: agentteams-apply.sh advertises unsupported --prune, --dry-run, and --watch flags
## Description

`install/agentteams-apply.sh` documents `--prune`, `--dry-run`, and `--watch` as supported usage examples. The wrapper forwards those arguments to `agt apply`, but the current CLI only registers `-f/--file` and the `worker` subcommand.

As a result, the documented commands perform preliminary container checks and may copy the YAML file before `agt apply` rejects the unknown flag.

## Reproduction

1. Start the AgentTeams Manager container.
2. Create a valid `resource.yaml`.
3. Run:

```bash
./install/agentteams-apply.sh -f resource.yaml --dry-run
```

The same mismatch applies to `--prune` and `--watch`.

## Expected behavior

Either the advertised behavior is implemented, or the wrapper rejects unsupported flags clearly before contacting the container runtime.

## Actual behavior

The wrapper accepts and forwards the flag, and `agt apply` reports it as unknown only after the wrapper has already performed preliminary container work.

No resource mutation is expected because Cobra rejects the unknown flag before the apply handler runs.

## Impact

This is a documentation and wrapper-contract issue rather than a data-loss bug. It misleads users and produces a late, less actionable failure.

## Proposed fix

- Remove unsupported usage examples.
- Reject the three known unsupported flags before container checks or file copies.
- Add a shell regression test and run it in CI.

Verified against upstream `main` at `124f06d`.

YangMingSJTU:
## Summary

- stop advertising `--prune`, `--dry-run`, and `--watch` in `agentteams-apply.sh`
- reject those unsupported flags before container checks or file copies
- add a shell regression test and run it from the Helm lint workflow

## Validation

- `bash -n install/agentteams-apply.sh tests/check-apply-wrapper-flags.sh`
- `bash tests/check-apply-wrapper-flags.sh`
- `git diff --check`

Fixes #1135


shiyiyue1102 (approved):
LGTM. The wrapper now rejects the three unsupported flags before any container-side work, and the regression test is wired into the current Helm lint workflow. The focused local checks and `helm-lint` CI pass.

---

LGTM。脚本现在会在执行任何容器侧操作前拒绝这三个未支持参数，回归测试也已接入当前 Helm lint workflow；针对性本地检查和 `helm-lint` CI 均已通过。

PR #1065, opened 2026-07-21
refactor: complete the AgentTeams hard-cut rename

shiyiyue1102:
## Summary

- hard-cut active source, installer, Helm, controller, runtime paths, environment contracts, storage identifiers, Matrix aliases, and docs to the canonical AgentTeams names
- remove retired installer wrappers, CLI/env/path aliases, image tags, migration branches, and release-name compatibility
- add a CI gate that rejects retired-brand content in active files and in every tracked path
- align the CoPaw bridge on the canonical controller entry point

## Intentional breaking boundary

This PR does not preserve backward compatibility. Fresh AgentTeams installs are the supported boundary; existing retired-name scripts, environment variables, container paths, Helm release naming, CLI aliases, and state migrations are removed by design.

Historical release notes, changelogs, blog prose, and legacy FAQ snapshots remain archival content, while their tracked filenames and all active links use AgentTeams names.

## Validation

- go test ./... (agentteams-controller)
- Helm dependency build, canonical render check, and helm lint
- make check-crd-sync
- all changed shell scripts: bash -n
- PowerShell installer parser validation
- Python compileall for CoPaw, Hermes, QwenPaw, plugins, and scripts
- AgentTeams hard-cut naming gate
- TeamHarness and WorkerFlow plugin manifest validation
- CoPaw canonical bridge materialization smoke test
- focused CoPaw tests: 30 passed; four unchanged mirror tests still expose pre-existing fixture drift, and the full existing bridge suite references pre-existing missing standard/runtime sync functions

PR #1063, opened 2026-07-20
refactor: complete AgentTeams runtime rename

shiyiyue1102:
## Summary

- rename the public installer and helper entrypoints to `agentteams-*`, and update Release Quick Start and current documentation
- rename the Helm chart directory, chart name, and template helpers from `hiclaw` to `agentteams`
- rename `hiclaw-controller` to `agentteams-controller`, move the Go module to `github.com/agentscope-ai/AgentTeams/agentteams-controller`, and make `agt` the canonical CLI
- rename runtime filesystem defaults to `/opt/agentteams`, `/root/agentteams-fs`, `/var/log/agentteams`, `/var/run/agentteams`, and `/data/agentteams-*`
- update CI, image build contexts, tests, docs, and Agent-facing assets to use the canonical AgentTeams names

## Upgrade compatibility

- keep deprecated `hiclaw-*` installer/helper wrappers and the `hiclaw` CLI symlink
- preserve existing resource names when upgrading a Helm release still named `hiclaw`
- migrate legacy env files, runtime paths, CoPaw state, and the embedded kine database; bind mounts that cannot be moved are exposed through canonical symlinks
- keep legacy image fallbacks and persisted/external contracts such as `sts-hiclaw`, `hiclaw-gateway`, builtin marker names, the Docker proxy name, and legacy Matrix aliases
- leave historical changelogs, blog posts, and the legacy FAQ unchanged

## Validation

- `go test ./...` in `agentteams-controller`
- `helm lint helm/agentteams`
- `bash tests/check-helm-rename-compat.sh`
- `bash tests/check-agentteams-rename-defaults.sh`
- `bash shared/tests/test-agentteams-paths.sh`
- `bash shared/tests/test-agentteams-env.sh`
- `bash shared/tests/test-oss-credentials.sh`
- Manager builtin merge tests and find-skill tests
- `bash tests/test-16-import-worker-package-default.sh`
- CoPaw config migration unit test
- shell syntax, PowerShell parser, CRD sync, and `git diff --check`

This is intentionally a Draft PR because the repository-wide mechanical rename is large and should receive focused compatibility review before merge.

PR #1152, opened 2026-08-07
fix(controller): auto-migrate legacy hiclaw.db to agentteams.db on ki…

LUOSENGWA:
## Problem

v1.2.0 renamed the embedded kine database from `hiclaw.db` to
`agentteams.db` (#1063/#1065) without a migration path. README still
promises "Upgrade to latest (preserves all data)", but after the rename an
in-place upgrade makes the embedded controller read the new DB name while
all existing state (workers, teams, CRDs) lives in the old file — the
controller silently loses its full state.

Reported from a real v1.2.0 → v1.2.1 upgrade: all Workers disappeared after
the upgrade; recovery required manually copying the old DB files to the new
name.

Note: #1063 originally shipped this migration (`agentTeamsDBPath`), but
#1065's hard-cut rename removed it because it referenced the retired brand
name. This PR restores it (as a copy, not a rename) and adds the two store
files to the rename-contract allowlist, following the existing installer
allowlist precedent from #1079.

## Fix

`StartKine` detects a legacy DB before opening the store and migrates it
once, at startup:

- If `hiclaw.db` exists and `agentteams.db` does not → copy it (plus the
  SQLite `-wal`/`-shm` sidecar files) to `agentteams.db`.
- **Idempotent**: an existing `agentteams.db` is never overwritten, so users
  who already migrated manually are untouched.
- **Rollback-safe**: copy instead of rename, so the legacy file is preserved.
- Migration failure aborts startup instead of silently opening an empty store.

---

## 问题背景

v1.2.0 将内嵌 kine 数据库文件名从 `hiclaw.db` 重命名为 `agentteams.db`
（#1063/#1065），但没有提供迁移路径。README 仍承诺 "Upgrade to latest
(preserves all data)"（升级保留所有数据），但重命名后，就地升级会让
内嵌 Controller 读取新库名，而全部既有状态（Worker、Team、CRD）仍在旧
文件中——Controller 会静默丢失全部状态。

来自真实 v1.2.0 → v1.2.1 升级案例：升级后所有 Worker 消失；恢复需要
手动把旧库文件复制为新库名。

说明：#1063 最初自带该迁移逻辑（`agentTeamsDBPath`），但 #1065 的
hard-cut rename 因迁移函数引用了已退役的品牌名而将其删除。本 PR 恢复该
迁移（改用 copy 而非 rename），并按 #1079 的 install 脚本白名单先例，
把两个 store 文件加入 rename-contract 白名单。

## 修复内容

`StartKine` 在打开存储前检测旧库并迁移一次（启动时）：

- 若 `hiclaw.db` 存在且 `agentteams.db` 不存在 → 复制（含 SQLite
  `-wal`/`-shm` 侧车文件）为 `agentteams.db`。
- **幂等**：已存在的 `agentteams.db` 绝不被覆盖，手动迁移过的用户不受影响。
- **可回退**：使用 copy 而非 rename，旧文件保留用于回退。
- 迁移失败会中止启动，而不是静默打开空库。

### DIFF
--- a/agentteams-controller/internal/store/kine.go
+++ b/agentteams-controller/internal/store/kine.go
@@ -3,6 +3,8 @@ package store
 import (
 	"context"
 	"fmt"
+	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"time"
@@ -39,6 +41,17 @@ func StartKine(ctx context.Context, cfg Config) (*KineServer, error) {
 		return nil, fmt.Errorf("failed to create data dir %s: %w", cfg.DataDir, err)
 	}
 
+	// v1.2.0 renamed the kine database from hiclaw.db to agentteams.db without
+	// migrating existing data. Detect a legacy DB and copy it to the new name
+	// (copy, not rename, so the old file is preserved as a rollback).
+	migrated, err := ensureKineDBMigration(cfg.DataDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to migrate legacy kine database: %w", err)
+	}
+	if migrated {
+		log.Printf("kine database migrated from hiclaw.db to agentteams.db in %s", cfg.DataDir)
+	}
+
 	dbPath := filepath.Join(cfg.DataDir, "agentteams.db")
 	dsn := fmt.Sprintf("sqlite://%s?_journal=WAL&cache=shared&_busy_timeout=30000", dbPath)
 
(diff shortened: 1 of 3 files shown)


shiyiyue1102:
Thanks for investigating this. After checking the released tags, v1.2.0 already uses `agentteams.db`, so a normal v1.2.0 → v1.2.1/v1.2.2 upgrade does not cross this database rename.

Supporting upgrades from pre-v1.2 releases would require a broader migration contract because the default data volume and other storage/runtime contracts also changed. Copying `hiclaw.db` only inside the currently mounted data directory would not cover the normal legacy-volume case.

Therefore, we do not plan to merge this PR as-is. If pre-v1.2 migration support is still required, please open an issue so the complete upgrade boundary can be discussed first.

---

感谢排查。核对已发布 Tag 后，v1.2.0 已经使用 `agentteams.db`，因此正常的 v1.2.0 → v1.2.1/v1.2.2 升级不会跨越这次数据库改名。

如果要支持 v1.2 之前版本的升级，则需要定义更完整的迁移契约，因为默认数据卷以及其他存储、运行时契约也发生了变化。仅在当前挂载目录内复制 `hiclaw.db`，无法覆盖常规的旧数据卷场景。

因此我们不建议按当前方案合并。如果仍需支持 pre-v1.2 迁移，请先创建 Issue，讨论完整的升级边界和方案。

PR #1168, opened 2026-08-11
fix(manager): standardize Worker Skill installation

### LINKED ISSUE #1180: fix(manager): make Worker Skill replacement transactional
## Problem

Replacing an existing Worker Skill has two consistency gaps:

1. If distribution fails for a Skill that was already assigned to the Worker, the installer sees the pre-existing assignment and keeps the replacement canonical files instead of restoring the previous version.
2. `mc mirror --overwrite` does not remove destination-only files, so scripts or references deleted from a replacement remain in Worker storage.

## Expected behavior

- Restore the previous canonical Skill whenever replacement distribution fails.
- Synchronize the Worker Skill destination exactly so removed files do not survive.
- Add regression coverage for both cases.

This issue tracks the follow-up requested during review of PR #1168.

NingHua-jzp:
## Summary

- route new Worker Skill ZIP attachments through a deterministic safe installer
- keep assign_when optional for explicit Worker assignment and use it only for automatic role matching
- validate archives before extraction, preserve complete Skill directories, and verify Worker CR reconciliation
- roll back newly staged canonical Skill files when downstream distribution fails before assignment
- restore the previous canonical version when replacement distribution fails
- synchronize replacement Skill directories exactly so deleted files do not remain in Worker storage
- prevent Manager prompts from bypassing the supported installation and distribution scripts
- update bilingual documentation and integration coverage

## Validation

- bash manager/tests/test-install-worker-skill.sh: 26 passed
- bash manager/tests/test-push-worker-skills.sh: 20 passed
- shell syntax checks and git diff --check passed
- built agentteams/manager-qwenpaw:skill-flow-20260811-120257-r2 locally
- installed the image into a local Manager instance and confirmed Running state
- sent a Worker Skill ZIP without assign_when through a real Matrix conversation
- verified canonical storage, MinIO, Worker CR assignment, Worker runtime files, and actual Skill response

## Related issue

Closes #1180


shiyiyue1102 (changes requested):
Thanks for the update. The overall installation flow looks reasonable, but two replacement-path issues still need to be addressed before merging:

1. When `--replace` fails for a Skill already assigned to the Worker, the failure check sees the pre-existing assignment and keeps the new canonical files instead of restoring the backup.
2. `mc mirror --overwrite` does not remove destination-only files, so scripts or references deleted from the new Skill version remain in Worker storage.

Could you restore the previous canonical version on replacement failure, make replacement synchronize the destination exactly, and add regression tests for both scenarios? Please also link the corresponding issue for this non-trivial behavioral change.

---

整体安装流程方向合理，但替换路径仍有两个问题需要在合并前修复：

1. 对已经分配给 Worker 的 Skill 执行 `--replace` 时，如果分发失败，失败检查会把原有分配误认为本次成功，从而保留新的 canonical 文件并删除备份。
2. `mc mirror --overwrite` 不会删除仅存在于目标端的文件，因此新版已经移除的脚本或参考文件仍会残留在 Worker 存储中。

请在替换失败时恢复旧 canonical 版本，实现目标目录的精确同步，并为这两个场景补充回归测试。另外，这是一项非简单行为变更，请关联对应 Issue。

shiyiyue1102 (approved):
LGTM. The replacement path now restores both canonical and distributed Worker Skill files after a partial mirror failure, reports compensating rollback failures explicitly, and is covered by focused regression tests. All 19 applicable CI checks passed; the 2 skipped checks are expected for this change.

---

LGTM。替换路径现在会在 mirror 部分失败后同时恢复 canonical 和 Worker 端分发文件，并明确报告补偿回滚失败；对应回归测试已经覆盖。19 个适用 CI 检查全部通过，另外 2 个跳过项符合预期。

Issue #1177, opened 2026-08-13
[Proposal] Recover submitted TeamHarness tasks when the completion wake-up is lost


jesseedcp:
## Summary

On current `main` (`aa650ccacc2ba6171d1b0b5efd2a49b1472abe5d`), TeamHarness persists a Worker result before the Leader continuation is triggered. `taskflow submit_task` writes `TaskMeta.status = submitted`, updates the matching project node, syncs the task, and returns `notificationNeeded`. However, `notificationNeeded` is only a hint; it does not send a message. The Worker must make a separate Matrix/message tool call.

If the Worker turn or process stops after `submit_task` succeeds but before that follow-up message is sent, the result remains available, but no component re-wakes the Leader. The project can remain at `submitted` even though the result and deliverables are present.

I would like to discuss a narrow first recovery slice for TeamHarness projects: recover **submitted-but-unaccepted** tasks with at-least-once wake-up and idempotent state transitions, while keeping result review and acceptance an explicit Leader action.

## Verified reproduction

I reproduced the persistence/notification gap against the exact commit above through the production `server.call_tool("taskflow", ...)` entry point. The external `mc` process boundary was replaced with a successful test double; the real `submit_task`, task/project writes, task sync path, and `notificationNeeded` construction were exercised.

1. Create and delegate a TeamHarness project task.
2. Let the Worker acknowledge it, write `result.md`, and call `taskflow submit_task`.
3. Confirm that the call returns `ok: true`, the task and project node are `submitted`, task sync succeeds, and the response contains `notificationNeeded`.
4. End execution immediately after the MCP result, without making the separate completion-message tool call.
5. Inspect the persisted state and registered background work.

Observed result (repeated three times):

```json
{
  "submit_ok": true,
  "synced": true,
  "task_status": "submitted",
  "project_status": "submitted",
  "notificationNeeded": {
    "event": "submit_task",
    "projectId": "project-1",
    "targetRoom": "!task-room:example.test"
  },
  "durable_or_automatic_continuation": false
}
```

Relevant current behavior:

- [`_notification_needed`](https://github.com/agentscope-ai/AgentTeams/blob/aa650ccacc2ba6171d1b0b5efd2a49b1472abe5d/plugins/teamharness/mcp/server.py#L3157-L3168) explicitly says that it does not send a message.
[...]

PR #1172, opened 2026-08-12
feat(projects): add human intervention and lifecycle write APIs

LUOSENGWA:
## Summary

#1169 added read-only project / workflow inspection. This PR adds
human intervention and lifecycle: pause / resume / replan a project plus
create / cancel task / complete project, so humans can control
agent-orchestrated workflows. Writes are code-level authorized, mtime
optimistic-locked, and take effect on the next worker read via a new
read-before-read pull in the TeamHarness MCP entry points.

## What's included

1. `POST /api/v1/projects` — create a project (structured, aligned with `create_project`)
2. `POST /api/v1/projects/{id}/pause` — set project status to `paused` (stops new dispatch; in-flight tasks continue)
3. `POST /api/v1/projects/{id}/resume` — set project status back to `active`
4. `POST /api/v1/projects/{id}/replan` — replace the DAG plan (validated: duplicate ids / unknown dependencies / cycles rejected)
5. `POST /api/v1/projects/{id}/tasks/{taskId}/cancel` — cancel a single task (reason required; terminal tasks 409)
6. `POST /api/v1/projects/{id}/complete` — accept/complete the whole project (all tasks terminal required)
7. Worker-side read-before-read: projectflow read paths (resolve_project / ready_nodes / ready_loop_nodes) and taskflow actions now pull `shared/projects/{id}/meta.json` from object storage first, so a Controller write takes effect on the next read — and plan.md re-renders when the authoritative status/tasks/plan_type change
8. RBAC: `ActionCreate`/`ActionUpdate` + "project" + `requireSameTeam` (Admin/Manager/TeamLeader/L2 human within accessibleTeams), plus handler-side `checkProjectAccess` (the middleware cannot map project -> team)
9. Safety: mtime optimistic lock (`StatMeta` compare-before-write -> 409 on concurrent worker push), state-machine guards (pause idempotency 409 / resume requires paused 409 / replan requires dag+active+no-in-flight 409 / cancel terminal 409 / complete all-terminal 409)
10. Audit fields `updated_by` / `updated_at` / `pause_reason` (scalar, survive the worker's read-modify-write)
11. Intervention notification: `SendMessageAsAdmin` to `source_room_id` after each successful write (best-effort)
12. Docs: bilingual `project-workflow-api.md` write endpoints + notification + authorization matrix
13. `agt project` CLI write commands — pause / resume / replan / create / cancel / complete (L2 human / admin / team leader without raw curl)
[...]

PR #1169, opened 2026-08-11
feat(controller): add project/workflow query API for human-visible workflows

LUOSENGWA:
## Summary

Workflows are orchestrated by agents (team leader / manager) via TeamHarness
MCP projectflow/taskflow, but humans currently have **no API to inspect them**.
This PR adds a Controller-level read API so any frontend (dashboard, console
plugin) can render the workflow graph and current state — including the
multi-tenant L2 path: an L2 human (Human CR, permissionLevel=2) authenticates
with their Matrix access token and sees every team they control in one view.

## What's included

1. `GET /api/v1/projects` — list projects (dual-prefix scan: `teams/{team}/shared/projects/` + `shared/projects/`; `?team=` filter; team leaders scan only their own prefix(es); L2 humans aggregate all `accessibleTeams`)
2. `GET /api/v1/projects/{id}/workflow` — workflow detail:
   - `nodes` / `edges` (DAG graph, `conditional` edges)
   - `next` — ready nodes (mirrors upstream `_ready_nodes`/`_ready_loop_nodes`: only raw `planned`/`assigned` tasks with all deps `completed`)
   - `interrupts` — human-decision waiting points (blocked task / loop `waiting_user`, plus a paused project surfaces as a `project` interrupt)
   - `loop` — loop plan state when `plan_type == "loop"`
   - `values` — state summary (LangGraph StateSnapshot.values analog, per-status `task_count`)
   - `source` / `requester` / `requester_report` / `reply_route` / `source_room_id` passthrough
   - status normalized to `pending | delegated | in-progress | completed | revision | blocked`
   - **`?includeTasks=true`** — attach a `tasks_detail` array with each task's
     TaskMeta (spec path, assigned-to, submission summary / result status,
     deliverables, result path, cancel reason) so dashboards and the console
     plugin can expand a task into spec + artifact links. Default `false`
     keeps the response lightweight; missing TaskMeta / per-task read errors
     are skipped (the node summary stays authoritative).
3. **Multi-tenant L2 access (A2)** — humans authenticate with a Matrix access
   token (validated via `/_matrix/client/v3/account/whoami`, mapped to a Human
   CR by its authoritative `Status.MatrixUserID`, `permissionLevel=2`). The
   human's `accessibleTeams` becomes the multi-team scope: `ListProjects`
   aggregates every accessible team, `GetProjectWorkflow` allows any of them.
   L2 humans resolve to a dedicated **read-only `RoleHuman`** (not
[...]

PR #1183, opened 2026-08-15
feat(teamharness): persist durable task continuation state

jesseedcp:
## Why

`submit_task` writes the Worker result before the Worker sends its separate completion message. If the process stops in between, the result survives but nothing reliably identifies that submission or records that the Leader still needs to review it.

This PR adds that durable state contract to the CoPaw-native tools and the standalone TeamHarness MCP server. It also updates the Controller cancel endpoint from #1172 so a human cancellation cannot leave a submitted task with `continuation.status=pending`.

## What changed

On first submit, TaskMeta records an immutable `submission_id`, `submitted_at`, canonical result digest, and deterministic pending continuation ID. An identical retry reuses the first submission; a different result is rejected. Late Worker calls cannot move a terminal task back to `in_progress` or `submitted`.

Leader accept/cancel decisions are fenced by the current `submissionId`. Repeating the same decision is idempotent, while stale or conflicting decisions fail before state is changed. CoPaw and the standalone MCP path use the same result-status and continuation-resolution rules.

The Controller's project cancel API now follows the same cancellation contract:

- a submitted TaskMeta requires the exact current `submissionId`;
- successful cancellation preserves the delivery ID and resolves the continuation as `cancelled`;
- ProjectMeta stores a small cancellation decision envelope before TaskMeta is written, so a failed second write can be retried without losing the original reason, replacement task, submission ID, or timestamp;
- a replan cannot drop, reopen, or remove-and-reuse a task ID that carries a committed cancellation decision;
- TaskMeta ownership is checked against both the route task ID and project ID before any write.

`GET .../workflow?includeTasks=true` returns `tasks_detail[].submission_id`. The single-task inspection endpoint from #1230 (`GET .../tasks/{taskId}`) also returns the same opaque identity, so callers can pass it straight into a fenced cancellation. The CLI exposes the full path without guessing the current generation:

```bash
agt get projects demo-project-001 --include-tasks -o json
agt project cancel demo-project-001 demo-project-001-01 \
  --reason "no longer needed" \
  --submission-id submission-123 \
  --team biz-team
```

Local state files use same-directory atomic replacement. Result files and deliverables are uploaded and verified before remote TaskMeta is committed. Partial project/task writes return retryable state, and a matching retry repairs the missing projection without rotating the submission identity or repeating the terminal transition.

## Compatibility

Existing CoPaw `submit_task()` callers still receive `TaskMeta`, and the legacy positional `event_id` slot is unchanged.

Legacy tasks without a submission identity remain cancellable without `submissionId`. Once TaskMeta has an identity, both TeamHarness and Controller cancellation require the exact value. The Controller API therefore treats `submissionId` as conditionally required rather than globally required.

The standalone MCP used to accept `FAILED` and `PARTIAL` in `submit_task`, then fail later because neither status had an acceptance mapping. It now rejects both at submission time. This PR also adds `INTERRUPTED`, matching CoPaw's result-status set. A caller that still sends `FAILED` or `PARTIAL` gets `unsupported result status` before task or project state is changed.

## Tests

Local checks:

- Linux Controller: `go build ./...`, `go vet ./...`, and `go test ./... -count=1`
- Race checks: `go test -race ./internal/server ./internal/workflow ./cmd/agt -count=1`
- Inspection-to-cancellation regression using an opaque submission ID; all upstream Mermaid/inspection tests retained
- 81 standalone continuation/project-pull tests
- 99 CoPaw taskflow/domain tests
- TeamHarness Ruby contract checks
- TeamHarness Ruby taskflow integration
- `git diff --check`

Rebased onto `main` at `ac3818d2`, including #1169, #1172, and the Mermaid/task-inspection changes from #1230.

## Follow-up

This PR does not add the periodic scanner, wake a sleeping Team Leader, or send a recovery event through Matrix. It also does not automatically accept a Worker result.

The follow-up for #1177 is the Controller side: a leader-elected scanner over the canonical pending submissions, Leader lifecycle wake-up, and structured retryable Matrix delivery. TeamHarness remains responsible for eligibility, fencing, and terminal decisions.

Related to #1177.

---
## Why

`submit_task` writes the Worker result before the Worker sends its separate completion message. If the process stops in between, the result survives but nothing reliably identifies that submission or records that the Leader still needs to review it.

This PR adds that durable state contract to the CoPaw-native tools and the standalone TeamHarness MCP server. It also updates the Controller cancel endpoint from #1172 so a human cancellation cannot leave a submitted task with `continuation.status=pending`.

## What changed

On first submit, TaskMeta records an immutable `submission_id`, `submitted_at`, canonical result digest, and deterministic pending continuation ID. An identical retry reuses the first submission; a different result is rejected. Late Worker calls cannot move a terminal task back to `in_progress` or `submitted`.

Leader accept/cancel decisions are fenced by the current `submissionId`. Repeating the same decision is idempotent, while stale or conflicting decisions fail before state is changed. CoPaw and the standalone MCP path use the same result-status and continuation-resolution rules.

The Controller's project cancel API now follows the same cancellation contract:

- a submitted TaskMeta requires the exact current `submissionId`;
- successful cancellation preserves the delivery ID and resolves the continuation as `cancelled`;
- ProjectMeta stores a small cancellation decision envelope before TaskMeta is written, so a failed second write can be retried without losing the original reason, replacement task, submission ID, or timestamp;
- a replan cannot drop, reopen, or remove-and-reuse a task ID that carries a committed cancellation decision;
- TaskMeta ownership is checked against both the route task ID and project ID before any write.

`GET .../workflow?includeTasks=true` returns `tasks_detail[].submission_id`. The single-task inspection endpoint from #1230 (`GET .../tasks/{taskId}`) also returns the same opaque identity, so callers can pass it straight into a fenced cancellation. The CLI exposes the full path without guessing the current generation:

```bash
agt get projects demo-project-001 --include-tasks -o json
agt project cancel demo-project-001 demo-project-001-01 \
--reason "no longer needed" \
--submission-id submission-123 \
--team biz-team
```

Local state files use same-directory atomic replacement. Result files and deliverables are uploaded and verified before remote TaskMeta is committed. Partial project/task writes return retryable state, and a matching retry repairs the missing projection without rotating the submission identity or repeating the terminal transition.

## Compatibility

Existing CoPaw `submit_task()` callers still receive `TaskMeta`, and the legacy positional `event_id` slot is unchanged.

Legacy tasks without a submission identity remain cancellable without `submissionId`. Once TaskMeta has an identity, both TeamHarness and Controller cancellation require the exact value. The Controller API therefore treats `submissionId` as conditionally required rather than globally required.

The standalone MCP used to accept `FAILED` and `PARTIAL` in `submit_task`, then fail later because neither status had an acceptance mapping. It now rejects both at submission time. This PR also adds `INTERRUPTED`, matching CoPaw's result-status set. A caller that still sends `FAILED` or `PARTIAL` gets `unsupported result status` before task or project state is changed.

## Tests

Local checks:

- Linux Controller: `go build ./...`, `go vet ./...`, and `go test ./... -count=1`
- Race checks: `go test -race ./internal/server ./internal/workflow ./cmd/agt -count=1`
- Inspection-to-cancellation regression using an opaque submission ID; all upstream Mermaid/inspection tests retained
- 81 standalone continuation/project-pull tests
- 99 CoPaw taskflow/domain tests
- TeamHarness Ruby contract checks
- TeamHarness Ruby taskflow integration
- `git diff --check`

Rebased onto `main` at `ac3818d2`, including #1169, #1172, and the Mermaid/task-inspection changes from #1230.

## Follow-up

This PR does not add the periodic scanner, wake a sleeping Team Leader, or send a recovery event through Matrix. It also does not automatically accept a Worker result.

The follow-up for #1177 is the Controller side: a leader-elected scanner over the canonical pending submissions, Leader lifecycle wake-up, and structured retryable Matrix delivery. TeamHarness remains responsible for eligibility, fencing, and terminal decisions.

Related to #1177.


jesseedcp:
I moved this out of draft and rebased it onto the current `main`.

While checking the merged #1172 path, I found that Controller-side cancellation could leave a submitted TaskMeta as `status=cancelled` but keep its continuation pending. The latest commit brings that endpoint under the same submission fence as the TeamHarness tools. It also keeps a small cancellation decision on the project node so a ProjectMeta-success/TaskMeta-failure retry cannot silently change the original reason or replacement task.

I added the corresponding replan fences as well: a task with a committed cancellation cannot be reopened, dropped, or removed and re-added under the same ID. `tasks_detail` now exposes the opaque submission ID, and the CLI has `--include-tasks`, `--submission-id`, and `--team` so the HTTP fence is actually usable.

The focused Go, Python, and Ruby suites are green locally. CI is running again after the rebase and force-push.

@shiyiyue1102, when you have time, could you take a look at this state-contract slice? The periodic scanner and Matrix wake-up are still intentionally left for the follow-up to #1177.

jesseedcp:
Rebased onto current `main` (`ac22c887`) and force-pushed. The rebase was clean, and the six commits are patch-equivalent to the previous series. The new CI run is green, including the TeamHarness contracts, Controller tests, image builds, and all five integration shards.

@shiyiyue1102, could you take another look when you have time? This PR is still only the state-contract slice; the Controller scanner, Leader wake-up, and Matrix retry will stay in a separate follow-up for #1177.

plugins/teamharness/mcp/server.py
```diff
@@ -3050,6 +3118,109 @@ def _accepted_node_status(result_status: Any) -> str:
...
+def _continuation_delivery_id(project_id: str, task_id: str, submission_id: str) -> str:
+    identity = "\0".join((project_id, task_id, submission_id, "result-submitted:v1"))
+    return hashlib.sha256(identity.encode("utf-8")).hexdigest()
+
+
+def _resolve_task_continuation(task: dict[str, Any], resolution: str) -> None:
+    continuation = task.get("continuation") if isinstance(task.get("continuation"), dict) else {}
+    if not continuation:
+        return
+    continuation["status"] = "resolved"
```
oss-maintainer, line 3162: `_resolve_task_continuation` unconditionally overwrites `resolved_at` with `_utc_timestamp()` on every call, unlike CoPaw's `_commit_task_decision` which uses `continuation.get("resolved_at") or _now()` to preserve the original timestamp. Today the callers guard against calling this for already-resolved continuations at a higher level, but the function itself is not defensive. If a future code path calls it for an already-resolved continuation (e.g. during a repair that re-runs after resolved_at was already set), it would silently rotate the timestamp, violating the design contract that `resolved_at` records the first resolution time and retries must not alter it. Consider using `continuation.setdefault("resolved_at", _utc_timestamp())` or an early return when `continuation.get("status") == "resolved"`.

oss-maintainer (commented):
### Summary
This PR adds durable task continuation state across the Controller (Go), CoPaw (Python), and MCP server (Python), introducing opaque submission identity, submission-fenced accept/cancel decisions, idempotent retry with CAS-style repair, atomic file writes, canonical result digests, and publish-last shared-storage commit ordering. The design is solid and the test coverage is extensive, covering idempotency, legacy adoption, concurrent races, cross-project rejection, and retry-after-failure scenarios. However, three areas warrant attention before merge: (1) `_resolve_task_continuation` in server.py unconditionally overwrites `resolved_at` on every call, which could clobber the original decision timestamp when a leader re-accepts the same submission; (2) the `_TASK_MUTATION_LOCKS` registry in task.py grows without bound as tasks are created, leaking memory over long-running worker processes; (3) the tightened `isPlainToken` in project_handler.go now rejects tokens starting with non-alphanumeric characters, which is a backward-incompatible validation change that could break existing project IDs or tokens in production. These are not critical correctness bugs but should be addressed or explicitly acknowledged.

---

copaw/src/copaw_worker/task.py
```diff
@@ -30,6 +44,9 @@ class TaskflowError(ValueError):
     "INTERRUPTED",
 }
 EFFECTIVE_RESULT_STATUSES = {"SUCCESS", "SUCCESS_WITH_NOTES"}
+TERMINAL_TASK_STATUSES = {"completed", "revision", "blocked", "cancelled"}
+_TASK_MUTATION_LOCKS: dict[str, threading.RLock] = {}
```
oss-maintainer, line 48: `_TASK_MUTATION_LOCKS` is a module-level dict that accumulates one `threading.RLock` per unique task meta.json path and is never pruned. In a long-running CoPaw process that handles many distinct tasks over its lifetime, this grows without bound. Each entry is small (string key + RLock), so this is unlikely to cause practical memory pressure, but it is a resource leak by design. Consider using a bounded cache (e.g. an `OrderedDict` capped at a few hundred entries with LRU eviction) or keying by `(workspace_dir, task_id)` with periodic cleanup.

agentteams-controller/internal/server/project_handler.go
```diff
@@ -2728,18 +2846,25 @@ func (h *ProjectHandler) CreateProject(w http.ResponseWriter, r *http.Request) {
...
-// key (no path traversal / separators).
+// isPlainToken reports whether s matches TeamHarness's safe-id contract:
+// [A-Za-z0-9][A-Za-z0-9._-]*.
 func isPlainToken(s string) bool {
-	for _, r := range s {
+	if s == "" {
+		return false
+	}
+	for i, r := range s {
+		alphaNumeric := r >= 'a' && r <= 'z' || r >= 'A' && r <= 'Z' || r >= '0' && r <= '9'
```
oss-maintainer, line 2856: The tightened `isPlainToken` now requires the first character to be alphanumeric (`[A-Za-z0-9]`), rejecting identifiers that start with `-`, `_`, or `.`. This is a backward-incompatible change for the `replacementTaskId` parameter in `POST /api/v1/projects/{id}/tasks/{taskId}/cancel`. If any existing caller (human CLI scripts, automated tooling, or other runtime integrations) passes replacement IDs starting with `_` or `-`, those requests will start returning 400 after upgrade. The test `TestCancelTask_InvalidReplacementID400` explicitly validates that `"_task"` and `"-task"` are now rejected. Consider whether this tightening is necessary for the cancellation feature or if the old more-permissive check should be preserved for backward compatibility.

plugins/teamharness/mcp/server.py
```diff
@@ -4367,79 +4767,291 @@ def _taskflow(arguments: dict[str, Any]) -> dict[str, Any]:
...
+                    "action": action,
+                    "task": task,
+                    "reused": True,
+                    "publishedArtifacts": [],
+                    "synced": synced,
+                    "notificationNeeded": _notification_needed(
+                        "submit_task",
+                        {"project_id": task.get("project_id", "")},
+                        task,
+                        summary=f"submit_task: {task_id} ({status})",
```
oss-maintainer, line 4847: In the `submit_task` retry path, the condition `if not task.get("result_digest") or task.get("resultDigest")` is misleading. After `_load_task` runs `_ensure_console_task_meta`, camelCase keys (including `resultDigest`) are popped from the task dict, so `task.get("resultDigest")` is always `None`/falsy at this point. The condition effectively simplifies to `if not task.get("result_digest")`. The dead `or` clause could confuse future maintainers into thinking camelCase migration is still needed here. Consider simplifying to `if not task.get("result_digest"):`.

agentteams-controller/internal/server/project_handler.go
```diff
@@ -2528,13 +2559,98 @@ func (h *ProjectHandler) CancelTask(w http.ResponseWriter, r *http.Request) {
 		writeError(w, http.StatusNotFound, "task meta not found")
 		return
 	}
+	if str(taskData["task_id"]) != taskID || str(taskData["project_id"]) != projectID {
+		writeError(w, http.StatusNotFound, "task meta not found")
+		return
```
oss-maintainer, line 2564: The new ownership check `str(taskData["task_id"]) != taskID || str(taskData["project_id"]) != projectID` correctly prevents cross-project task cancellation. However, if `taskData["project_id"]` is `nil` (absent from the stored JSON), `str(nil)` returns `""`, and the check `"" != projectID` would reject the cancellation with 404 even for a legitimately owned task. All tasks created by TeamHarness runtimes set `project_id`, so this should not affect normal workflows, but legacy or manually-created TaskMeta files without `project_id` would become un-cancellable through this API endpoint.

plugins/teamharness/mcp/server.py
```diff
@@ -3071,28 +3242,159 @@ def _payload_bool_field(payload: dict[str, Any], names: tuple[str, ...], default
```
oss-maintainer, line 3243: The `_accept_task_result` function now requires `role == "leader"` via `_role(arguments)`, and `_role` was changed to prioritize the runtime environment role (`AGENTTEAMS_AGENT_ROLE`) over the payload role argument. This is correct security hardening that prevents Workers from forging Leader operations. Verify that all production Leader deployment configurations set `AGENTTEAMS_AGENT_ROLE=leader`, since the argument-level `role` override is now ignored when a runtime role is present. Existing test infrastructure was updated accordingly.

jesseedcp:
Picked up two concrete review notes in `6bc6f1fe`. `_resolve_task_continuation` now keeps the first `resolved_at` when an idempotent accept repairs a missing TaskMeta fence; the regression test starts from an already-resolved continuation and checks that its original timestamp survives. I also removed the dead camelCase `resultDigest` condition, since persisted TaskMeta uses `result_digest`.

Locally, the continuation, project-pull, and CoPaw taskflow suites pass (66 + 15 + 99), along with the Ruby contract/taskflow checks. The new GitHub run is fully green as well: TeamHarness, Controller, all image builds, and all five integration shards passed.

oss-maintainer (approved):
Re-review of the incremental commit pushed since the last review (8d7594c2..6bc6f1fe): `fix(teamharness): preserve first continuation resolution`.

The change is correct and safe: preserving an existing `resolved_at` keeps repair/re-accept paths from overwriting the original resolution timestamp, and dropping the redundant `resultDigest` fallback is safe because `_first_text()` already prefers `result_digest` and `_write_task` normalizes camelCase keys before persisting. The new test directly exercises the preservation path on the repair flow.

LGTM. Trivial change, looks good.

---

shiyiyue1102:
This branch currently conflicts with main, so it cannot be merged or serve as a reliable base for #1233 yet. Please rebase onto the latest main, resolve the conflicts, and push the refreshed branch; we will re-run the final review and CI on that head.

当前分支与 main 存在冲突，暂不能合并，也不能作为 #1233 的可靠前置。请基于最新 main rebase 并解决冲突后推送；我们会针对新 head 重新复核并跑 CI。

shiyiyue1102 (approved):
LGTM. The current head exposes the existing opaque submission fence through task inspection and verifies inspection-to-fenced-cancellation end to end; it does not alter the durable continuation transition contract. All required build and integration checks are green.

PR #1193, opened 2026-08-19
fix: make Worker skill recovery failures non-blocking

NingHua-jzp:
## Summary

- treat an existing canonical Worker Skill copy as satisfying its declarative assignment when Manager recovery fails
- continue Worker config and container reconciliation when a declared Skill is missing and cannot be restored, while preserving a non-blocking status warning
- align English and Chinese documentation with Dashboard distribution updating Worker.spec.skills, the current Marketplace UI, and restart-based reload behavior

## Validation

- GOMODCACHE=/private/tmp/agentteams-go-mod GOCACHE=/private/tmp/agentteams-go-cache go test ./... (from agentteams-controller)
- git diff --check origin/main...HEAD
- checked documentation for stale Dashboard Skill distribution semantics


shiyiyue1102:
Thanks for the update. I think the non-blocking recovery direction is reasonable, with two ownership and observability details worth adjusting:

1. Please keep Skill reconciliation owned by `WorkerReconciler` and separate it from runtime-config ownership. Team-owned QwenPaw Workers currently skip the entire `ReconcileMemberConfig` call, which also skips the nested `reconcileMemberSkills` step. `TeamReconciler` should remain responsible only for Team-owned runtime config; the Worker reconcile should still restore and verify its assigned Skills.
2. When a remote Skill refresh fails but an existing canonical copy is available, please keep the failure non-blocking and continue running the Worker, but surface a sanitized warning in `Worker.status.message` and the Controller log. The warning should identify the Skill and requested version or label without exposing credentials that may be embedded in the remote source URL.

Could you add focused coverage for the Team-owned QwenPaw path and the “requested remote version failed, existing copy retained” path?

---

感谢更新。Skill 恢复失败不阻塞 Worker reconcile 的方向是合理的，不过建议补齐两个职责边界和可观测性细节：

1. Skill reconcile 应继续由 `WorkerReconciler` 负责，并与 runtime config 的归属解耦。当前 Team-owned QwenPaw Worker 会跳过整个 `ReconcileMemberConfig`，从而连带跳过其中的 `reconcileMemberSkills`。`TeamReconciler` 仍应只负责 Team-owned runtime config；Worker reconcile 应继续恢复和校验自身分配的 Skills。
2. remote Skill 刷新失败但规范目录中仍有旧副本时，请保持非阻塞并继续运行 Worker，同时在 `Worker.status.message` 和 Controller 日志中展示脱敏告警。告警可以标识 Skill 及请求的 version/label，但不能暴露 remote source URL 中可能携带的凭证。

请补充 Team-owned QwenPaw 场景，以及“请求的 remote 版本刷新失败、继续保留旧副本”场景的针对性测试。

shiyiyue1102 (approved):
LGTM. Worker-owned Skill reconciliation is now separated from Team-owned runtime config, remote refresh failures remain non-blocking and visible through credential-safe warnings, and focused tests cover Team-owned QwenPaw Workers plus version/label retention. The full current CI matrix passes, including the rerun of qwenpaw manager with hermes worker.

PR #1196, opened 2026-08-21
fix(release): wait for versioned images

shiyiyue1102:
## Summary

- gate GitHub Release publication on all versioned release images being present
- require both `linux/amd64` and `linux/arm64` for every image
- fail closed when image building is incomplete instead of publishing an installable release early

## Why

Tag pushes currently start `Build Images` and `Release` independently. The Release can therefore publish its installer and quick-start instructions before the matching images exist. This change keeps the existing parallel trigger but prevents the GitHub Release from becoming the latest stable version until all nine image manifests are complete.

Manual `workflow_dispatch` releases now also require images to be built first, which avoids relying on a workflow-created tag to trigger the image workflow.

## Validation

- `git diff --check`
- workflow YAML parse
- extracted gate shell passes `bash -n`
- two-platform jq predicate accepts a valid manifest and rejects an amd64-only manifest
- live read-only verification against all nine `v1.2.2` registry manifests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,54 @@ jobs:
             echo "prerelease=false" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Wait for versioned multi-architecture images
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          images=(
+            openclaw-base
+            agentteams-controller
+            agentteams-embedded
+            agentteams-manager
+            agentteams-manager-qwenpaw
+            agentteams-worker
+            agentteams-copaw-worker
+            agentteams-qwenpaw-worker
+            agentteams-hermes-worker
+          )
+          images_ready=0
+          attempt=1
+          max_attempts=80
+
+          while [ "${attempt}" -le "${max_attempts}" ]; do
+            missing=()
+            for image in "${images[@]}"; do
+              ref="${REGISTRY}/${REPO}/${image}:${VERSION}"
+              manifest="$(docker buildx imagetools inspect "${ref}" --raw 2>/dev/null || true)"
+              if ! jq -e '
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -48,6 +48,54 @@ jobs:
             echo "prerelease=false" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Wait for versioned multi-architecture images
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          images=(
```
oss-maintainer, line 55: **[Info]** The image list is hardcoded and must be kept in sync with the push targets in `build.yml` (currently all 9 images match: `openclaw-base`, `agentteams-controller`, `agentteams-embedded`, `agentteams-manager`, `agentteams-manager-qwenpaw`, `agentteams-worker`, `agentteams-copaw-worker`, `agentteams-qwenpaw-worker`, `agentteams-hermes-worker`). If a new image is added to the build matrix later, this gate will fail closed after the 40-minute wait — safe, but consider adding a comment here pointing to `build.yml` so maintainers remember to update both places.

oss-maintainer (approved):
### Summary
Adds a release gate that waits (up to 80 x 30s) for all 9 versioned images to exist in the registry with both linux/amd64 and linux/arm64 manifests before publishing the GitHub Release, failing closed otherwise.

Verified: the hardcoded image list matches the current push targets in `build.yml` (all 9 images), `REGISTRY`/`REPO` are workflow-level env vars available to the step, `docker buildx imagetools inspect` and `jq` are preinstalled on ubuntu runners, and builds are multi-arch (`MULTIARCH_PLATFORMS ?= linux/amd64,linux/arm64` with QEMU). The jq check uses `.manifests[]?` so single-platform or missing manifests fail closed as intended. One non-blocking note inline about keeping the image list in sync with `build.yml`.

LGTM — prevents publishing an installable release before its images exist.

---

PR #1201, opened 2026-08-23
fix(api): reject unknown spec fields and empty workerMembers on Team create/update

### LINKED ISSUE #1200: Team create/update silently drops unknown spec fields, producing memberless Teams and phantom "removed from desired member set" evictions
## Summary

The Team (and Worker/Human/Manager) create/update endpoints decode request bodies with a plain `json.Decoder`, without `DisallowUnknownFields`. Any spec field that does not match the controller's schema version is **silently dropped** and the request still succeeds. When the dropped field carries membership (`workers` on ≥v1.2 controllers, `workerMembers` on ≤v1.1.x controllers, or a simple typo), a Team with **zero members** is stored. The team-room membership reconciler then computes the desired member set as leader + admin only and repeatedly evicts every actually-joined worker from the team room with reason `"removed from desired member set"` — on every reconcile.

From the user's perspective the Team looks healthy (`Active`) while workers are kicked out of the team room again and again, mid-conversation.

## Reproduction

Observed on an embedded-mode installation (clean install; reproducible):

1. Apply a Team manifest whose membership field does not match the controller's schema. We hit this on a v1.1.x-era controller using the v1.2 contract:

```yaml
spec:
  leader: {name: coordinator}
  workerMembers:
    - {name: coordinator, role: team_leader}
    - {name: analyzer,    role: worker}
    - {name: fixer,       role: worker}
    - {name: tester,      role: worker}
    - {name: evaluator,   role: worker}
```

2. `apply` succeeds (201) and the Team reaches `Active`, but the stored spec contains **no membership at all**. Verified by reading the stored object directly from the kine store:

```
stored spec keys: ['description', 'leader', 'peerMentions']   # workerMembers silently gone
```

3. The controller log then shows repeated eviction of already-joined members:

```
"room member not desired; attempting removal" ... "room":"<team room>" ... currentCount=6 desiredCount=2
membership change reason: "removed from desired member set"
```

We logged 64 such removals; every worker shows `team: None` and has to be re-invited/re-joined by hand. We ultimately needed an external membership-guard loop (poll Matrix every 30s, re-invite + re-join evicted workers) to keep a multi-agent pipeline alive.

## Same defect class exists on current `main` (v1.2.3)

- `CreateTeam` / `UpdateTeam` (and the Worker/Human/Manager handlers) still decode with `json.NewDecoder(r.Body).Decode(&req)` — unknown fields (e.g. the legacy `workers` field, or typos like `humanMember`) are silently dropped (`internal/server/resource_handler.go`).
- `UpdateTeam` accepts an explicit `"workerMembers": []` and replaces the whole membership with nothing — a silent membership wipe that lands in the same eviction path (the CRD marks `workerMembers` with `MinItems=1`, but that is not enforced on this path).

Because v1.2.0 changed the membership contract (`workers` → `workerMembers`, #1072), manifests written across versions are especially likely to hit this silent-drop path.

## Expected behavior

- Unknown fields in resource create/update bodies should be rejected with HTTP 400 naming the offending field, with a migration hint for removed/renamed fields (e.g. pre-1.2 `workers` → `workerMembers`).
- Updating a Team with an empty `workerMembers` list should be rejected; omitting the field should keep the current membership.
- (Follow-up, related to #740.) Consider making the membership reconciler non-destructive when the desired set resolves to fewer members than expected, instead of evicting joined members on every reconcile.

## Related

- #740 — phantom removal of members from Team rooms
- #1072 — v1.2 membership contract change (`workers` → `workerMembers`)

viking714:
Fixes #1200

## Problem

Team create/update handlers decode request bodies with a plain \`json.Decoder\` (no \`DisallowUnknownFields\`). Spec fields that do not match the controller's schema version are silently dropped and the request still succeeds. When the dropped field carries membership, a Team with zero members is stored, and the team-room membership reconciler then repeatedly evicts every joined worker from the team room with reason \`"removed from desired member set"\` (observed 64 evictions on a clean install; see #1200 for the full reproduction).

Additionally, \`UpdateTeam\` accepts an explicit \`"workerMembers": []\` and silently wipes the whole membership — the same eviction path (the CRD marks \`workerMembers\` \`MinItems=1\`, but that is not enforced here).

## Changes

- Add \`decodeResourceRequest\`: all resource create/update request bodies (Worker/Team/Human/Manager) are now decoded with \`DisallowUnknownFields\`. Unknown fields yield HTTP 400 naming the offending field; the removed pre-1.2 \`workers\` field carries a migration hint pointing at \`workerMembers\`.
- \`UpdateTeam\` rejects an explicit empty \`workerMembers\` list; omitting the field keeps the current membership (existing behavior preserved).

## Tests

- \`TestCreateTeamRejectsUnknownFields\` — legacy \`workers\` hint + generic unknown-field rejection; no Team is persisted.
- \`TestUpdateTeamRejectsEmptyWorkerMembersAndUnknownFields\` — empty-list wipe rejected and membership preserved; unknown fields rejected on update.

Both files are gofmt-clean. Note: my local toolchain is Go 1.16 while the module requires Go 1.25, so the suite was not executed locally — relying on CI to run it. Happy to adjust per maintainer preference (e.g. scoping strict decoding to Team endpoints only).

### DIFF
--- a/agentteams-controller/internal/server/resource_handler.go
+++ b/agentteams-controller/internal/server/resource_handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
 	"time"
 
 	v1beta1 "github.com/agentscope-ai/AgentTeams/agentteams-controller/api/v1beta1"
@@ -19,6 +20,39 @@ import (
 // k8sUpdateMaxRetries is the max attempts for Get→patch spec→Update against
 // optimistic locking conflicts when the controller updates status between Get and Update.
 const k8sUpdateMaxRetries = 3
+
+// unknownFieldPattern extracts the offending field name from the error
+// produced by json.Decoder.DisallowUnknownFields.
+var unknownFieldPattern = regexp.MustCompile(`unknown field "([^"]+)"`)
+
+// legacyFieldHints maps removed or renamed spec fields to migration guidance.
+// Without strict decoding, manifests written for another controller version
+// (or with a typo) are silently stripped during decode: the resource is
+// created/updated without its members, and the membership reconciler then
+// repeatedly evicts the orphaned members from the team room ("removed from
+// desired member set").
+var legacyFieldHints = map[string]string{
+	"workers": "`workers` was removed in v1.2.0; create standalone Worker resources and reference them by name via `workerMembers`",
+}
+
+// decodeResourceRequest decodes a resource create/update request body,
+// rejecting unknown fields so that schema drift between the client manifest
+// and the controller version fails loudly instead of silently dropping spec
+// fields such as team membership.
+func decodeResourceRequest(r *http.Request, v interface{}) error {
+	dec := json.NewDecoder(r.Body)
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(v); err != nil {
+		if m := unknownFieldPattern.FindStringSubmatch(err.Error()); m != nil {
+			if hint, ok := legacyFieldHints[m[1]]; ok {
+				return fmt.Errorf("unknown field %q: %s", m[1], hint)
+			}
+			return fmt.Errorf("unknown field %q in request body; check the field name against the current resource schema", m[1])
+		}
+		return err
+	}
+	return nil
+}
 
 // ResourceHandler handles declarative CRUD operations on CRs.
 //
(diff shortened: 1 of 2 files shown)


oss-maintainer (approved):
### Summary
Adds strict request-body decoding (`DisallowUnknownFields`) to all resource create/update REST endpoints plus an explicit rejection of an empty `workerMembers` list on Team update, so schema drift or typos fail loudly instead of silently dropping membership and triggering repeated room evictions.

Verified against the current codebase:
- All write endpoints are covered (`CreateWorker`, `UpdateWorker`, `CreateTeam`, `UpdateTeam`, `CreateHuman`, `CreateManager`, `UpdateManager`); there is no `UpdateHuman` handler, so nothing is missed.
- `CreateTeam` already rejects empty `workerMembers`, so the new update-side check closes the real gap symmetrically.
- The `workers` → `workerMembers` migration hint is accurate (field removed in v1.2.0), and the regex fallback returns the original error when the message shape changes.
- Tests cover unknown-field rejection on create (nothing persisted), the legacy hint, empty-list rejection on update (membership preserved), and unknown fields on update.

LGTM, nice defensive fix with solid test coverage.

---

PR #1204, opened 2026-08-27
feat: add experimental DeepSeek Harness worker runtime

### LINKED ISSUE #1203: Add DeepSeek Harness as an experimental Worker runtime
## Problem

AgentTeams supports several Worker runtimes, but DeepSeek Harness cannot currently join a Team as a managed Worker. Running DSH in a separate container is not enough: it still needs the Worker identity and Matrix rooms created by the Controller, TeamHarness role context, Workspace sync, and a session that survives Pod replacement.

## Proposed behavior

Add `deepseek-harness` as an experimental, Worker-only runtime.

A DSH Worker should be able to:

- start from a normal Worker or Team resource;
- receive Matrix text, image, and file messages and return Workspace files;
- keep one DSH session per Matrix room and restore it after a restart;
- persist Matrix cursor and delivery state so replayed events do not run the same model turn twice;
- read the same TeamHarness member and role context as the other Worker runtimes.

For retries, a restarted `processing` event should keep the same attempt identity. A new attempt should only be created after the previous attempt is recorded as failed.

## Scope

This would add the Worker image, Controller/CRD/Helm wiring, local installer option, CI/release jobs, documentation, and runtime tests. It would not add a DeepSeek Harness Manager runtime.

The integration should stay marked experimental and pin a tested DSH version while the upstream package is still on an RC release. The Kubernetes/MinIO path can be covered in the first PR; live Alibaba OSS verification can follow separately.

I have an implementation ready and will link the PR to this issue.

jesseedcp:
## Summary

Closes #1203.

This PR adds DeepSeek Harness as an experimental Worker runtime. It is Worker-only; Manager runtime selection is unchanged. The image pins `@deepseek-ai/dsh@0.1.1-rc.2` so the adapter is tied to a version we have actually tested.

## What changed

- The Controller, CRD, Helm chart, CLI, and local installers now recognize `deepseek-harness` as a Worker runtime.
- A managed Worker image runs DSH headlessly behind a Matrix bridge and receives the normal Controller-generated identity, room, Team, model, and storage configuration.
- The TeamHarness adapter injects the member/role context into DSH and uses DSH session events for turn recovery.
- Matrix text, image, and file input are supported. Attachments are confined to the Workspace inbox, and changed outbox files are returned to Matrix.
- Build, release, Helm, installer, and integration workflows include the new image and runtime checks.
- English, Chinese, and Japanese runtime documentation now list the experimental Worker option.

## Restart and retry behavior

Each Matrix room has its own persisted DSH session. The bridge also persists its Matrix cursor, event status, output snapshot, and reply transaction identity through the existing Worker storage path.

A Pod restart while an event is still `processing` reuses the same attempt and DSH message ID. Only an event recorded as `failed` advances to a new attempt. This lets DSH reuse a completed turn after a crash without turning an actual model failure into a permanent no-op retry.

Team-room filtering ignores ordinary replies from peer Agents and accepts a new peer task only when the Matrix event contains a structured mention for that Worker. This prevents two Agents from replying to each other indefinitely.

## Verification

- `go test -p 1 ./...` in `agentteams-controller`: pass
- Python bridge/runtime suite: 27 tests pass locally; 2 symlink-confinement cases are skipped on Windows
- Node retry/recovery suite: 2/2 pass
- Bash and PowerShell installer checks: pass
- Actionlint and Helm lint: pass
- Worker image build and real DSH model smoke: pass
- Kind Team E2E: Controller-created Leader and Worker, role context, quiet-room check, persisted event replay after Pod replacement, room-session recovery after Leader replacement, and Matrix file round trip all pass

## Current limits

The runtime is intentionally marked experimental because the pinned DSH dependency is still an RC release. The Kubernetes path has been exercised against the repository's MinIO-backed test environment. I have not run the same recovery test against a live Alibaba OSS deployment yet.


shiyiyue1102:
Thanks for putting this together. Adding DeepSeek Harness as an experimental managed Worker runtime is valuable for AgentTeams’ heterogeneous-runtime direction. This PR covers not only the Worker image, but also Controller/CRD/Helm/installer/release wiring, TeamHarness context, session recovery, idempotent delivery, and file round trips. That breadth makes it a strong foundation.

Before merging, I think we should close a few Matrix routing gaps:

1. Bridge replies containing a visible Matrix user ID do not emit `m.mentions`, so DSH-to-DSH task completion messages can be ignored.
2. The inbound filter drops all Agent replies, even when the reply explicitly targets this Worker through `m.mentions`.
3. Unmentioned human messages in a Team Room currently trigger every DSH member.

Could you make the routing room-aware, preserve/emit structured mentions, accept explicitly targeted replies, and add E2E coverage for DSH Worker completion reaching a DSH Leader and for unmentioned Team Room chatter remaining quiet?

One release detail: the PowerShell installer maps `latest` to the hard-coded v1.2.3 baseline when checking DSH availability. Please ensure the v1.2.4 release updates or otherwise probes this gate so the default Windows installation can expose the new runtime.

---

感谢完成这项工作。将 DeepSeek Harness 作为实验性的托管 Worker Runtime 接入，对 AgentTeams 支持异构 Agent Runtime 的方向很有价值。这个 PR 不只是增加了 Worker 镜像，还完整覆盖了 Controller、CRD、Helm、安装与发布链路，以及 TeamHarness 上下文、Session 恢复、幂等投递和文件往返，是一个很扎实的基础。

合并前建议补齐几个 Matrix 路由问题：

1. Bridge 回复即使正文包含可见的 Matrix 用户 ID，也不会生成 `m.mentions`，因此 DSH 到 DSH 的任务完成消息可能被忽略。
2. 入站过滤会丢弃所有 Agent reply，即使该 reply 已通过 `m.mentions` 明确指向当前 Worker。
3. Team Room 中未 mention 的 Human 消息目前会触发所有 DSH 成员。

建议让路由逻辑区分 Personal Room 与 Team Room，保留并生成结构化 mention，允许显式指向当前 Worker 的 reply，并补充两条 E2E：DSH Worker 的完成消息能够到达 DSH Leader，以及 Team Room 中未 mention 的普通消息保持安静。

另外还有一个发布细节：PowerShell 安装脚本在判断 DSH 是否可用时，会把 `latest` 映射到硬编码的 v1.2.3 基线。请确保 v1.2.4 发布时同步更新或动态探测这个门禁，使默认 Windows 安装能够展示新 Runtime。

shiyiyue1102:
补充看了一下这次新增镜像对发布链路的影响。这个问题不影响我对 DSH 接入价值的认可，但当前实现确实把一个 experimental runtime 绑定进了核心版本的发布列车：tag 构建会无条件加入 `deepseek-harness-worker`，Worker 的多架构 push 在同一个 `make` 调用中顺序执行，同时 Helm 默认 tag 和安装脚本都直接复用 `global.imageTag` / `AGENTTEAMS_VERSION`。以后每增加一个 Runtime，核心版本发布耗时和失败面都会线性增长。

目前 PR CI 中 DSH 的单架构构建约 1 分 53 秒，而且是并行任务，因此它还不是普通 Integration CI 的关键路径；真正需要解耦的是正式 tag 的多架构发布。

成熟项目通常把扩展实现与核心发布解耦：

- [KServe ServingRuntime](https://kserve.github.io/website/docs/0.16/concepts/resources/servingruntime) 把 Runtime 的镜像、支持格式和版本作为可注册资源，新增自定义 Runtime 不需要修改 Controller；
- [Apache Airflow Providers](https://airflow.apache.org/docs/apache-airflow-providers/) 独立版本、独立发布，可以单独升级或回退；
- [Dify Plugins](https://docs.dify.ai/en/develop-plugin/publishing/marketplace-listing/release-overview) 也是独立包和独立分发渠道。

建议 AgentTeams 对 experimental/community runtime 采用类似的 Runtime Catalog + 独立发布列车：

1. 核心仓库保留 Runtime 接口、Controller 调度和兼容性测试；
2. DSH 镜像仅在 `deepseek-harness/**`、共享 Runtime ABI 变化或手动触发时发布；
3. DSH 使用独立版本或不可变 digest，Helm/安装器不要默认拼接核心 `AGENTTEAMS_VERSION`；
4. 核心 release 只校验 catalog 中镜像存在，并对已发布镜像跑兼容性 smoke test；未变化时复用上次验证过的 digest。

我不建议把所有 Runtime 合并成一个大镜像，那会放大镜像体积、依赖冲突和安全更新范围。短期最小改动是把 DSH 从 tag 的 `all` 目标中移出，增加独立 workflow 和独立 tag；中期再把 OpenClaw、CoPaw、Hermes、QwenPaw、DSH 统一收敛成带兼容版本的 Runtime Catalog。至少建议在 DSH 转 GA 前完成这个解耦，最好从 experimental 阶段就避免形成锁步发布约定。

---

I took an additional look at the image-release impact. This does not reduce the value of the DSH integration, but the current implementation couples an experimental runtime to the core release train: tag builds always include `deepseek-harness-worker`, multi-architecture Worker pushes run sequentially in one `make` invocation, and both Helm and the installer derive the DSH tag from `global.imageTag` / `AGENTTEAMS_VERSION`. As more runtimes are added, core release time and failure surface will grow linearly.

In the current PR CI, the single-architecture DSH build took about 1m53s and ran in parallel, so it is not the Integration CI critical path. The part that needs decoupling is the production tag and multi-architecture release path.

Mature ecosystems normally separate extensions from the core release: KServe registers runtime images and supported versions through ServingRuntime resources without controller changes; Airflow Providers are independently versioned and released; Dify Plugins are independently packaged and distributed.

I suggest adopting the same Runtime Catalog plus independent release-train model for experimental/community runtimes:

1. Keep the runtime contract, controller scheduling, and compatibility tests in core.
2. Publish the DSH image only when `deepseek-harness/**`, the shared runtime ABI, or an explicit manual trigger changes.
3. Give DSH an independent version or immutable digest; Helm and the installer should not derive it from the core `AGENTTEAMS_VERSION`.
4. Let a core release verify that the catalog image exists and run a compatibility smoke test against it, while reusing the last verified digest when DSH itself is unchanged.

I would not combine all runtimes into one large image, because that increases image size, dependency conflicts, and the blast radius of security updates. The smallest near-term change is to remove DSH from the tag-build `all` target and add a dedicated workflow and tag. In the medium term, OpenClaw, CoPaw, Hermes, QwenPaw, and DSH can converge on a versioned Runtime Catalog. This should be completed before DSH graduates to GA, and preferably while it is still experimental so lockstep release does not become an established contract.

jesseedcp:
按这两轮 review 的意见补完了，主要是两个提交：

- `cffe5cc3` 把 Matrix 路由改成 room-aware：Personal Room 仍可直接对话；Team Room 只有明确 mention 当前 Worker（或 room mention）才会触发。Agent reply 不再一律丢弃，只要 `m.mentions` 指向当前 Worker 就会接收。DSH 回复里出现已知 Agent 的 Matrix ID 时，也会写入对应的 `m.mentions`。
- `2cd48faf` 把 DSH 镜像从核心发布列车拆了出来。现在使用独立的 `v0.1.0` runtime 版本和 `deepseek-harness-v*` / dedicated workflow；core tag、RC 和 Makefile 的聚合 push 都不会再发布或打 DSH 的 tag。Helm、安装器和 Controller 默认值也都固定到独立版本。core release 只检查该镜像是否存在，并对镜像内的 DSH/TeamHarness/bridge contract 做 smoke test。

PowerShell 安装器的 `latest` 门禁也改成了运行时查询 GitHub 最新稳定版，API 不可用时才回退到脚本里的基线；Bash 做了同样处理，并补了两边的测试。

我在本地 kind 环境重跑了完整 Team E2E。下面这些都过了：

- unmentioned Team Room chatter 不触发 Leader/Worker
- DSH Worker 的完成消息带 Leader 的 `m.mentions`，DSH Leader 能收到并回复
- replay 去重
- Leader Pod 替换后的跨消息 session 恢复
- Matrix file -> Workspace -> DSH -> Matrix file

PR CI 已全部通过，包括 DSH tests/image build、Helm/CRD、Controller tests 和 5 组 integration shards。Runtime Catalog 这次没有一起展开，先按建议落了独立发布列车这个短期边界。

shiyiyue1102:
Follow-up after locally regressing the current head `8b011039`.

The standalone DSH Worker path is largely in place. I built the ARM64 image from this PR, created a `deepseek-harness` Worker through the PR Controller, and verified the image entrypoint, TeamHarness adapter/skills, Matrix text ingestion, retry/final-failure delivery, Matrix file download/upload, bridge-state persistence, container recreation, and MinIO-backed DSH session restore. Running the packaged `agentteams-dsh` against an isolated OpenAI-compatible mock also completed and resumed a real DSH session successfully.

There are still a few items to address before merge:

1. **Agent-to-Agent replies are not reliably targeted.** The success path derives `m.mentions` only from Matrix IDs that the model happens to include in its answer. The current Team E2E explicitly asks the Worker to echo the Leader ID, which masks the normal case where a completion simply says that the task is done. Final failure messages and uploaded file events also carry no `m.mentions`. Since the receiving DSH Agent drops Agent messages without a structured mention, these replies can be silently ignored. Please automatically mention the source Agent for replies and propagate the same target to success text, terminal failures, and files; add an E2E that does not ask the model to repeat the Leader ID.
2. **The dedicated release workflow still interpolates `inputs.version` and `github.ref_name` directly into Bash before validation.** Please pass these values through step-level `env` variables and validate/use the quoted shell variables, including the image-build step.
3. **E2EE is selectable but unsupported by this bridge.** With `AGENTTEAMS_MATRIX_E2EE=1`, room traffic is `m.room.encrypted`, while the DSH bridge only handles `m.room.message`. Please either reject/gate this runtime combination, fail fast with a clear message, or add E2EE support and tests.
4. **The pinned default image is not yet deliverable.** A live manifest lookup for `higress-registry.cn-hangzhou.cr.aliyuncs.com/agentteams/agentteams-deepseek-harness-worker:v0.1.0` still returns `not found`. Please publish and verify both `linux/amd64` and `linux/arm64` manifests before the default is relied on.

The existing GitHub checks are green. Locally, the focused Go, installer, and Node tests passed; the Python suite passed 30/31, with the remaining failure caused by the macOS `/tmp` versus `/private/tmp` path alias in the test assertion.

---

补充一下对当前 head `8b011039` 的本地回归结果。

Standalone DSH Worker 的主体链路已经基本具备。我从本 PR 构建了 ARM64 镜像，通过 PR Controller 创建了 `deepseek-harness` Worker，并验证了镜像入口、TeamHarness Adapter/Skills、Matrix 文本入站、重试与最终失败回复、Matrix 文件下载/上传、Bridge 状态持久化、容器删除重建，以及基于 MinIO 的 DSH Session 恢复。使用隔离的 OpenAI-compatible Mock 运行镜像内正式的 `agentteams-dsh`，真实 DSH Session 的执行与 resume 也都成功。

合并前仍建议处理以下问题：

1. **Agent-to-Agent 回复不能稳定定向。** 成功路径只会从模型答案正文中扫描 Matrix ID 来生成 `m.mentions`。现有 Team E2E 明确要求 Worker 原样输出 Leader ID，因此掩盖了普通完成回复不包含 ID 的场景。最终失败消息和上传的文件事件也没有 `m.mentions`。接收端 DSH Agent 会丢弃没有结构化 mention 的 Agent 消息，因此这些回复可能被静默忽略。建议回复 Agent 消息时自动 mention 原发送者，并统一覆盖成功文本、最终失败和文件事件；同时补充一条不要求模型复述 Leader ID 的 E2E。
2. **独立发布工作流仍在校验前把 `inputs.version` 和 `github.ref_name` 直接插入 Bash。** 建议通过 step-level `env` 传入，再严格引用和校验 Shell 变量，镜像构建步骤也采用相同方式。
3. **当前可选择 E2EE，但 Bridge 不支持。** 开启 `AGENTTEAMS_MATRIX_E2EE=1` 后房间事件为 `m.room.encrypted`，DSH Bridge 只处理 `m.room.message`。建议拒绝或门禁该组合、启动时明确失败，或者补齐 E2EE 实现与测试。
4. **固定的默认镜像尚不可交付。** 实时查询 `higress-registry.cn-hangzhou.cr.aliyuncs.com/agentteams/agentteams-deepseek-harness-worker:v0.1.0` 仍返回 `not found`。在默认配置依赖它之前，需要发布并确认 `linux/amd64`、`linux/arm64` manifest 均存在。

当前 GitHub Checks 均为绿色。本地聚焦的 Go、安装器和 Node 测试通过；Python 通过 30/31，剩余一项是 macOS `/tmp` 与 `/private/tmp` 路径别名导致的测试断言失败。

oss-maintainer (commented):
### Summary
This PR adds an experimental DeepSeek Harness worker runtime with a Dockerfile, entrypoint, Python Matrix bridge (channel loop, session state, file exchange), and controller/CRD/backend wiring. The controller-side changes are clean: UsesMemberRuntimeConfig correctly generalizes the QwenPaw config path, all three backends (docker/kubernetes/sandbox) handle the new runtime consistently, the CRD enum is extended backward-compatibly, and test coverage is thorough. The bridge scripts are well-structured with atomic state writes, idempotent Matrix delivery via deterministic transaction IDs, room-scoped session continuation, and bounded completed-event pruning. The main concern is defense-in-depth: run_dsh copies the full container environment (including storage credentials and Matrix token) into the DSH subprocess, when DSH only needs DEEPSEEK_API_KEY and TEAMHARNESS_* variables — the MCP server legitimately needs the storage/Matrix credentials but receives them through its own cordis.patch.yml env mapping, so a tighter environment construction for DSH would reduce the credential surface visible to the DSH process. The health endpoint binding to 0.0.0.0 is standard for containers but could be tightened to 127.0.0.1. Overall the PR is well-engineered with good error handling, restart recovery, and test coverage; the credential isolation improvement is worth considering before merging.

---

deepseek-harness/scripts/matrix_bridge.py
```diff
@@ -0,0 +1,503 @@
...
+    workspace: Path,
+    timeout_seconds: int,
+    *,
+    session_id: str,
+    resume: bool,
+    event_id: str,
+    attempt: int,
+) -> str:
+    environment = os.environ.copy()
+    environment.update(
```
oss-maintainer, line 342: `run_dsh` copies the full process environment (`os.environ.copy()`) into the DSH subprocess. This includes `AGENTTEAMS_FS_ACCESS_KEY`, `AGENTTEAMS_FS_SECRET_KEY`, `AGENTTEAMS_WORKER_MATRIX_TOKEN`, and `AGENTTEAMS_WORKER_GATEWAY_KEY` — credentials that DSH itself does not need (only the MCP server subprocess does, via `cordis.patch.yml` env mapping). Consider constructing a minimal environment for DSH (keeping only `DEEPSEEK_API_KEY`, `DEEPSEEK_BASE_URL`, `DSH_HOME`, `TEAMHARNESS_*`, `PATH`, and locale variables) so that a compromised or buggy DSH plugin cannot read storage or Matrix credentials from its own `/proc/self/environ`.

deepseek-harness/scripts/deepseek-harness-worker-entrypoint.sh
```diff
@@ -0,0 +1,78 @@
...
+mkdir -p "${WORKER_HOME}" "${RUNTIME_DIR}"
+export HOME="${WORKER_HOME}"
+export DSH_HOME="${WORKER_HOME}/.dsh"
+export TEAMHARNESS_RUNTIME_CONFIG="${RUNTIME_CONFIG}"
+export TEAMHARNESS_WORKSPACE="${WORKER_HOME}/workspace"
+export TEAMHARNESS_DSH_SKILL_ROOT="${RUNTIME_DIR}/dsh-skills"
+export TEAMHARNESS_PYTHON="/usr/bin/python3"
+export AGENTTEAMS_PLUGIN_DIR="/opt/agentteams/plugins/teamharness"
+export AGENTTEAMS_MATRIX_USER_ID="@${WORKER_NAME}:${AGENTTEAMS_MATRIX_DOMAIN}"
+export DEEPSEEK_API_KEY="${DEEPSEEK_API_KEY:-${AGENTTEAMS_WORKER_GATEWAY_KEY:-}}"
```
oss-maintainer: `export DEEPSEEK_API_KEY="${DEEPSEEK_API_KEY:-${AGENTTEAMS_WORKER_GATEWAY_KEY:-}}"` silently sets an empty `DEEPSEEK_API_KEY` when both variables are unset. DSH will then fail with an opaque authentication error from the gateway. Consider adding an explicit guard (`if [ -z "${DEEPSEEK_API_KEY}" ]; then log "ERROR: no LLM credential available"; exit 1; fi`) so the failure message identifies the missing credential rather than surfacing a generic HTTP 401/403 from the model provider.

deepseek-harness/scripts/matrix_bridge.py
```diff
@@ -0,0 +1,503 @@
...
+        self.wfile.write(body)
+
+    def log_message(self, _format: str, *_args: object) -> None:
+        return
+
+
+def serve_health() -> None:
+    port = int(os.getenv("AGENTTEAMS_CONSOLE_PORT", "8088"))
+    server = ThreadingHTTPServer(("0.0.0.0", port), HealthHandler)
+    server.timeout = 1
```
oss-maintainer, line 387: The health endpoint binds to `0.0.0.0:8088`. While this is standard for container health checks, the server accepts arbitrary paths and returns 404 for anything other than `/healthz`. If the Worker Pod is reachable from outside the cluster (e.g., via a misconfigured NetworkPolicy), any client can probe the port. Binding to `127.0.0.1` would be sufficient for kubelet probes and would eliminate the external surface.

deepseek-harness/scripts/bridge_state.py
```diff
@@ -0,0 +1,145 @@
...
+    def mark_completed(self, event_id: str, reply_event_id: str) -> None:
+        record = self.data["events"].setdefault(event_id, {})
+        record.update(
+            {
+                "status": "completed",
+                "reply_event_id": reply_event_id,
+                "updated_at": int(time.time()),
+            }
+        )
+        completed = sorted(
```
oss-maintainer, line 128: `mark_completed` iterates all events, filters completed ones, and sorts them by `updated_at` on every call. With the 2000-entry cap this is bounded, but the sort is O(n log n) per event completion. Since completions are infrequent (one per Matrix message), this is acceptable, but consider moving the pruning to a periodic background task or only pruning when the event count exceeds a threshold to avoid redundant work.

agentteams-controller/internal/controller/member_reconcile.go
```diff
@@ -395,7 +395,7 @@ func ReconcileMemberConfig(ctx context.Context, d MemberDeps, m MemberContext, s
 		aiGatewayURL = m.ModelProviderInfo.IntranetURL
 	}
 
-	if effectiveRuntime == backend.RuntimeQwenPaw || m.DeployMode == v1beta1.DeployModeEdge {
+	if backend.UsesMemberRuntimeConfig(effectiveRuntime) || m.DeployMode == v1beta1.DeployModeEdge {
```
oss-maintainer, line 398: The `UsesMemberRuntimeConfig` abstraction correctly generalizes the QwenPaw-specific config path to include DeepSeek Harness. All reconcile paths (member_reconcile, team_controller, worker_controller) are updated consistently, and the hash functions are renamed without changing their output for existing inputs. CRD enum extension (`deepseek-harness` added) is backward compatible. Backend implementations (docker, kubernetes, sandbox) all handle the new runtime with the same pattern as existing runtimes. Test coverage is thorough.

jesseedcp:
Addressed this follow-up in `e6ad4acf`, `ef220303`, and `58e7e130`.

Replies now target the source Agent from the inbound event instead of relying on the model to repeat a Matrix ID. The same structured mention is carried on success text, terminal failures, and returned files. I added a small marker for automatic source mentions so the target receives the reply once without two Agents bouncing it back and forth. The Team E2E no longer asks the Worker to echo the Leader ID; it passed Worker-to-Leader delivery, no-bounce, and unmentioned Team Room chatter staying quiet.

The release workflow now passes version/ref values through step environment variables, validates quoted shell variables, and checks the published manifest for both `linux/amd64` and `linux/arm64`. E2EE and missing model credentials fail fast at startup. I also fixed the macOS `/tmp` alias assertion.

Locally, 42 Python tests and 2 Node tests pass, along with actionlint, shellcheck, the focused Go packages, AMD64/ARM64 image builds, and the full Kind Team E2E.

The official `v0.1.0` manifest is the one item I cannot publish from the contributor fork: that needs the upstream registry credentials and the dedicated workflow to be available on the upstream side. The workflow is ready to verify both platforms immediately after publishing.

I also checked the automated environment/health suggestions. The DSH parent still needs the Matrix and MinIO values because `cordis.patch.yml` resolves them from that environment when it creates the TeamHarness MCP child; stripping them there would break the adapter. Binding health to loopback would also break Kubernetes probes that hit the Pod IP, so I left both unchanged. The missing-credential startup guard was added.

CI note: the first GitHub run passed every DSH, Controller, build, and four integration jobs; one CoPaw shard failed because its post-delete Higress query was temporarily unavailable. Contributors cannot use the upstream rerun button, so I pushed the empty `ed4c634a` commit to start a clean run. That run passed the original CoPaw shard and four of five integration shards. The remaining QwenPaw shard failed in `test-19-human-and-team-admin`: the Team lookup returned 404 even though both member Workers had already been provisioned. That test passed in the first run and in the sibling runtime shards, and none of the DSH changes touch that path. A maintainer rerun of the failed job should be enough; I am not going to keep adding empty commits to roll the full matrix again.

deepseek-harness/scripts/deepseek-harness-worker-entrypoint.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+set -euo pipefail
+
+if [ "${AGENTTEAMS_MATRIX_E2EE:-0}" = "1" ]; then
```
oss-maintainer: The E2EE guard recognizes only `AGENTTEAMS_MATRIX_E2EE=1`, but the Manager startup also treats `true` as enabled. A DSH Worker configured with the supported `true` value will therefore start its plaintext Matrix bridge against encrypted rooms and silently ignore `m.room.encrypted` traffic instead of failing fast as documented. Normalize the boolean or explicitly accept both `1` and `true`, and add coverage for the latter.

oss-maintainer (commented):
### Summary
Re-review of the 20 commits pushed since the last review (8b011039..ed4c634a): Matrix mention-routing fixes, turn-identity recovery, release-workflow hardening, and expanded test coverage (+372/-36 across 15 files). The incremental changes are coherent and well tested overall; one E2EE-guard inconsistency should be addressed before approval.

### Findings
- **[Warning]** deepseek-harness/scripts/deepseek-harness-worker-entrypoint.sh:4 — the E2EE guard only recognizes `AGENTTEAMS_MATRIX_E2EE=1`, while the controller (`agentteams-controller/internal/config/config.go`) and the manager startup scripts also treat `true` as enabled. A DSH worker deployed under a `true`-enabled E2EE configuration would bypass this guard and start its plaintext Matrix bridge against encrypted rooms instead of failing fast.

### Suggestions
Normalize the guard to accept both supported values, mirroring the existing pattern in `manager/scripts/init/start-manager-agent.sh`:

```bash
if [ "${AGENTTEAMS_MATRIX_E2EE:-0}" = "1" ] || [ "${AGENTTEAMS_MATRIX_E2EE:-}" = "true" ]; then
```

and add an entrypoint test case for `AGENTTEAMS_MATRIX_E2EE=true` alongside the existing `=1` case in `deepseek-harness/tests/test_entrypoint.py`.

---

jesseedcp:
Small follow-up in `35302ccc`: the startup guard now rejects both supported E2EE values, `1` and `true`, and the new test covers the `true` case. The focused suite is green locally (43 Python tests plus shellcheck). The fresh GitHub DSH, Controller, Helm, and CRD jobs are green too; image and integration jobs are still running.

oss-maintainer (approved):
### Summary
Incremental re-review of the single new commit pushed since the last review (35302ccc "fix(dsh): reject true-valued Matrix E2EE"). The previously flagged E2EE-guard inconsistency is resolved: the entrypoint now rejects both `AGENTTEAMS_MATRIX_E2EE=1` and `true` before startup, matching the controller config parsing semantics, and the new test in `deepseek-harness/tests/test_entrypoint.py` covers the true-valued case with a nonzero-exit assertion. No regression found in this delta.

### Findings
None — no new issues in the incremental change.

---

shiyiyue1102 (approved):
LGTM. The previously reported Matrix routing, release-workflow input handling, and E2EE fail-fast gaps are fixed on the current head, with focused local tests and the full GitHub check suite passing. The v0.1.0 runtime image will be published through the dedicated release flow.

---

LGTM。此前提出的 Matrix 路由、发布工作流输入处理和 E2EE 启动门禁问题均已在当前 head 修复；聚焦本地测试及 GitHub 全量检查均已通过。v0.1.0 Runtime 镜像按独立正式发布流程产出。

Issue #63, opened 2026-03-06
群聊消息无回复


luodadaRyan:
按照提示操作，在群里发消息没有得到回复，不确定哪里还有问题

<img width="2348" height="829" alt="Image" src="https://github.com/user-attachments/assets/dd0689bf-a0ca-4db7-a5ab-1a546af6cf9f" />

shiyiyue1102:
群聊消息无回复先按 Matrix 房间触发链路处理。@Sunrisea 麻烦看下是 mention、allowFrom 还是房间成员关系导致消息没有进到 runtime。

---
If there is no reply to the group chat message, first press the Matrix room to trigger the link. @Sunrisea Please check whether it is mention, allowFrom or room membership that causes the message not to enter the runtime.

PR #1213, opened 2026-09-01
fix(agentconfig): emit agents.list so group-chat @mention triggers reply

gavin-ygy:
OpenClaw derives group-chat mention regexes from agents.list[].identity.name, but GenerateOpenClawConfig only emitted agents.defaults. With an empty agents.list, buildMentionRegexes produced no patterns, wasMentioned was always false in group rooms, and the Matrix monitor dropped every room message with "skipping room message (reason: no-mention)". DMs are unaffected because DM delivery does not require a mention.

Emit agents.list with a single "main" entry whose name/identity.name is the runtime name (req.WorkerName). Apply the same fix to the Manager template (manager-openclaw.json.tmpl), which had the same gap.

Add a regression assertion in TestGenerateOpenClawConfig_Basic.

Fixes the group-chat no-reply reported in #63.

### DIFF
--- a/agentteams-controller/internal/agentconfig/generator_test.go
+++ b/agentteams-controller/internal/agentconfig/generator_test.go
@@ -61,6 +61,21 @@ func TestGenerateOpenClawConfig_Basic(t *testing.T) {
 	if modelCfg["primary"] != "agentteams-gateway/qwen3.5-plus" {
 		t.Errorf("agents.defaults.model.primary = %v, want agentteams-gateway/qwen3.5-plus", modelCfg["primary"])
 	}
+
+	// Verify agents.list carries the worker identity so OpenClaw can build
+	// group @mention regexes (regression guard for group-chat no-reply issue).
+	list, ok := agents["list"].([]interface{})
+	if !ok || len(list) != 1 {
+		t.Fatalf("agents.list missing or wrong shape: %v", agents["list"])
+	}
+	listEntry := list[0].(map[string]interface{})
+	identity := listEntry["identity"].(map[string]interface{})
+	if listEntry["id"] != "main" {
+		t.Errorf("agents.list[0].id = %v, want main", listEntry["id"])
+	}
+	if identity["name"] != "worker-alice" {
+		t.Errorf("agents.list[0].identity.name = %v, want worker-alice", identity["name"])
+	}
 }
 
 func TestGenerateOpenClawConfig_TeamWorker(t *testing.T) {
--- a/agentteams-controller/internal/agentconfig/generator.go
+++ b/agentteams-controller/internal/agentconfig/generator.go
@@ -134,6 +134,15 @@ func (g *Generator) GenerateOpenClawConfig(req WorkerConfigRequest) ([]byte, err
 			},
 		},
 		"agents": map[string]interface{}{
+			"list": []interface{}{
+				map[string]interface{}{
+					"id":       "main",
+					"name":     req.WorkerName,
+					"identity": map[string]interface{}{
+						"name": req.WorkerName,
+					},
+				},
+			},
 			"defaults": map[string]interface{}{
 				"timeoutSeconds": 1800,
 				"workspace":      "~",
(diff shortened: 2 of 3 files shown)


agentteams-controller/internal/agentconfig/generator_test.go
```diff
@@ -61,6 +61,21 @@ func TestGenerateOpenClawConfig_Basic(t *testing.T) {
...
+	list, ok := agents["list"].([]interface{})
+	if !ok || len(list) != 1 {
+		t.Fatalf("agents.list missing or wrong shape: %v", agents["list"])
+	}
+	listEntry := list[0].(map[string]interface{})
+	identity := listEntry["identity"].(map[string]interface{})
+	if listEntry["id"] != "main" {
+		t.Errorf("agents.list[0].id = %v, want main", listEntry["id"])
+	}
+	if identity["name"] != "worker-alice" {
```
oss-maintainer, line 76: **[Warning]** `listEntry["identity"].(map[string]interface{})` is an unchecked type assertion — if the emitted shape is wrong the test panics instead of failing with a readable message. A two-value assertion (or a prior `t.Fatalf` guard, as used for `list` above) would keep the regression guard robust.

agentteams-controller/internal/agentconfig/generator_test.go
```diff
@@ -61,6 +61,21 @@ func TestGenerateOpenClawConfig_Basic(t *testing.T) {
...
+		t.Errorf("agents.list[0].identity.name = %v, want worker-alice", identity["name"])
+	}
 }
 
 func TestGenerateOpenClawConfig_TeamWorker(t *testing.T) {
```
oss-maintainer, line 81: **[Info]** `TestGenerateOpenClawConfig_TeamWorker` could use the same `agents.list[0].identity.name == req.WorkerName` assertion, since the generator emits `agents.list` for team workers too — this would guard the team/leader path against the same regression.

oss-maintainer (commented):
### Summary
This PR makes `GenerateOpenClawConfig` and `manager-openclaw.json.tmpl` emit `agents.list` with a `main` entry carrying the runtime identity, fixing the group-chat no-reply issue (#63) where `buildMentionRegexes` produced no patterns. The core fix is correct and well scoped, and the new unit test guards the generated shape for the basic worker path. However, two parallel config-generation paths are not covered, so the same bug persists for them — worth addressing before merge. Overall verdict: the controller-provisioned Worker and Manager paths are safe; extend the fix to fully close #63.

### Suggestions
- Extend the fix to the static Worker template `manager/agent/skills/worker-management/references/worker-openclaw.json.tmpl` (rendered by `generate-worker-config.sh`), adding the same `agents.list` block from the worker name, e.g.:
```json
"agents": {
  "list": [
    { "id": "main", "name": "${WORKER_NAME}", "identity": { "name": "${WORKER_NAME}" } }
  ]
}
```
- Inject `agents.list` in the existing-config upgrade branch of `start-manager-agent.sh` so upgraded Managers don't keep the bug.

---

### Additional notes (not anchored to a changed line)
- **[CRITICAL]** `manager/agent/skills/worker-management/references/worker-openclaw.json.tmpl:67` — This static Worker config template is still missing `agents.list`, so Workers created via `generate-worker-config.sh` (which renders this template) will not have the identity entry OpenClaw needs to build group-chat mention regexes — the same bug this PR fixes would persist on that path. The script already has the worker name available; add an `agents.list` block here with `id`/`name`/`identity.name` set from the worker name to match the generator fix. _(line outside diff)_
- **[WARNING]** `manager/scripts/init/start-manager-agent.sh:738` — When `/root/manager-workspace/openclaw.json` already exists, the upgrade `jq` pass only mutates `agents.defaults` and `channels.matrix`; it never injects `agents.list`. Managers upgraded from a pre-fix install therefore keep the group-chat no-mention bug until the config is regenerated from the template. Consider injecting `agents.list` in the existing-config upgrade branch. _(line outside diff)_

PR #1208, opened 2026-08-31
feat(controller): add team-scoped worker knowledge base file endpoints (+ checkpoint scoped-access fix)

LUOSENGWA:
# Team-scoped worker knowledge base file endpoints: read, write, download

## Summary

L2 humans (and team leaders) currently have **no path to their team's knowledge base** — not read, not write. The only way to reach a worker's workspace files is the Controller's `/docker/` proxy, which is admin/Manager-only: an L2 human's Matrix token gets `403` on every attempt, and the checkpoint endpoints (#1186) expose file *indexes* and change summaries, not the knowledge text itself.

This PR proxies four QwenPaw workspace-file endpoints through the Controller, following the exact fixed-subpath proxy pattern established by the worker checkpoint endpoints (#1186):

- `GET /api/v1/workers/{name}/workspace-files/tree?path=memory|digest[/...]` — paginated directory listing of the knowledge trees
- `GET /api/v1/workers/{name}/workspace-files/file-metadata?path=...` — size/etag/modified of one knowledge file
- `GET /api/v1/workers/{name}/workspace-files/file-content?path=...[&offset=&limit=]` — bounded UTF-8 chunk read with `next_offset` continuation
- `PUT /api/v1/workers/{name}/workspace-files/file-content?path=...` — ETag-guarded save of one knowledge file (create or update)
- `GET /api/v1/workers/{name}/workspace-files/file-download?path=...` — bounded stream of one knowledge file with attachment headers

The addressable surface is a **path allowlist, identical for reads and writes**: `MEMORY.md` (long-term memory), `memory/**` (daily notes) and `digest/**` (distilled knowledge). Everything else in the workspace — `SOUL.md`, `PROFILE.md`, `TODO.md`, `checkpoints/`, `skills/`, and all dot directories (`.copaw/agent.json` carries the worker's Matrix token and storage credentials) — is rejected with `400` **before the request reaches the worker**, even though the upstream path resolver would serve a directly requested dot path.

[...]

PR #1186, opened 2026-08-16
feat(projects): add intervention history and worker checkpoint read endpoints

LUOSENGWA:
# Projects: intervention history + worker checkpoint read endpoints

## Summary

Adds read endpoints for two complementary timelines: project intervention history and worker execution checkpoints.

- **Project history** — the write side landed in agentteams/AgentTeams#1172 (every intervention snapshots the pre-change meta into `history/{unixNano}.json`, gc keeps 50), but there is no way to read the timeline back. This PR adds `GET /projects/{id}/history` (snapshot list, newest first) and `GET /projects/{id}/history/{timestamp}` (one snapshot verbatim).
- **Worker checkpoints** — QwenPaw 2.1 ships a checkpoint system (framework hook auto-snapshots after every response round + manual `/checkpoint snapshot`, stored in `checkpoints/shadow.git`) exposed on each worker's qwenpaw app at `:8088`. The Controller proxies two read-only subpaths — `GET /workers/{name}/checkpoints/graph` and `/status` — so L2 humans and frontends can inspect every worker's execution timeline.

## What's included

1. `agentteams-controller/internal/server/project_handler.go`:
   - `resolveSingleProjectMatch` — same 0/1/many resolution as `resolveSingleProject` but returns the whole match (including the storage key), so history handlers can derive the sibling `history/` prefix (reused verbatim from the write side).
   - `GetProjectHistory` — lists snapshots under `history/`, newest first; timestamps serialized as **strings** (19-digit nanoseconds exceed the JS safe-integer range); defensive filter drops unexpected files; empty history returns `"snapshots": []`.
   - `GetProjectHistorySnapshot` — returns one snapshot's raw meta JSON verbatim (read side stays decoupled from the meta schema); timestamp must match `^[0-9]{19}$` (doubles as the traversal guard).
2. `agentteams-controller/internal/server/worker_checkpoints.go` (new):
[...]

PR #1214, opened 2026-09-01
fix(controller): checkpoint proxy scoped check compared member name as team name

LUOSENGWA:
# What

4-line fix in the worker checkpoint proxy (`agentteams-controller/internal/server/worker_checkpoints.go`): the scoped-caller check now compares the caller's team against the **Team CR name** (`findTeamMember`'s first return value) instead of the **member (worker) name** (its second return value). Plus one regression test.

# Why

`findTeamMember` returns `(team *Team, memberName string, isLeader bool, err error)`. The checkpoint proxy used `memberName` in `caller.TeamMatches(...)` — but `TeamMatches` expects the **team** name. For any real team the worker's member name differs from the team name, so the check was **always false**: every team leader and every L2 human received `404` when reading checkpoints of their own team's workers, even though `#1186` documented that exact access path as supported. Cross-team callers also got `404`, so the suite (which only asserted cross-team `404`) passed for the wrong reason and never noticed.

Effect: the documented L2/leader checkpoint read path silently did not work for the roles it was built for.

# Fix

- Compare against `teamObj.Name` (first return value) — the same resolution chain as `ResourceHandler.GetWorker` and the new workspace-files proxy.
- All six `findTeamMember` call sites were audited: this was the only one misusing the member-name return value.
- Behavior change is strictly additive for in-scope callers (they now get the documented `200`); cross-team/unknown callers are unchanged (`404`, W8 anti-probing preserved).

# Test

- `TestCheckpoint_L2HumanInScopeAllowed` (new): an L2 human in the worker's own team reads its checkpoints → `200`. Before the fix this returned `404`.
- Existing cross-team `404` and leader cases still pass.
- Full `go test ./...` green on the `main` baseline (the only failure is a pre-existing environment issue in `internal/executor`: the sandbox lacks the `unzip` binary, which CI has).

# Scope

- Pure controller-side authorization logic; no worker-app API is involved, so there is **no runtime version gate** (applies to workers on any runtime — the checkpoint proxy is runtime-agnostic).
- Split out of the workspace-files PR (`#1208`) per review, to keep that PR focused on the storage/authorization feature.

---

## 摘要

worker checkpoint 代理（#1186）的 scoped 调用方检查把调用方团队与 `findTeamMember` 的**第二个返回值（成员/worker 名）**而非**团队名**做比较，导致 `TeamMatches` 恒为 false：所有团队 leader / L2 人类读自己团队 worker 的 checkpoints 一律 `404`——#1186 文档承诺的访问路径对目标角色实际不工作。跨团队调用方同样拿 `404`，而原测试套件只断言跨团队 `404`（在 bug 存在时也"恰好"通过），因此从未暴露。

修复：改与第一个返回值（Team CR 名）比较——与 `ResourceHandler.GetWorker` 同款链路。全量审计 6 个 `findTeamMember` 调用点，仅此处误用成员名返回值。行为变化对 in-scope 调用方纯增益（从错误的 `404` 变为文档承诺的 `200`），跨团队/未知调用方不变（`404`，W8 防探测保持）。

新增回归测试 `TestCheckpoint_L2HumanInScopeAllowed`。纯 controller 侧授权逻辑，不涉及 worker app API，无 runtime 版本门。按 review 要求从 #1208 拆出，保持该 PR 聚焦。

### DIFF
--- a/agentteams-controller/internal/server/worker_checkpoints.go
+++ b/agentteams-controller/internal/server/worker_checkpoints.go
@@ -134,15 +134,21 @@ func (h *CheckpointHandler) proxyCheckpoint(w http.ResponseWriter, r *http.Reque
 		return
 	}
 	// Resolve the owning team for the scoped-caller check (same chain as
-	// ResourceHandler.GetWorker: standalone workers hide as 404).
-	_, team, _, err := findTeamMember(r.Context(), h.client, h.namespace, name)
+	// ResourceHandler.GetWorker: standalone workers hide as 404). Note:
+	// findTeamMember's second return value is the member (worker) name,
+	// not the team name — the check must compare against the Team CR name.
+	teamObj, _, _, err := findTeamMember(r.Context(), h.client, h.namespace, name)
 	if err != nil {
 		writeK8sError(w, "get worker checkpoints", err)
 		return
 	}
+	teamName := ""
+	if teamObj != nil {
+		teamName = teamObj.Name
+	}
 	if caller := authpkg.CallerFromContext(r.Context()); caller != nil &&
 		(caller.Role == authpkg.RoleTeamLeader || caller.Role == authpkg.RoleHuman) &&
-		!caller.TeamMatches(team) {
+		!caller.TeamMatches(teamName) {
 		httputil.WriteError(w, http.StatusNotFound, "worker not found")
 		return
 	}
--- a/agentteams-controller/internal/server/worker_checkpoints_test.go
+++ b/agentteams-controller/internal/server/worker_checkpoints_test.go
@@ -236,6 +236,27 @@ func TestCheckpoint_TeamLeaderCrossTeamDenied(t *testing.T) {
 	}
 }
 
+// TestCheckpoint_L2HumanInScopeAllowed locks the scoped-caller fix:
+// findTeamMember's second return value is the member (worker) name, not
+// the team name, so the in-scope check must compare against the Team CR
+// name — an in-scope L2 human must resolve 200.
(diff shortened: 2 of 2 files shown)


shiyiyue1102 (approved):
LGTM. This focused fix now compares scoped callers against the owning Team CR name instead of the Worker member name, and the new in-scope L2 regression test covers the previously broken path. CI is green.

---

LGTM。这个聚焦修复改为使用所属 Team CR 名称进行 scoped caller 判断，不再误用 Worker 成员名；新增的本团队 L2 回归测试也覆盖了此前失效的访问路径。CI 已通过。

oss-maintainer (approved):
LGTM. Verified the fix against the current codebase:

- `findTeamMember` returns `(*v1beta1.Team, memberName string, isLeader bool, err)` — the old code destructured the **second** return value (member/worker name) and passed it to `caller.TeamMatches(...)`, which expects a team name. Since a worker's member name practically never equals its Team CR name, the scoped check was effectively always false for in-scope leaders/L2 humans.
- The new code uses `teamObj.Name` (first return value) with a proper nil guard: standalone workers (`teamObj == nil`) resolve to `teamName == ""`, and `TeamMatches("")` returns false, so the documented "standalone workers hide as 404" behavior is preserved.
- The new `TestCheckpoint_L2HumanInScopeAllowed` regression test covers exactly the previously-broken path (in-scope L2 human → 200), complementing the existing cross-team 404 cases.
- Consistent with the resolution chain used by `ResourceHandler.GetWorker`. Trivial, well-scoped change.

---

PR #1225, opened 2026-09-04
fix(manager): restore worker settings after restart

### LINKED ISSUE #575: 升级后，worker容器不能恢复原来的设置 || After the upgrade, the worker container cannot restore its original settings.
### Bug Description

正常升级，当worker容器重建后，发现原本的部分配置无法恢复，让manager调查问题如下：

Image

### Steps to Reproduce

HICLAW_VERSION=v1.0.9 bash <(curl -sSL https://higress.ai/hiclaw/install.sh) 

### AI Analysis

_No response_

### Relevant Logs

```shell

```

### Component

Manager Agent

### Version / Commit

_No response_

---
### Bug Description

After the normal upgrade, after the worker container was rebuilt, it was found that some of the original configurations could not be restored. Let the manager investigate the problem as follows:

Image

### Steps to Reproduce

HICLAW_VERSION=v1.0.9 bash <(curl -sSL https://higress.ai/hiclaw/install.sh)

### AI Analysis

_No response_

### Relevant Logs

```shell

```

### Component

Manager Agent

### Version/Commit

_No response_

014-code:
Closes #575

## Summary

- Remove the Manager startup legacy Worker recreation loop; controller reconciliation owns backend recovery.
- Preserve Worker.spec.env when the controller recreates a missing backend.
- Add Manager startup regression coverage and an integration assertion for settings preservation.

## Root cause

Manager rebuilt existing Workers from persisted credentials only, which omitted user settings and could race controller reconciliation during restart or upgrade. The controller already has the complete Worker spec and is the correct owner for recovery.

## Validation

- bash manager/tests/test-manager-startup-recovery.sh
- bash -n manager/scripts/init/start-manager-agent.sh
- go test ./test/integration/controller (envtest)
- Targeted Go tests: internal/controller, internal/server, internal/service, cmd/agt

### DIFF
--- a/manager/tests/test-manager-startup-recovery.sh
+++ b/manager/tests/test-manager-startup-recovery.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# Regression tests for Manager startup Worker recovery ownership.
+
+set -uo pipefail
+
+PASS=0
+FAIL=0
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+STARTUP_SCRIPT="${PROJECT_ROOT}/manager/scripts/init/start-manager-agent.sh"
+
+pass() { echo "  PASS: $1"; PASS=$((PASS + 1)); }
+fail() { echo "  FAIL: $1"; echo "       expected: $2"; echo "       got:      $3"; FAIL=$((FAIL + 1)); }
+
+assert_contains() {
+    local desc="$1" needle="$2" haystack="$3"
+    if printf '%s\n' "${haystack}" | grep -qF -- "${needle}"; then
+        pass "${desc}"
+    else
+        fail "${desc}" "contains '${needle}'" "not found"
+    fi
+}
+
+assert_not_contains() {
+    local desc="$1" needle="$2" haystack="$3"
+    if ! printf '%s\n' "${haystack}" | grep -qF -- "${needle}"; then
+        pass "${desc}"
+    else
+        fail "${desc}" "does not contain '${needle}'" "found '${needle}'"
+    fi
+}
+
+startup_source="$(<"${STARTUP_SCRIPT}")"
+
+echo "=== Manager startup Worker recovery ownership ==="
+assert_contains "startup still detects the configured container runtime" \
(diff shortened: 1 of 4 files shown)


manager/tests/test-manager-startup-recovery.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
oss-maintainer, line 1: This regression script is not invoked by any Makefile target or GitHub workflow (the integration runner only executes `tests/test-*.sh`, and the Manager image does not copy `manager/tests/`). As written, it will never run in CI — consider wiring it into an existing test target or workflow so the recovery-ownership invariant stays enforced.

oss-maintainer (commented):
﻿### Summary
Removes the Manager startup legacy Worker recreation loop (controller reconciliation owns backend recovery) and adds regression coverage; core change is sound — verified against main: the controller's backend recreation already merges `Worker.spec.Env` into the backend Create request, and the integration mock exposes `FindCreateReq` with the recorded Env, so the new assertion is valid. One wiring gap noted below.

### Findings
- **[Warning]** manager/tests/test-manager-startup-recovery.sh:1 — the new regression script is not invoked by any Makefile target or GitHub workflow, so it never runs in CI

### Suggestions
Wire `manager/tests/test-manager-startup-recovery.sh` into an existing test target/workflow so the "controller owns Worker recovery" invariant is actually enforced on future changes.

---

oss-maintainer (approved):
﻿LGTM — the follow-up commit (1f640c2) addresses the earlier CI-wiring feedback: `manager/tests/test-manager-startup-recovery.sh` is now invoked by the Helm Lint workflow, with both the startup script and the regression test included in its `pull_request` path filters.

Verified against `main`:
- Controller backend recreation passes `Worker.spec.Env` into the backend Create request (`internal/controller/member_reconcile.go`), so recovery owned by the controller preserves worker settings; the new `TestWorkerPodDeleted_Recreates` assertion (`FindCreateReq` + env check) covers this.
- Docker local deployment still runs the `agentteams-controller` container, so removing the Manager-startup recreation loop leaves recovery coverage in socket mode.
- The regression script itself is a sound static guard (grep assertions + `bash -n`) for the startup path.

Note for maintainers: the follow-up workflow runs on this fork PR currently sit in `action_required` and need approval by someone with repo admin rights, which I don't have.

---

PR #1231, opened 2026-09-07
feat(controller): add team-scoped QwenPaw worker runtime-config proxy + loop-change notification

LUOSENGWA:
# Team-scoped QwenPaw worker running-config proxy

> **Standalone** — rebased onto current `main`; the earlier stacked #1211 (skill catalog) commits are no longer in this diff. They remain tracked in #1211 (paused pending the #1221 Skill Center design discussion), and this PR contains only the runtime-config proxy + loop-change notification.

## Summary

There was no Controller API for reading or writing a QwenPaw worker's running config — an L2 human had to `docker exec` into the worker to change the workbench 5-tab settings (ReAct / Loop / LLM-retry / long-term-memory / tool-level). This PR adds proxy endpoints (mirroring the #1216 approval proxy):

```
GET/PUT  /api/v1/workers/{name}/runtime-config     (the 5-tab running-config)
GET      /api/v1/workers/{name}/loops              (Loop Engine catalog)
GET      /api/v1/workers/{name}/loops/status       (per-loop status)
GET      /api/v1/workers/{name}/loops/custom       (custom loop modes)
POST     /api/v1/workers/{name}/loops/custom
PUT      /api/v1/workers/{name}/loops/custom/{loop}
DELETE   /api/v1/workers/{name}/loops/custom/{loop}
```

**Upstream contract (verified against the pinned qwenpaw 2.0.1 wheel; still present in 2.2.x):** `GET/PUT /api/workspace/running-config` (active-agent scope — an AgentTeams worker is a single-agent qwenpaw app, so the app resolves the agent) plus the `/api/loops*` router.

- `PUT runtime-config` is **field-level**: the proxy GETs the current full object, merges only the provided top-level keys, and PUTs the merged full object back — the upstream PUT *replaces the entire running section*, so a bare partial forward would wipe unrelated settings. An update that changes nothing skips the upstream write (idempotent no-op, 200).
- When the **loop section** is changed (a runtime-config PUT whose keys include `loop`, or a `/loops/custom` CRUD write), the handler pushes a Matrix room message to the worker's team room **only after the upstream write succeeded (2xx)**: **@team-leader + @changer** (leader self-edit → leader only). A rejected write (409/422/404/5xx) never notifies.
- **`approval_level` is rejected via runtime-config for every role** (400) — it is owned by the #1216 endpoint `PUT /api/v1/workers/{name}/approval` (safety-write + audit semantics).

## Design

| Decision | Rationale |
|---|---|
[...]

PR #1253, opened 2026-09-15
fix(controller): forward the loops/status session selectors

LUOSENGWA:
The `/api/loops/status` upstream endpoint answers **per session**: without `chat_id` or `session_id` it always reports `{"state":"idle"}`. The proxy introduced in #1231 dropped the query string entirely, so the proxied endpoint could never report a real loop state for the workbench.

- GET `loops/status` now forwards the two allowlisted selectors (`chat_id` / `session_id`, normalized); unknown keys and query parameters on every other route stay rejected (strict-whitelist discipline, same as the checkpoint proxy).
- Contract notes updated (2.0.1 review base; unchanged under the 2.2.x pin the branch carries).
- New tests: `TestRuntimeConfig_LoopsStatusForwardsSessionQuery` (allowlisted forwarding) and `TestRuntimeConfig_UnknownQueryRejected` (unknown key + wrong route).

Follow-up to #1231.

---

`/api/loops/status` 上游端点按会话应答：无 `chat_id`/`session_id` 时恒返 `{"state":"idle"}`。#1231 的代理把 query 整体丢弃，代理后的端点永远无法反映真实 loop 状态。

- GET `loops/status` 现转发两个白名单选择器（`chat_id` / `session_id`，规范化排序）；未知键与其它路由上的 query 一律拒绝（严格白名单纪律，与 checkpoint 代理一致）。
- 契约注释更新（评审基线 2.0.1；分支现在的 2.2.x pin 下不变）。
- 新增测试两枚：白名单转发 + 未知键/错误路由拒绝。

#1231 的 follow-up。


shiyiyue1102 (approved):
LGTM. The proxy forwards only the supported loops/status session selectors and rejects unsupported query parameters. The targeted runtime-config tests passed locally and current CI checks passed.

PR #1259, opened 2026-09-16
fix(qwenpaw): make demo-agent disable non-fatal during startup

ChestonGu:
## Problem

QwenPaw returns HTTP 409 for `PATCH /agents/{id}/toggle` while the agent is still loading in the background (the `is_agent_startup_in_progress` gate — identical in 2.0.1 and 2.2.x). On slow-starting deployments the retry loop in `disable_agent_if_present` (120 × 1s) exhausts and raises; the startup error path in `run_qwenpaw` then terminates the qwenpaw process and re-raises, putting the worker container into CrashLoopBackOff. We observed all three pods of a team stuck in this loop.

## Fix

Disabling the built-in demo agent (`QwenPaw_QA_Agent_0.2`) is cosmetic — the worst outcome of a failed disable is one idle demo agent:

- Degrade the failure to a warning and continue startup
- Lower the 409 retry budget from 120 to 30 now that failure is no longer fatal

## Testing

Deployed on our test cluster: all qwenpaw worker pods now start 1/1 Running with no restarts; previously they cycled CrashLoopBackOff during the agent-loading window.


shiyiyue1102 (approved):
LGTM. The exception boundary is limited to disabling the optional demo agent; failures in required startup configuration still terminate startup. A temporary fault-injection test reproduces the fatal behavior on main and passes on this PR. The focused API/lifecycle suite passed (55 passed, 1 skipped).

The broader local test run retains the same workflow-configuration assertion failure seen on main; Matrix plugin direct-import tests were excluded. CI image builds and all five enabled integration shards now pass. The Hermes shard initially failed the runtime-switch migration-state assertions and passed on one unchanged-head rerun; this remains a potential timing issue, not a claim that the test flake was fixed. Secret-dependent shards are excluded by the fork workflow.

---

审核通过。异常处理仅覆盖可选 demo agent 的禁用，必要启动配置失败仍会终止启动。临时故障注入用例在 main 上复现退出、在本 PR 上通过；API/生命周期测试为 55 通过、1 跳过。

扩大本地测试范围后，保留了 main 同样存在的工作流配置断言失败，未纳入 Matrix 插件直接导入测试。CI 镜像构建及 5 个启用的集成分组均已通过；Hermes 分组首次迁移状态断言失败，在相同提交上重跑一次通过，仍需视为潜在时序波动，不能据此声称已修复该测试。fork 工作流未执行依赖密钥的分组。

Issue #1262, opened 2026-09-16
TeamHarness complete_project can mark projects completed with non-terminal tasks


apricity12345:
### What happened
On AgentTeams v1.2.3, TeamHarness's `projectflow complete_project` action can mark a project as `completed` even when its DAG or Loop still contains non-terminal tasks such as `planned`.

The current implementation updates the project status without validating that every task graph node has reached a terminal status.

### What I expected
`complete_project` should reject the action unless every DAG/Loop task is in a terminal status. A project should not become terminal while some tasks are still schedulable.

### Steps to Reproduce

1. Create a TeamHarness project with a DAG task.
2. Leave the task in `planned` status.
3. Call the `projectflow` action `complete_project` for that project.
4. Observe that the result is successful and the project status becomes `completed`, even though the task is still non-terminal.

### AI Analysis

Root cause: TeamHarness's project completion path updates project status directly, but does not enforce the aggregate invariant over the task graph.

- The implementation sets `project["status"] = "completed"` and updates the Loop status if present.
- It does not inspect `project["tasks"]` or `project["loop"]["tasks"]` before completing the project.
- The Controller project API already requires all tasks to be terminal before completion, so TeamHarness and Controller can currently disagree on the same project state.

Suggested fix:

- Before writing `completed`, inspect the active DAG or Loop task graph.
- If any task is non-terminal, return an error listing the task IDs and statuses.
- Use the same terminal-status set as the Controller project API, e.g. `completed`, `revision`, `blocked`, and `cancelled`.

### Relevant Logs

```shell
No runtime logs attached. This is reproducible directly at the code level: a TeamHarness project with one `planned` task can be completed successfully through `projectflow complete_project`.
```

### Component

Other

### Version / Commit

v1.2.3 / 223ddc2

PR #1271, opened 2026-09-16
fix(copaw): refuse complete_project while project tasks are not terminal

### LINKED ISSUE #1261: CoPaw complete_project can mark projects completed while tasks are still planned
### What happened
On AgentTeams v1.2.3, CoPaw's `complete_project` can mark a project as `completed` even when its DAG or Loop still contains non-terminal tasks such as `planned` or `assigned`.

The current implementation writes the project status directly and does not inspect the task graph before completing the project.

### What I expected
`complete_project` should reject completion unless every task in the DAG/Loop graph has reached a terminal status. A project should not become terminal while work is still schedulable.

### Steps to Reproduce

1. Create a CoPaw project.
2. Add a DAG task and leave it in `planned` or `assigned` status.
3. Call `complete_project` for that project.
4. Observe that the project status becomes `completed` even though the task graph still has non-terminal work.

### AI Analysis

Root cause: `complete_project` treats project completion as a status write, but does not enforce the aggregate invariant over the task graph.

- The function currently rebuilds `ProjectMeta` with `status="completed"` and writes it back.
- It does not read or inspect the project plan before completing the project.
- Project completion should require every DAG/Loop task to be in a terminal status, such as `completed`, `revision`, `blocked`, or `cancelled`.

Suggested fix:

- Parse the project plan before writing terminal status.
- If any task is non-terminal, raise an error listing the offending task IDs and statuses.
- Reuse the same terminal-status definition used by the Controller project API so CoPaw and Controller cannot disagree.

### Relevant Logs

```shell
No runtime logs attached. This is reproducible directly at the code level: a project with one `planned` task can be passed to `complete_project` and still become `completed`.
```

### Component

CoPaw Worker

### Version / Commit

v1.2.3 / 223ddc2

dchaudhari7177:
Fixes #1261.

## What

`complete_project` in `copaw/src/copaw_worker/task.py` rebuilt the project meta with `status="completed"` and wrote it back without reading the plan. A project whose DAG, or whose Loop's current iteration, still had `pending` or `delegated` tasks became terminal while that work was still schedulable.

It now reads the plan first, takes the Loop's current-iteration tasks or the DAG tasks (the same selection `prepare_task` uses), and raises `TaskflowError` unless every task is in `TERMINAL_TASK_STATUSES` (`completed`, `revision`, `blocked`, `cancelled`). That's the invariant the controller's `CompleteProject` already enforces with a 409, so CoPaw and the controller no longer disagree on the same project. The error lists the tasks that block completion:

```text
complete_project requires every project task to be terminal; not terminal: tp-open-02 (pending), tp-open-03 (delegated)
```

`projectflow` already turns a `TaskflowError` into `{"ok": false, "error": ...}`, and the project meta is left as it was. A project with no tasks still completes, which is what `test_project_lifecycle_actions_only_update_meta_status` covers.

TeamHarness's `complete_project` (#1262) has the same gap, but its tests are the Ruby-driven `test-projectflow.rb` / `test-taskflow.rb`, and at least one of them (`daily-plan-2026-06-03`) completes a project that still has a planned task. I couldn't run Ruby here, so I've left that one for a separate change rather than edit tests I can't execute.

## Tests

Two new tests in `copaw/tests/test_taskflow_tool.py`, both through the `projectflow` tool:

- **DAG:** one `[x]`, one `[ ]` and one `[~]` task. Completion is refused, the error names only `tp-open-02 (pending)` and `tp-open-03 (delegated)`, and `meta.json` stays `active`. After marking them `[!]` (blocked) and `[-]` (cancelled), completion succeeds.
- **Loop:** a pending current-iteration task blocks completion; once it's `[x]`, completion succeeds.

Run the way `test-controller.yml` runs it (`agentscope==2.0.4.post1`, `PYTHONPATH=copaw/src`), on Windows / Python 3.12:

```text
$ python -m pytest copaw/tests/test_taskflow_tool.py
this branch:                     5 failed, 96 passed
task.py restored from main:      7 failed, 94 passed   (the two new tests)
```

The 5 failures are identical on both trees, so they aren't from this change. One is a `UnicodeEncodeError` from Windows' cp1252 console encoding. The other four are `test_leader_decision_maps_result_status_to_terminal_state` cases failing on the assertion at line 258. I didn't dig into those, so I can't say whether Ubuntu CI sees them.


shiyiyue1102 (approved):
LGTM. The completion guard checks DAG tasks or the current Loop iteration before updating project metadata, with focused regression coverage. Controller tests and all applicable integration checks now pass on this commit.

PR #1275, opened 2026-09-16
fix(controller): probe the worker backend before ensure-ready reports Ready

### LINKED ISSUE #1260: ensure-ready can return Ready when the worker backend is stopped
### What happened
On AgentTeams v1.2.3, `POST /api/v1/workers/{name}/ensure-ready` can return `Ready` even when the actual worker backend container is stopped or missing.

The handler trusts two stale signals:

- the Worker CR's `Status.Phase == "Running"`
- the controller's in-memory readiness map

If the real backend was stopped or removed after those signals were set, `ensure-ready` does not probe the backend and does not call `Start`.

### What I expected
`ensure-ready` should treat the backend as the final runtime check. When the CR says `Running`, it should still probe the backend; if the backend reports `stopped` or `not found`, it should clear stale readiness and enter the start/reconcile path. If backend status cannot be determined, it should fail closed rather than return `Ready`.

### Steps to Reproduce

1. Create a worker and wait until its CR `Status.Phase` is `Running` and the controller ready map marks it ready.
2. Stop or remove the actual worker container outside the controller, for example with `docker stop` or `docker rm`.
3. Call `POST /api/v1/workers/{name}/ensure-ready`.
4. Observe that the response is still `Ready`, and the backend is not restarted.

### AI Analysis

Root cause: `EnsureReady` treats the CR status snapshot and the in-memory ready flag as authoritative runtime truth.

- `Status.Phase` is written asynchronously by the reconciler and can remain `Running` after an external backend change.
- The in-memory ready map only records the worker's last successful self-report; it cannot detect a later backend stop.
- The handler only calls `Start` when the CR phase is `Stopped` or `Sleeping`.

Suggested fix:

- For a CR phase of `Running`, call the backend's `Status()` method.
- If status is `stopped` or `not found`, set the needs-start path and clear the stale ready flag.
- If backend status returns an error, fail closed instead of returning `Ready` based on stale state.

### Relevant Logs

```shell
No runtime logs attached. This was reproduced with controller-level state: CR phase `Running`, ready map `true`, and backend status `stopped`; `EnsureReady` returned `Ready` without calling `Start`.
```

### Component

Other

### Version / Commit

v1.2.3 / 223ddc2

dchaudhari7177:
Fixes #1260.

## What

`EnsureReady` (`agentteams-controller/internal/server/lifecycle_handler.go`) returned `Ready` based on two signals the backend never refreshes: the Worker CR's `Status.Phase == "Running"` and the in-memory ready flag from the worker's last self-report. It only called `Start` for a `Stopped` or `Sleeping` phase. A container stopped or removed after those signals were set kept coming back `Ready` and was never restarted.

For a `Running` phase the handler now asks the detected backend with `Status()`:

| backend reports | `ensure-ready` does | response |
|---|---|---|
| `running` | nothing new | `Ready` if the worker has self-reported, as before |
| `stopped` / `not_found` | the existing start path: `spec.state=Running`, `Start`, clear the ready flag, `status.phase=Running` | `Running` |
| `starting`, `failed`, `unknown` | no start, since the backend already has it or it needs attention | `Running` |
| `Status()` error | fails closed, logs the error | `Running` |

Some notes on the choices:

- **Not ready is not the same as a stale flag.** In the `starting`/`failed`/error rows the ready flag is left alone, and only `Ready` is withheld. A worker that really is running gets `Ready` back on the next call once the backend answers `running`. Clearing the flag there would force a fresh self-report after a transient Docker API error.
- **No backend detected keeps the old behaviour.** `DetectWorkerBackend` returning `nil` means there's nothing to probe, so the self-report still decides. I pinned that in a separate test rather than turning it into a hard failure.
- The HTTP status stays `200` in every row. `agt worker ensure-ready` prints the phase, and the manager's `lifecycle-worker.sh` reads `/status` rather than this endpoint, so neither consumer changes.

## Tests

`TestLifecycleEnsureReadyProbesBackendWhenRunning` is table-driven. Each case starts from a `Running` CR with the ready flag already set, which is the stale state from the issue. It asserts the response phase, the number of `Start` calls, the ready flag afterwards and the CR phase. `stubWorkerBackend` gained a `statusErr` field for the error row. `TestLifecycleEnsureReadyWithoutBackendKeepsSelfReport` covers the nil-backend case.

```text
$ go test ./internal/server/ -run TestLifecycleEnsureReady -count=1 -v
--- PASS: TestLifecycleEnsureReadyStartsSleepingWorker
--- PASS: TestLifecycleEnsureReadyProbesBackendWhenRunning   (6 subtests)
--- PASS: TestLifecycleEnsureReadyWithoutBackendKeepsSelfReport
ok
```

With only `lifecycle_handler.go` reverted to `main`, five of the six subtests fail: stopped, not_found, starting, failed and status error. `running backend stays ready` passes on both, as it should.

`go vet ./internal/server/` is clean, and `gofmt` is clean on both files. The full `./internal/server/` run has five failures here: `TestSkills_Upload*` ×4 and `TestTeamSkillStandardZipToWorkerE2E`. The same five fail on `main` on this machine (Windows, zip paths), so they are unrelated. `./cmd/controller` does not build locally for want of a 64-bit cgo toolchain, so that is CI's.


shiyiyue1102:
Thanks for the fix. One fail-closed gap remains at `backendRunning := b == nil`: `DetectWorkerBackend()` also returns nil when a configured backend is unavailable. In Docker mode, `Available()` returns false if the socket is missing or the daemon ping fails/times out. With a Running CR and a stale ready flag, this skips `Status()` and still returns `Ready` when the daemon is unreachable.

Could you default to not-ready when backend detection fails? If a backend-free mode must retain the old behavior, please distinguish it from a configured-but-unavailable backend and add a regression test for the latter.

On head `8e5afc53`, the existing `TestLifecycle*` tests pass locally. An additional test with a registered backend whose `Available()` returns false reproduces the issue: expected non-Ready, got `Ready`.

dchaudhari7177:
Good catch — that gap is real, and `60a0a16` closes it.

The root cause is that `DetectWorkerBackend()` overloads `nil` for two
situations that need opposite answers: nothing registered at all, versus a
registered backend that is currently unreachable. `backendRunning := b == nil`
gave both the permissive answer.

Added `Registry.HasWorkerBackends()`, which reports registration regardless of
availability, and gated the Ready answer on it:

```go
backendRunning := b == nil && !h.registry.HasWorkerBackends()
```

So backend-free mode keeps the old behavior (`TestLifecycleEnsureReadyWithoutBackendKeepsSelfReport`
still passes unchanged), while configured-but-unavailable now fails closed.

The regression test you described is in as
`TestLifecycleEnsureReadyWithUnavailableBackendIsNotReady`: a registered stub
with `Available() == false`, a `Running` CR and a stale ready flag. I confirmed
it reproduces before the fix —

```
--- FAIL: TestLifecycleEnsureReadyWithUnavailableBackendIsNotReady
    lifecycle_handler_test.go:272: expected response phase Running, got "Ready"
```

— and passes after. It also asserts `Status()` is never called on an
unavailable backend, so the fix can't regress into probing a dead daemon.
All ten `TestLifecycle*` tests and `internal/backend` pass locally.

shiyiyue1102 (approved):
Verified 60a0a16: configured-but-unavailable backends now fail closed, with regression coverage. The complete internal/backend and internal/server packages pass locally. Controller CI and integration checks pass after the requested rerun. The earlier readiness blocker is resolved.

PR #1282, opened 2026-09-17
ci: slim integration matrix and retire standalone CoPaw regression

shiyiyue1102:
## Summary

Integration CI repeats shared scenarios across runtime combinations and continues running standalone CoPaw regression as CoPaw moves to QwenPaw. Reduce the integration matrix from 10 to 7 jobs and default script invocations from 74 to 47 (36.5% fewer invocations; wall-clock savings are not measured).

- Retire CoPaw interaction, controller, and legacy Team DAG shards. Preserve OpenClaw, QwenPaw, and Hermes interaction/runtime-specific checks and QwenPaw TeamHarness coverage.
- Run fixed runtime-switch and name-validation scenarios once on QwenPaw; run Python Manager startup/migration checks once instead of repeating them per Worker runtime.
- Keep the CoPaw worker image build for the real CoPaw-to-QwenPaw migration in test 23, and retain Manager migration coverage in test 28.
- Skip unrelated Manager session stabilization for controller-only scenarios and all no-LLM runs. Schedule test 100 cleanup last instead of between tests 10 and 11, and wait for resource finalizers as well as container removal within its existing timeout.
- Use the configured MCP client environment in direct QwenPaw probes and wait for the expected team role before checking projected context.
- Remove the obsolete pull_request_target shard-expansion bridge, which would otherwise reintroduce deselected cases. Add lightweight scheduling regression checks to CI.

## Coverage tradeoff

CoPaw no longer receives standalone full regression. Legacy test 21 remains available for manual diagnostics but leaves the default matrix and release-baseline filter. QwenPaw test 26 covers its TeamHarness collaboration path; it is not assertion-for-assertion equivalent to the retired CoPaw DAG scenario. No scenario files or product runtime code are removed.

## Validation

- `python3 tests/check-integration-scheduling.py`: 8 tests passed, covering runtime allocation, retained migration builds, fork/no-secret selection, cleanup ordering and asynchronous resource finalization, session waits, and explicit filters.
- `bash -n tests/run-all-tests.sh`: passed.
- Workflow YAML parsing and `git diff --check`: passed.
- Full container/LLM E2E has not been run locally; CI validation is pending.


shiyiyue1102:
The first CI run exposed a cleanup synchronization gap: test 100 saw no containers and immediately asserted Worker deletion while controller finalizers were still completing. Fixed in fd8552f2 by waiting for both resources and containers within the existing 120-second limit. API failures and timeouts still fail the test; 7 local regression checks pass.

The QwenPaw test 19 failure is separate: its Leader exited during startup after `/api/mcp/tools/teamharness` returned HTTP 502. That failure remains under investigation; no runtime code or assertions were weakened. Merge remains blocked pending the new head's full CI results.

shiyiyue1102:
The updated run passes cleanup on the OpenClaw and Hermes controller shards, and QwenPaw test 19 no longer fails. The remaining failures exposed two test-context issues, addressed in 8578cd12:

- The test 26 direct MCP probe let the container's initial `standalone` environment override the configured client's reconciled `team_leader` role. It now honors explicit client environment values and only recovers masked values from known sources. A regression test reproduces the old role mismatch and passes with the fix.
- Test 18 now waits for each member's expected role in TEAMS.md, rather than treating a generic context header as proof that the role update has completed. Role assertions and the existing timeout remain unchanged.

Eight local checks and shell syntax validation pass. No product/runtime code or result assertions were changed; full CI on the new head is still required before merge.

shiyiyue1102:
The role probe still failed on 8578cd12 because QwenPaw 2.0.1 treats custom environment keys as secrets and masks their API response values (confirmed in the pinned wheel's `mcp_binding.py`). The previous change only corrected literal-value precedence.

0315f071 now recovers masked role values from the Controller-projected runtime.yaml member role, failing if that role is absent. It does not infer privileges from the requested action or hard-code a leader identity. Eight local checks pass.

The same CI run passed tests 18/19 and cleanup, but separately failed CoPaw state persistence/migration in test 23 and TeamHarness startup during skill distribution in test 24. Those remain merge blockers pending diagnosis and new-head verification; no assertions or runtime behavior were relaxed.

Issue #1286, opened 2026-09-17
[Bug] Driver policy passes readback but cross-session calls get driver_policy_denied; Worker console ports publish on 0.0.0.0 (patch offered)


Shuuyou:
Hi AgentTeams team — we are **CyberGuard** ([github.com/elsechord/CyberGuard](https://github.com/elsechord/CyberGuard)), a finalist project in the GOAI 2026 Agent Infra track building an autonomous security-operations team on top of AgentTeams v1.2.2. Two local-hardening findings from our deployment; for the second one we already carry a one-line patch and would be glad to upstream it as a PR.

### Environment

- AgentTeams v1.2.2 (pinned release, commit `849182af`), Docker Desktop, Windows
- QwenPaw Workers; investigation MCP servers registered through Higress

### Steps to reproduce (condensed)

1. Configure the Driver policy for two investigation MCP drivers through the native management API: default `deny`, allow exactly three named investigation tools for Matrix sources.
2. Read the policy back after the write and evaluate the intended allow/deny cases with the installed native policy evaluator — all cases pass deterministically at configuration time.
3. Start a new Worker session and have the model call one of the allowed tools from its task room.

### Actual

Cross-session tool calls still fail with `driver_policy_denied`, even though the same policy evaluates as `allow` in the deterministic check. For us, policy propagation/evaluation context across sessions remains an unresolved issue.

A related observation: a policy rule scoped to **Matrix sources** never matched in practice; we suspect the request-context channel/source fields are missing at evaluation time on some paths. We only got predictable behavior with exact-tool rules that do not depend on request-context channel fields.

### Expected

A policy that reads back correctly and passes the native evaluator's allow/deny checks should be enforced identically in live sessions — or the docs/API should reveal which session/context attributes participate in evaluation so mismatches become debuggable.

### Questions / suggestions

- Is Driver policy evaluation scoped per session, per driver instance, or global? Which context (channel, source, room) is available at evaluation time?
- Would you consider a "policy test" API that evaluates a policy against a live session context, so configuration-time checks match runtime behavior?
- Possibly related: #1283 (lost update in `DriverManager.reload_driver` reverting concurrent policy writes) — a different failure mode in the same subsystem; linking it here in case it interacts with the cross-session propagation we observed.

### Symptom

[...]

PR #1287, opened 2026-09-18
fix(backend): bind automatic worker console ports to loopback (HostIP)

Shuuyou:
## Background

As reported in #1286 (and confirmed by the maintainer), workers that expose a console via `AGENTTEAMS_CONSOLE_PORT` get an automatically published host port in the Docker backend, but the binding is created without a `HostIp`. Docker therefore publishes the port on `0.0.0.0`, which makes the worker console reachable from every host interface — a broader exposure than intended for an automatically assigned, unauthenticated-by-default debug console.

Location: `agentteams-controller/internal/backend/docker.go`, `buildCreatePayload` (the "Console port binding (CoPaw workers)" block).

## What this PR changes

- The automatic console port binding now sets `HostIP: "127.0.0.1"`, so the console is only reachable from the host itself (one-line change plus an expanded comment).
- Adds two regression tests in `internal/backend/docker_test.go`:
  - `TestDockerCreateBindsConsolePortToLoopback` — goes through `DockerBackend.Create` against a mock Docker API and asserts the wire payload binds `8088/tcp` to `HostIp 127.0.0.1` with the reported `ConsoleHostPort`; also asserts workers without `AGENTTEAMS_CONSOLE_PORT` publish no ports.
  - `TestBuildCreatePayloadPreservesExplicitPortMappingHostIP` — asserts explicit `CreateRequest.Ports` mappings keep their `HostIP` verbatim (empty stays empty = all interfaces; a specific IP passes through).

## Scope

- Explicit port mappings (`CreateRequest.Ports`) are untouched and keep their existing behavior, including the ability to request all-interfaces bindings.
- Addresses the HostIP part of #1286 only. The policy-evaluation aspect raised in the issue remains under investigation there and is intentionally out of scope here to keep this PR minimal.

## Design questions for review

Since this changes the default bind address of the automatic console port, I'd appreciate maintainer input on the configuration surface:

1. **Configurability** — is a hard-coded `127.0.0.1` acceptable for now, or should the loopback address be configurable (env var / controller flag / worker spec field) before this lands? If configurability is preferred, happy to extend the PR accordingly.
2. **Remote-access compatibility path** — for deployments that legitimately need the console reachable off-host, is publishing an explicit port mapping (which already supports a caller-chosen `HostIP`) the intended path going forward, or should there be a first-class opt-in (e.g. a spec toggle) instead?

## Test results

`go test ./internal/... ./cmd/... -count=1` (same command as the `test-unit` CI target, Go 1.25):

```
=== RUN   TestDockerCreateBindsConsolePortToLoopback
--- PASS: TestDockerCreateBindsConsolePortToLoopback (0.03s)
=== RUN   TestBuildCreatePayloadPreservesExplicitPortMappingHostIP
--- PASS: TestBuildCreatePayloadPreservesExplicitPortMappingHostIP (0.00s)
PASS
ok  	github.com/agentscope-ai/AgentTeams/agentteams-controller/internal/backend  1.282s
```

All other unit packages pass as well (`internal/backend`, `internal/executor`, `internal/service`, `internal/server`, `cmd/agt`, ...).

---

Thanks for the pointer to the exact location in #1286 — it made this easy to scope. Feedback welcome, especially on the two questions above.


shiyiyue1102 (approved):
LGTM. The automatic worker console binding is restricted to loopback while explicit port mappings retain their existing behavior. Reviewed the container creation and checkpoint access paths; go test ./internal/backend -count=1 passes on this commit.

Shuuyou:
Thank you for the fast review and merge — and for the scoping guidance on the issue, which made this a much tighter change than our original local patch. We'll keep an eye on the bind-address configurability question raised in the description and are happy to follow up if you decide to expose it as a config surface. Thanks again!

PR #1290, opened 2026-09-18
fix(build): pin pnpm to the OpenClaw package manager version

shiyiyue1102:
## Problem
The OpenClaw base image installs an unpinned pnpm. On current main, the resulting package-manager switch selects a standalone pnpm executable for OpenClaw's `pnpm@10.32.1` requirement. OpenClaw's build runner invokes `npm_execpath` through Node.js, which parses the ELF binary as JavaScript and fails at `canvas:a2ui:bundle` with `SyntaxError: Invalid or unexpected token`.

## Change
Pin the npm-installed pnpm to `10.32.1`, matching the pinned OpenClaw commit's `packageManager`. The runner then receives `pnpm.cjs` and can execute it through Node.js. Only `openclaw-base/Dockerfile` changes.

## Validation
- Reproduced the same ELF/SyntaxError failure in an isolated Linux arm64 container using the actual `pnpm-runner.mjs` and helper from OpenClaw commit `2f35b6fa6b65d012e3b0c9f24af3f8a4b617a6e0`: exit 1.
- Installed `pnpm@10.32.1` in the same container and reran the same nested-script probe: exit 0, using `/usr/lib/node_modules/pnpm/bin/pnpm.cjs`.
- `git diff --check` passed.
- Full image build and integration regression remain for CI; the local probe validates the failing invocation mechanism only.

Failing main run: https://github.com/agentscope-ai/AgentTeams/actions/runs/35332589097


shiyiyue1102:
CI confirms the pnpm fix: OpenClaw base and all downstream image builds passed, as did the OpenClaw interaction and controller suites.

The remaining QwenPaw TeamHarness failure is unrelated to the runner: 116/118 assertions passed, including actual task completion and leader verification. The two failures are `auth:api` expectations for the external `gateway.example.com` fixture, which must not receive gateway credentials. PR #1291 corrects those assertions; its local real-model TeamHarness run passed 118/118. Keeping this PR scoped to the pnpm pin.

PR #1291, opened 2026-09-18
fix(embedded): package worker skills and correct stopped worker status

shiyiyue1102:
## Problem

The embedded Controller image included the skill push script but omitted `manager/agent/worker-skills`. Creating a QwenPaw worker with `github-operations` therefore logged a skill assignment warning and then crashed during startup with `assigned skill files are missing`. The status API could simultaneously report `phase: Running` and `containerState: stopped` because it retained the CR phase when reading fresh backend state.

## Changes

- Package the on-demand worker skills in the embedded image.
- Report `Stopped` when the backend is stopped but the stored phase is Running/Ready, preserving intentional Sleeping status.
- Require runtime readiness in the worker creation integration test.
- Correct the TeamHarness fixture assertion: an external MCP host must not receive the gateway Authorization credential.

## Validation

- Added status regression tests; reproduced failures before the fix, then passed `go test ./internal/server ./internal/controller`.
- Built fresh Controller, embedded, and QwenPaw Worker images from main `40279ce0` plus this patch on Linux arm64 (OrbStack). Reused the unchanged QwenPaw Manager source image.
- Verified the skill exists in the image itself, then recreated the isolated Controller and created fresh workers without copying files into running containers.
- Eight local scripts (01–06, 25, 26) completed with **170 passing assertions, 0 failures**. Real Qwen `qwen3.6-plus` regression: startup, worker creation, task assignment, human intervention, heartbeat, name validation, and TeamHarness workflow passed. TeamHarness: **118/118**, including live delegation, submitted artifacts, leader verification, and hot updates without container restart.
- Independently checked completed task metadata and README output; executed the amended Python deliverable with and without a name argument.
- Multi-worker collaboration produced shared frontend/backend artifacts. The first output lacked same-origin integration; an explicit acceptance follow-up caused the agents to correct it. Independently ran the final server and verified HTTP 200 for both `/` (HTML) and `/api/hello` (contract JSON), with task metadata completed. The test-06 script also completed successfully; its broad assertions were supplemented by these artifact checks.
- Stopped a test container and confirmed `Stopped / stopped`; restarted it afterward. Project create/pause/resume/complete also passed.

The local harness uses unique worker names and isolated resources. QwenPaw token metrics are not reliable in the existing JSONL-oriented metrics helper, so zero counters are not treated as performance evidence. This validation does not cover amd64, Kubernetes, or other agent runtimes.


shiyiyue1102:
@johnlanni All CI checks are passing, including the real-model integration suites. Could you review this for v1.2.4? Merging it first also fixes the unrelated TeamHarness assertions blocking #1290.

